PR #5, opened 2018-07-19
Adding tests, tweaking built_value code.

RedBrogdon:


Sfshaza:
Please review, @filiph!

jsonexample/lib/built_value/built_complex_object.dart
```diff
@@ -23,14 +23,19 @@ abstract class BuiltComplexObject
   @nullable
   double get aDouble;
 
+  @nullable
```
filiph: This makes me sad, but I know you're doing it for a reason.

jsonexample/test/simple_object_unit_test.dart
```diff
@@ -0,0 +1,196 @@
...
+// found in the LICENSE file.
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:jsonexample/built_value/built_simple_object.dart';
+import 'package:jsonexample/built_value/built_value_serializers.dart';
+import 'package:jsonexample/dart_convert/converted_simple_object.dart';
+import 'package:jsonexample/json_serializable/serializable_simple_object.dart';
+
+void main() {
+  const Map<String, dynamic> typicalObjectJson = {
```
filiph: [nit] I think the Dart style, if you want to type-annotate, is:

```
const typicalObjectJson = <String,dynamic>{
  // ...
}
```

And in this particular case, you don't even need to do that, since it's going to be inferred.

Types on the left are encourages for top-level members, but local variables tend to rely on inferred types.
https://www.dartlang.org/guides/language/effective-dart/design#prefer-type-annotating-public-fields-and-top-level-variables-if-the-type-isnt-obvious

If the second parameter wasn't dynamic, you could just leave generics altogether. But because of [this rule](https://www.dartlang.org/guides/language/effective-dart/design#prefer-annotating-with-dynamic-instead-of-letting-inference-fail), you want to use it.

jsonexample/test/simple_object_unit_test.dart
```diff
@@ -0,0 +1,196 @@
...
+      expect(simpleObject.anInt, 1);
+      expect(simpleObject.aDouble, 1.0);
+      expect(simpleObject.aListOfStrings, ['one', 'two', 'three']);
+      expect(simpleObject.aListOfInts, [1, 2, 3]);
+      expect(simpleObject.aListOfDoubles, [1.0, 2.0, 3.0]);
+    });
+
+    test('Empty object', () {
+      final simpleObject =
+          SerializableSimpleObject.fromJson(<String, dynamic>{});
```
filiph: [nit] Why not have this empty map among the constants as well?

jsonexample/test/simple_object_unit_test.dart
```diff
@@ -0,0 +1,196 @@
...
+      expect(simpleObject.anInt, 1);
+      expect(simpleObject.aDouble, 1.0);
+      expect(simpleObject.aListOfStrings, []);
+      expect(simpleObject.aListOfInts, []);
+      expect(simpleObject.aListOfDoubles, []);
+    });
+
+    test('Extra properties', () {
+      final simpleObject =
+          ConvertedSimpleObject.fromJson(unexpectedPropertyjson);
```
filiph: Not sure what you're testing with the `unexpectedPropertyjson`. Are you testing that it `returnsNormally`? In that case, the test should say "Extra properties are ignored".

Is there a way to force the deserialization to throw when the object has extra properties? In that case, we'd want to test that.

jsonexample/test/simple_object_unit_test.dart
```diff
@@ -0,0 +1,196 @@
...
+  const Map<String, dynamic> emptyListJson = {
+    'aString': 'Blah, blah, blah.',
+    'anInt': 1,
+    'aDouble': 1.0,
+    'aListOfStrings': [],
+    'aListOfInts': [],
+    'aListOfDoubles': []
+  };
+
+  const Map<String, dynamic> unexpectedPropertyjson = {
```
filiph: `unexpectedPropertyJson`

jsonexample/test/simple_object_unit_test.dart
```diff
@@ -0,0 +1,196 @@
...
+    'anInt': 1,
+    'aDouble': 1.0,
+    'aListOfStrings': ['one', 'two', 'three'],
+    'aListOfInts': [1, 2, 3],
+    'aListOfDoubles': [1.0, 2.0, 3.0],
+    'unexpectedProperty': 'Whoops!'
+  };
+
+  group('ConvertedSimpleObject unit tests', () {
+    test('Typical object', () {
```
filiph: [nit] The concatenated message should ideally read like a sentence. For example: "ConvertedSimpleObject deserializes a typical object".

```
group('ConvertedSimpleObject', () {
  test('deserializes a typical object', () {
```

jsonexample/test/simple_object_unit_test.dart
```diff
@@ -0,0 +1,196 @@
...
+      final simpleObject = ConvertedSimpleObject.fromJson(typicalObjectJson);
+
+      expect(simpleObject, isNotNull);
```
filiph: [nit] Purists would say that there should be one expect statement per unit test. I'm not a purist.

I do think these tests may be unnecessarily thorough. Especially below, where you're using json_serializable. Those tests look like something I could find in the json_serializable package. Here we could just make sure that the things don't throw, or that one of the members is what we expect it to be.

Definitely a nit, though.

jsonexample/test/widget_tests_test.dart
```diff
@@ -0,0 +1,193 @@
...
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:jsonexample/dart_convert/converted_simple_object.dart';
+import 'package:jsonexample/dart_convert/converted_complex_object.dart';
+import 'package:jsonexample/widgets.dart';
+
+void main() {
+  group('SimpleObjectView widget test', () {
+    testWidgets('Typical object', (WidgetTester tester) async {
```
filiph: dtto about test descriptions reading like sentences:

"SimpleObjectView shows JSON contents" or something like that

jsonexample/test/widget_tests_test.dart
```diff
@@ -0,0 +1,193 @@
...
+      );
+
+      expect(find.text('"Blah, blah, blah"'), findsOneWidget);
+      expect(find.text('1'), findsOneWidget);
+      expect(find.text('1.0'), findsOneWidget);
+      expect(find.text('["one", "two", "three"]'), findsOneWidget);
+      expect(find.text('[1, 2, 3]'), findsOneWidget);
+      expect(find.text('[1.0, 2.0, 3.0]'), findsOneWidget);
+
+      for (int i = 1; i <= 4; i++) {
```
filiph: [nit] I think that logic like this in unit tests is generally frowned upon. No big deal.

jsonexample/lib/built_value/built_complex_object.dart
RedBrogdon: Yeah, I did the other two first, so it's been changed to match. I'm annoyed that I didn't catch it earlier, but that's why one writes tests, I guess. 😄

jsonexample/test/simple_object_unit_test.dart
RedBrogdon: Done.

jsonexample/test/simple_object_unit_test.dart
RedBrogdon: I'm just testing that an extra, unexpected property is ignored and doesn't break anything. Updated the test description to better state this.

jsonexample/test/simple_object_unit_test.dart
RedBrogdon: Done.

jsonexample/test/simple_object_unit_test.dart
RedBrogdon: Rewrote all of 'em.

jsonexample/test/widget_tests_test.dart
RedBrogdon: Rewritten.

jsonexample/test/widget_tests_test.dart
RedBrogdon: I was torn about it myself. I don't like having loops in the tests, but these get enormous otherwise. Leaving as-is for now, but I'll more than likely come back and clean this up.

jsonexample/test/simple_object_unit_test.dart
RedBrogdon: You're right, but I'm not sure what other tests would go in here beyond the type of thing I have. I do like that it effectively compares all three approaches to ensure that they all pass the same tests, which is nice.

jsonexample/test/simple_object_unit_test.dart
RedBrogdon: Originally it was just the brackets {}, so it felt silly. Makes more sense now, though, so fixed.

PR #6, opened 2018-07-27
Adding Shrine to travis script.

RedBrogdon:


RedBrogdon:
This PR adds a dummy test to the Shrine app and restructures the travis script to use an array of project names that it builds and tests in order.

PR #9, opened 2018-08-01
Fixing unchangeable Shrine categories.

RedBrogdon:
This updates the Shrine app so that category selection and filtering work properly with the scoped model code already introduced for the shopping cart.

Due to the way the Backdrop's closing animation was triggered, this change has the unfortunate side effect of forcing the user to manually close the back layer after choosing a category. That'll need to be corrected in the next PR.

### DIFF
--- a/shrine/lib/category_menu_page.dart
+++ b/shrine/lib/category_menu_page.dart
@@ -13,55 +13,53 @@
 // limitations under the License.
 
 import 'package:flutter/material.dart';
-import 'package:meta/meta.dart';
+import 'package:scoped_model/scoped_model.dart';
 
 import 'colors.dart';
+import 'model/app_state_model.dart';
 import 'model/product.dart';
 
 class CategoryMenuPage extends StatelessWidget {
-  final Category currentCategory;
-  final ValueChanged<Category> onCategoryTap;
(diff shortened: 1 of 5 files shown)


shrine/lib/category_menu_page.dart
```diff
@@ -13,55 +13,53 @@
...
-                categoryString,
-                style: theme.textTheme.body2
-                    .copyWith(color: kShrineBrown900.withAlpha(153)),
-                textAlign: TextAlign.center,
-              ),
-            ),
+    return ScopedModelDescendant<AppStateModel>(
+      builder: (context, child, model) => GestureDetector(
+            onTap: () => model.setCategory(category),
+            child: model.selectedCategory == category
```
filiph: [nit] One of these days, we should refactor this. Something like having a (private?) sub-widget, or something like `_buildCategory(Category category, bool selected)`.

RedBrogdon: Eesh, yeah, that's a big builder. There's parts of this I'm reluctant to change because they're code from the codelab. Once the team is done with 105, though, we should have code parity between this version and the final codelab one, and can suggest changes from either side.

PR #10, opened 2018-08-07
Add platform_view_swift based on platform_view example

mblaughton:
This is a port of the _platform_view_ example implemented in Swift instead of Objective-C. I tried to keep it true to the original, and the port matches the original version's behavior on my device.

For reference, this was originally created as a [PR for the main flutter repository](https://github.com/flutter/flutter/pull/18394) but @RedBrogdon suggested it would fit better in the samples repo. The current PR also implements the changes recommended in the link.


platform_view_swift/lib/main.dart
```diff
@@ -0,0 +1,104 @@
+// Copyright 2018, the Flutter project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
```
RedBrogdon: Is `dart:io` getting used in this file?

platform_view_swift/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: platform_view_swift
+
+dependencies:
+  flutter:
+    sdk: flutter
+
+  collection: 1.14.6 # THIS LINE IS AUTOGENERATED - TO UPDATE USE "flutter update-packages --force-upgrade"
```
RedBrogdon: These version numbers (and the comments next to them) are holdovers from running `flutter update-packages`, which isn't the recommended approach for typical apps, so they should probably just be re-pinned manually. 

Actually, now that I'm looking at these deps, are any of these being used anywhere in this sample? It's possible they were copied over from another sample when the the original, Obj-C platform view sample was made, and they're no longer necessary.

Try cutting them all out and see if you can still build. :)

platform_view_swift/ios/Runner/AppDelegate.swift
```diff
@@ -0,0 +1,44 @@
...
+
+        let platformViewController = controller.storyboard?.instantiateViewController(withIdentifier: "PlatformView") as! PlatformViewController
+        platformViewController.counter = call.arguments as! Int
+        platformViewController.delegate = self
+
+        let navigationController = UINavigationController(rootViewController: platformViewController)
+        navigationController.navigationBar.topItem?.title = "Platform View"
+        controller.present(navigationController, animated: true, completion: nil)
+        }
+      else {
```
RedBrogdon: style nit: should this be `} else {` on one line?

platform_view_swift/lib/main.dart
mblaughton: Good catch -- removed.

platform_view_swift/pubspec.yaml
mblaughton: You were right -- not needed.

platform_view_swift/ios/Runner/AppDelegate.swift
mblaughton: Yes, I think you're right (though it will never look normal to me). :)

RedBrogdon:
Thanks. I'll update the index to include this one.

mit-mit:
@RedBrogdon are we moving `platform_view` to this repo also? Seems odd for the two to be in separate repos?

PR #13, opened 2018-08-31
Adds model layer and list view to veggieseasons.

RedBrogdon:


veggieseasons/lib/data/model.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:veggieseasons/data/veggie_provider.dart';
+
+class AppState extends Model {
+  List<Veggie> veggies;
+
+  AppState() : veggies = VeggieProvider.veggies;
+
+  List<Veggie> get allVeggies => veggies;
+
+  Veggie getVeggie(int id) => veggies.firstWhere((v) => v.id == id);
```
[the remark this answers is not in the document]
RedBrogdon: Another day, another Dart method I didn't know existed. Fixed.

veggieseasons/lib/data/model.dart
```diff
@@ -0,0 +1,71 @@
...
+  List<Veggie> get unavailableVeggies {
+    Season currentSeason = _getSeasonForDate(DateTime.now());
+    return veggies.where((v) => !v.seasons.contains(currentSeason)).toList();
+  }
+
+  List<Veggie> get favoriteVeggies =>
+      veggies.where((v) => v.isFavorite).toList();
+
+  List<Veggie> searchVeggies(String terms) => veggies
+      .where((v) => v.name.toLowerCase().contains(terms.toLowerCase()))
```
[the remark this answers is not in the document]
RedBrogdon: Ack.

veggieseasons/lib/data/veggie_provider.dart
```diff
@@ -0,0 +1,592 @@
+// Copyright 2018 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:veggieseasons/data/veggie.dart';
+
+class VeggieProvider {
```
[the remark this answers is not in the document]
RedBrogdon: Changed.

We should talk about this one offline, though. I wasn't planning to incorporate Firebase into the app, so we should make sure we've both got the same ideas for the finished product. 😄

veggieseasons/lib/widgets/veggie_headline.dart
```diff
@@ -0,0 +1,74 @@
...
+class VeggieHeadline extends StatelessWidget {
+  final Veggie veggie;
+
+  const VeggieHeadline(this.veggie);
+
+  List<Widget> _buildSeasonDots(List<Season> seasons) {
+    List<Widget> widgets = <Widget>[];
+
+    for (Season season in seasons) {
+      widgets.add(SizedBox(width: 4.0));
```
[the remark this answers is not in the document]
RedBrogdon: Yes. I used to use `Padding` for everything, but switched to using `SizedBox` if I'm just trying to block out space between widgets in a single dimension. Got the idea from someone on the eng team, but now I can't remember whom.

Was this just a coding-style question, or is there a functional reason you'd recommend using `Padding`? I'm not wed to it.

veggieseasons/pubspec.yaml
```diff
@@ -11,9 +11,26 @@ dependencies:
...
   flutter_test:
     sdk: flutter
 
 flutter:
+  assets:
+  - assets/images/placeholder.png
+  fonts:
+  - family: NotoSans
+    fonts:
+    - asset: assets/fonts/NotoSans-Regular.ttf
```
filiph: Are you really using all these font variants? Just checking.

veggieseasons/lib/data/model.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:scoped_model/scoped_model.dart';
+import 'package:veggieseasons/data/veggie.dart';
+import 'package:veggieseasons/data/local_veggie_provider.dart';
+
+class AppState extends Model {
+  List<Veggie> veggies;
+
+  AppState() : veggies = LocalVeggieProvider.veggies;
+
+  List<Veggie> get allVeggies => veggies;
```
filiph: [nit] This seems like a remnant of something from the past. Why not just use `veggies` since it's public.

veggieseasons/lib/data/model.dart
```diff
@@ -0,0 +1,71 @@
...
+      .where((v) => v.name.toLowerCase().contains(terms.toLowerCase()))
+      .toList();
+
+  void toggleFavorite(int id) {
+    Veggie veggie = getVeggie(id);
+    veggie.isFavorite = !veggie.isFavorite;
+    notifyListeners();
+  }
+
+  Season _getSeasonForDate(DateTime date) {
```
filiph: This method can be static and [nit] should probably live with Season.

veggieseasons/lib/main.dart
```diff
@@ -14,9 +15,12 @@ void main() {
   ]);
 
   runApp(
-    CupertinoApp(
-      color: Styles.appBackground,
-      home: HomeScreen(),
+    ScopedModel<AppState>(
+      model: AppState(),
+      child: CupertinoApp(
+        color: Color(0xffd0d0d0),
```
filiph: This could be `Colors.gray[350]` (which is `0xFFD6D6D6`).

veggieseasons/lib/screens/list.dart
```diff
@@ -4,19 +4,65 @@
...
+    return cards;
+  }
+
   @override
   Widget build(BuildContext context) {
-    return CupertinoPageScaffold(
-      navigationBar: CupertinoNavigationBar(
-        middle: Text('List'),
+    String dateString = DateFormat.yMMMMd("en_US").format(DateTime.now());
+    final model = ScopedModel.of<AppState>(context, rebuildOnChange: true);
```
filiph: I _think_ `rebuildOnChange` is `true` by default, in which case you can drop this.

veggieseasons/lib/widgets/veggie_headline.dart
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    return widgets;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return GestureDetector(
+      onTap: () {
+        /* navigation forthcoming */
```
filiph: [nit] Prepending `TODO(redbrogdon):` to the comment here will make it easier for the rest of us to make you feel bad later.

veggieseasons/lib/data/model.dart
```diff
@@ -0,0 +1,71 @@
+// Copyright 2018 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:scoped_model/scoped_model.dart';
+import 'package:veggieseasons/data/veggie.dart';
+import 'package:veggieseasons/data/local_veggie_provider.dart';
+
+class AppState extends Model {
```
filiph: You seem to be using ScopedModel yet never actually changing the state. I suppose that's just because you haven't implemented that functionality yet. Otherwise, you might be better off just using `InheritedWidget`.

veggieseasons/pubspec.yaml
RedBrogdon: Fair point. I have used them all at one point or another, but might not be right now. Let me get the rest of the app through review, and I'll double check at that point.

veggieseasons/lib/data/model.dart
RedBrogdon: Fixed by taking `veggies` private as `_veggies`.

veggieseasons/lib/data/model.dart
RedBrogdon: Made static. Season is just an enum.

veggieseasons/lib/main.dart
RedBrogdon: `Colors` is part of Material, so I can't use it. I made a constant for this in `Styles` that I wasn't using, though, so thanks for catching it.

veggieseasons/lib/screens/list.dart
RedBrogdon: I think it may have been in the past, but the default is `false` now:

https://github.com/brianegan/scoped_model/blob/master/lib/scoped_model.dart#L168

veggieseasons/lib/widgets/veggie_headline.dart
RedBrogdon: 😆

veggieseasons/lib/data/model.dart
RedBrogdon: Updates are coming. I have more of a complete app in https://github.com/redbrogdon/veggieseasons. I'm chunking it up while landing here to keep the PRs manageable.

PR #17, opened 2018-09-07
Adds search screen to veggieseasons.

RedBrogdon:
FWIW, this is my first experience with text editing, so don't be afraid to suggest a refactor of how the SearchScreen and SearchBar widgets are splitting work.

### DIFF
--- a/veggieseasons/lib/screens/search.dart
+++ b/veggieseasons/lib/screens/search.dart
@@ -13,9 +13,7 @@ import 'package:veggieseasons/widgets/veggie_headline.dart';
 
 class SearchScreen extends StatefulWidget {
   @override
-  _SearchScreenState createState() {
-    return _SearchScreenState();
-  }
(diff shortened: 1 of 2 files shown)


veggieseasons/lib/screens/search.dart
```diff
@@ -4,19 +4,88 @@
...
   @override
-  Widget build(BuildContext context) {
-    return CupertinoPageScaffold(
-      navigationBar: CupertinoNavigationBar(
-        middle: Text('Search'),
-      ),
-      backgroundColor: Styles.scaffoldBackground,
-      child: Center(
-        child: Text('Not yet implemented.'),
+  _SearchScreenState createState() {
```
xster: nit: just use one line for these

RedBrogdon: Fixed. I always forget that I can do that.

PR #19, opened 2018-10-09
Add bottom padding to search results

filiph:
Fixes an issue with the search screen where iOS virtual keyboard overlaps the bottom of the search results listing.


RedBrogdon:
I talked with @xster about this, and he recommended treating it as an issue with the Cupertino package:

https://github.com/flutter/flutter/issues/22923

We could pad this in the meantime, as you've done, but I'm concerned that we'll advertising a technique we don't recommend in the long term. So I'm going to close this. Feel free to reopen if you think it's important enough, though.

PR #20, opened 2018-10-17
Add Google Maps Place Tracker sample app to flutter/samples.

kenzieschmoll:
Sample app that displays places on a map. Add/edit places. Interact with map. Iterations to follow.


kenzieschmoll:
Hans, I made some style changes after reading through the style guide. Please let me know of any issues you see. Thanks!

kenzieschmoll:
screenshot_1539809084
screenshot_1539883681
screenshot_1539805569

place_tracker/lib/place_details.dart
```diff
@@ -0,0 +1,215 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:google_maps_flutter/google_maps_flutter.dart';
+
+import 'place.dart';
+
+class PlaceDetails extends StatefulWidget {
```
HansMuller: Classes like this usually do not mutate the state (place) that's passed to them. They just notify the parent widget that the user wants a change So:

```
class PlaceDetails extends StatefulWidget {
  const PlaceDetails({
    Key key,
    @required this.place,
    @required this.onChanged,
  }) : /* TBD asserts */ super(key: key);

  final Place place;
  final ValueChanged<Place> onChanged;

  @override
  PlaceDetailsState createState() => PlaceDetailsState();
}
```

The Stateful parent provides the `onChanged` callback and rebuilds its PlaceDetails child with the new place value.

The state class would not have a _place field, it would just refer to `widget.place`.

place_tracker/lib/place_details.dart
```diff
@@ -0,0 +1,215 @@
...
+  PlaceDetails(this._place);
+
+  final Place _place;
+
+  @override
+  PlaceDetailsState createState() => PlaceDetailsState(_place);
+}
+
+class PlaceDetailsState extends State<PlaceDetails> {
+  PlaceDetailsState(Place place) {
```
HansMuller: This isn't needed.

place_tracker/lib/place_details.dart
```diff
@@ -0,0 +1,215 @@
...
+  GoogleMapController mapController;
+  final TextEditingController _nameController = TextEditingController();
+  final TextEditingController _descriptionController = TextEditingController();
+
+  void onMapCreated(GoogleMapController controller) async {
+    mapController = controller;
+    mapController.addMarker(MarkerOptions(position: _place.latLng));
+  }
+
+  Widget _starBar() {
```
HansMuller: This function, `_nameTextField`, `_descriptionTextField`, and `_map`, can be stateless widgets. For example:
```
class _StarBar extends StatelessWidget {
  const _StarBar({
    Key key,
    this.maxStars = 5;
    @required this.rating,
    @required this.onChanged;
  }) : /* TBD asserts */ super(key: key);

  final int maxStars;
  final int rating;
  final ValueChanged<int> onChanged;

  @override
    return Row(
      mainAxisAlignment: MainAxisAlignment.center,
      children: List.generate(maxStars, (int index) {
        IconButton(
          icon: Icon(Icons.star),
          iconSize: 40.0,
          color: rating >= index ? Colors.amber : Colors.grey[400],
          onPressed: () {
            onChanged(rating);
          },
        ),
      }).toList(),
    );
  }
}
```

Note also: it's best to eliminate centering etc wrapping around a leaf widget like this. It's easier to understand things if the parent adds that stuff. I've left out the `Center` in this case; maybe you want to create the Row with `mainAxisSize: MainAxisSize.min` here.

place_tracker/lib/place_details.dart
```diff
@@ -0,0 +1,215 @@
...
+    );
+  }
+
+  Widget _detailsBody() {
+    return ListView(
+      padding: EdgeInsets.fromLTRB(24.0, 12.0, 24.0, 12.0),
+      children: <Widget>[
+        _nameTextField(),
+        _descriptionTextField(),
+        _starBar(),
```
HansMuller: Build a _StarBar here instead (likewise for the other TBD widgets) which creates a new Place and notifies the parent widget:

```
_StarBar(
  rating: widget.place.starRating,
  onChanged: (int value) {
    onChanged(place.copyWith(starRating: value));
  }
)
```
The parent PlaceMap would create its PlaceDetails child like this:
```
PlaceDetails(
 place: _places[marker],
 onChanged: (Place value) {
   _updatePlaceAndMarker(returnPlace, marker);
 },
)
```

place_tracker/lib/place.dart
```diff
@@ -0,0 +1,37 @@
...
+
+enum PlaceCategory {
+  FAVORITE,
+  VISITED,
+  WANT_TO_GO,
+}
+
+class Place {
+  Place(
+      {@required this.latLng,
```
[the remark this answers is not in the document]
RedBrogdon: 🤣

place_tracker/lib/place_details.dart
kenzieschmoll: In this example, what would the provided onChanged callback be?

_StarBar(rating: 5, onChanged: ???)

place_tracker/lib/place_details.dart
kenzieschmoll: I implemented this with a slight modification. I only want to notify the PlaceMap that the place has changed if the user presses the save button. Otherwise, the user would not have the ability to revert accidental edits (which they can do by just pressing back). So I made a local _place within PlaceDetailsState and that is the Place that the PlaceDetails page displays. Upon pressing save, this local place is passed back to the PlaceMap via the onChanged method of PlaceDetails.

place_tracker/lib/place_details.dart
kenzieschmoll: Removed.

place_tracker/lib/place_details.dart
kenzieschmoll: Done.

place_tracker/lib/place_details.dart
kenzieschmoll: Done with slight modification. See earlier comment.

place_tracker/lib/place.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:flutter/material.dart';
+import 'package:google_maps_flutter/google_maps_flutter.dart';
+
+enum PlaceCategory {
+  FAVORITE,
+  VISITED,
+  WANT_TO_GO,
+}
+
+class Place {
```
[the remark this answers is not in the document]
kenzieschmoll: Done.

place_tracker/lib/place.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:flutter/material.dart';
+import 'package:google_maps_flutter/google_maps_flutter.dart';
+
+enum PlaceCategory {
+  FAVORITE,
```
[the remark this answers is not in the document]
kenzieschmoll: Done.

place_tracker/.gitignore
```diff
@@ -0,0 +1,71 @@
+# Miscellaneous
```
kenzieschmoll: @RedBrogdon 
Is there anything in .gitignore that I need to modify to put this in flutter/samples?

place_tracker/lib/place_map.dart
```diff
@@ -0,0 +1,423 @@
...
+
+  final PlaceCategory selectedPlaceCategory;
+  final bool visible;
+  final ValueChanged<PlaceCategory> onChanged;
+
+  @override
+  Widget build(BuildContext context) {
+    return Opacity(
+      opacity: visible ? 1.0 : 0.0,
+      child: Padding(
```
HansMuller: In this case Container is a tiny bit more compact:
```
child: Container(
  padding: ...,
  Alignment: Alignment.bottomCenter,
  child: ButtonBar(...),
);
```

place_tracker/lib/place_map.dart
```diff
@@ -0,0 +1,423 @@
...
+        draggable: true,
+        icon: BitmapDescriptor.defaultMarkerWithHue(BitmapDescriptor.hueGreen),
+      ),
+    );
+    setState(() {
+      _pendingMarker = newMarker;
+    });
+  }
+
+  void _confirmAddPlace(BuildContext context) async {
```
HansMuller: Not sure why `BuildContext` is a parameter here (you can refer to `State.context` implicitly).  Does this context refer to a different Scaffold?

place_tracker/lib/place_util.dart
```diff
@@ -0,0 +1,49 @@
...
+        return BitmapDescriptor.defaultMarkerWithHue(BitmapDescriptor.hueAzure);
+    }
+  }
+
+  static List<Place> getPlacesForCategory(
+      PlaceCategory category, Map<Marker, Place> places) {
+    List<Place> favoritePlaces = List<Place>();
+    List<Place> visitedPlaces = List<Place>();
+    List<Place> wantToGoPlaces = List<Place>();
+    places.forEach((Marker marker, Place poi) {
```
HansMuller:
```
final List<Place> placesInCategory = <Place>[];
for (Place place in places.values) {
  if (place.category != category)
    continue;
  placesInCategory.add(place);
}
return placesInCategory;
```

Or

```
return places.values.where((Place place) => place.category == category).toList();
```

place_tracker/lib/place_util.dart
```diff
@@ -0,0 +1,49 @@
...
+      case PlaceCategory.visited:
+        return BitmapDescriptor.defaultMarkerWithHue(
+            BitmapDescriptor.hueViolet);
+      case PlaceCategory.wantToGo:
+      default:
+        return BitmapDescriptor.defaultMarkerWithHue(BitmapDescriptor.hueAzure);
+    }
+  }
+
+  static List<Place> getPlacesForCategory(
```
HansMuller: Looks like this could be a private method on PlaceMap (and it's always applied to _selectedCategory).

place_tracker/lib/place_util.dart
```diff
@@ -0,0 +1,49 @@
+import 'package:google_maps_flutter/google_maps_flutter.dart';
+import 'place.dart';
+
+class PlaceUtil {
+  static BitmapDescriptor getPlaceMarkerIcon(PlaceCategory category) {
```
HansMuller: Could be a private static on PlaceMap.

Maybe the switch should just determine the color...

place_tracker/lib/place_details.dart
```diff
@@ -159,57 +150,82 @@ class PlaceDetailsState extends State<PlaceDetails> {
...
+    Key key,
+    @required this.center,
+  }) : assert(center != null);
+
+  final LatLng center;
+
+  GoogleMapController mapController;
+
+  void onMapCreated(GoogleMapController controller) async {
+    mapController = controller;
```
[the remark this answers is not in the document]
kenzieschmoll: Can't I just perform these functions in the onMapCreated callback I am passing in the constructor?  i.e.

_Map(
  ...
  controller: _mapController,
  onMapCreated: (GoogleMapController controller)  {
    setState( () { 
     **_mapController = controller;
     _mapController.addMarker(MarkerOptions(position: center));**
    });
  }
);

Then in the _Map class, I can just used the class member onMapCreated like so:

GoogleMap(
  onMapCreated: onMapCreated,
  ...
),

place_tracker/lib/place_map.dart
```diff
@@ -0,0 +1,423 @@
...
+    );
+  }
+
+  void _zoomToFitPlaces(List<Place> places) {
+    // Default min/max values to latitude and longitude of center.
+    double minLat = _center.latitude;
+    double maxLat = _center.latitude;
+    double minLong = _center.longitude;
+    double maxLong = _center.longitude;
+    places.forEach((Place place) {
```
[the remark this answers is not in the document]
kenzieschmoll: Done. Are for in loops generally preferred over foreach loops?

place_tracker/lib/place_map.dart
```diff
@@ -0,0 +1,423 @@
...
+
+  @override
+  void initState() {
+    _selectedPlaceCategory = PlaceCategory.favorite;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
```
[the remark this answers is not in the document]
kenzieschmoll: Hmm this is giving me the error: "The constructor being called isn't a const constructor"

place_tracker/lib/place_map.dart
```diff
@@ -0,0 +1,423 @@
...
+        visible: place.category == _selectedPlaceCategory,
+      ),
+    );
+    _places[marker] = place;
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
+    mapController?.onInfoWindowTapped?.remove(_onInfoWindowTapped);
```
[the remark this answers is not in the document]
kenzieschmoll: I'm actually not sure that it is needed either. Removed.

place_tracker/lib/place_map.dart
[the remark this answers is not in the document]
HansMuller: Yes

place_tracker/lib/place_map.dart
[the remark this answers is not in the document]
HansMuller: Sorry about that. The appbar's parts appear to be const so I thought you could just apply const to the entire thing. You'll need to make the parameters const instead.

place_tracker/lib/place_map.dart
kenzieschmoll: Trying 
`Scaffold.of(State.context).showSnackBar(
  ...
);`
gives me the error
"Instance member 'context' can't be accessed using static access.'

HansMuller: By `State.context` I meant the `context` field in `State`.

kenzieschmoll: Oh I see. I got this error before I did some refactoring, but I got it to resurface using the context field in State here: 

Scaffold.of() called with a context that does not contain a Scaffold.
E/flutter ( 2186): No Scaffold ancestor could be found starting from the context that was passed to Scaffold.of(). This usually happens when the context provided is from the same StatefulWidget as that whose build function actually creates the Scaffold widget being sought.
E/flutter ( 2186): There are several ways to avoid this problem. The simplest is to use a Builder to get a context that is "under" the Scaffold. For an example of this, please see the documentation for Scaffold.of():
E/flutter ( 2186):   https://docs.flutter.io/flutter/material/Scaffold/of.html
E/flutter ( 2186): A more efficient solution is to split your build function into several widgets. This introduces a new context from which you can obtain the Scaffold. In this solution, you would have an outer widget that creates the Scaffold populated by instances of your new inner widgets, and then in these inner widgets you would use Scaffold.of().
E/flutter ( 2186): A less elegant but more expedient solution is assign a GlobalKey to the Scaffold, then use the key.currentState property to obtain the ScaffoldState rather than using the Scaffold.of() function.

HansMuller:
```
class _Home extends StatelessWidget {
  const _Home({ Key key }) : super(key: key);

  @override
    return PlaceMap(
      center: const LatLng(45.521563, -122.677433),
    );
  }
}

void main() {
  runApp(
    MaterialApp(
      debugShowCheckedModeBanner: false,
      title: 'Place Tracker',
      home: _Home(),
    );
}
```

kenzieschmoll: I added this modification but I am still getting the same error above.

place_tracker/lib/place_map.dart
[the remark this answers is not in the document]
kenzieschmoll: Consts added where applicable in previous commit :)

place_tracker/lib/place_map.dart
HansMuller: Sorry, I didn't really stop to think about that. The change I was suggesting above wasn't needed.

PlaceMap's build function creates its Scaffold, so Scaffold.of() can't find it because: same context.

Using a Builder to construct the Scaffold's body, should solve the problem.

place_tracker/lib/place_map.dart
```diff
@@ -0,0 +1,437 @@
...
+       assert(onCancelPressed != null),
+       super(key: key);
+
+  final bool visible;
+  final ArgumentCallback<BuildContext> onSavePressed;
+  final VoidCallback onCancelPressed;
+
+  @override
+  Widget build(BuildContext context) {
+    return Builder(builder: (BuildContext context) {
```
HansMuller: This Builder doesn't appear to be needed.

place_tracker/lib/place_map.dart
kenzieschmoll: I was able to remove the builder from _AddPlaceButtonBar entirely. I still need to pass context into the _confirmAddPlace method because otherwise, it will try to use the State's context, resulting in the before-mentioned error. I am passing the context from _AddPlaceButtonBar's main build method.

kenzieschmoll: Ah nevermind I see what you mean after reading your other comment below.

place_tracker/lib/place_map.dart
kenzieschmoll: Removed builder and added one around the scaffold's body. Also added a comment.

place_tracker/lib/place_map.dart
```diff
@@ -0,0 +1,441 @@
...
+        markerHue = BitmapDescriptor.hueViolet;
+        break;
+      case PlaceCategory.wantToGo:
+      default:
+        markerHue = BitmapDescriptor.hueAzure;
+    }
+    return BitmapDescriptor.defaultMarkerWithHue(markerHue);
+  }
+
+  static List<Place> _getPlacesForCategory(
```
HansMuller: I don't think the line breaks really help here.  
```
static List<Place> _getPlacesForCategory(PlaceCategory category, Map<Marker, Place> places) {
  return places.values.where((Place place) => place.category == category).toList();
}
```

place_tracker/lib/place_map.dart
```diff
@@ -0,0 +1,441 @@
...
+
+    // Add stub data on creation so we have something interesting to look at.
+    final Map<Marker, Place> places = await _initializeStubPlaces();
+    _zoomToFitPlaces(
+      _getPlacesForCategory(_selectedPlaceCategory, places),
+    );
+  }
+
+  Future<Map<Marker, Place>> _initializeStubPlaces() async {
+    for (Place place in PlaceStubData.places) {
```
HansMuller: It's a little more efficient to use `Future.wait` to wait for a list of Futures to complete.
```
await Future.wait PlaceStubData.places.map((Place place) => _initializeStubPlace(place)).toList();
```

place_tracker/lib/place_stub_data.dart
```diff
@@ -0,0 +1,99 @@
+import 'package:google_maps_flutter/google_maps_flutter.dart';
+
+import 'place.dart';
+
+class PlaceStubData {
+  static List<Place> places = [
```
HansMuller: This looks like it can be `static const` (or failing that, static final).
```
  static const List<Place> places = <Place>[
```

place_tracker/pubspec.yaml
```diff
@@ -0,0 +1,73 @@
...
+     git:
+       url: git://github.com/flutter/plugins
+       path: packages/google_maps_flutter
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+
+
+# For information on the generic Dart part of this file, see the
```
HansMuller: You could delete the unused boilerplate since it's not really part of the example.

place_tracker/lib/place_map.dart
kenzieschmoll: If I don't use line breaks, the line length will go over the suggested 80 char limit.

place_tracker/lib/place_map.dart
kenzieschmoll: Done.

place_tracker/lib/place_map.dart
HansMuller: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#prefer-line-length-of-80-characters

Although 80 character lines are preferred, if you look at the flutter sources you'll see that we often exceed that limit.

kenzieschmoll: Fixed.

PR #21, opened 2018-10-26
Added crashy sample.

kf6gpe:
Cleaning up repositories --- merge the crashy sample with the samples directory so we can archive the crashy repository.


crashy/lib/main.dart
```diff
@@ -0,0 +1,171 @@
...
+                  throw new StateError('This is an async Dart exception.');
+                }
+                bar() async {
+                  await foo();
+                }
+                await bar();
+              },
+            ),
+            new RaisedButton(
+              child: new Text('Java exception'),
```
RedBrogdon: "Method channel" might be better here, unless this doesn't work on iOS.

crashy/README.md
```diff
@@ -0,0 +1,25 @@
+# Crashy
```
RedBrogdon: One of the things we've been trying to do with each of the samples is give people a little head start by listing goals for the sample and the important files. The important files for this one would really just be main.dart and the missing dsn.dart, and the goals would just be to understand how using a zone can help you trap exceptions and report on them.

You can see the jsonexample's README for an example of how this works elsewhere:

https://github.com/flutter/samples/tree/master/jsonexample

kf6gpe:
@RedBrogdon Will fix the things you point out. So to be clear --- you're suggesting instead of retargeting this to samples, we retarget it to plugins as an example? I'm fine with that --- it kind of makes sense anyway, since it's a sample for a plugin, not how we use part of Flutter itself.

crashy/lib/main.dart
kf6gpe: If it's OK with you, I may punt on this suggestion for the repo refactor, and file an issue to do the same, just to separate concerns here.

crashy/lib/main.dart
```diff
@@ -0,0 +1,171 @@
...
+      // In development mode simply print to console.
+      FlutterError.dumpErrorToConsole(details);
+    } else {
+      // In production mode report to the application zone to report to
+      // Sentry.
+      Zone.current.handleUncaughtError(details.exception, details.stack);
+    }
+  };
+
+  // This creates a [Zone] that contains the Flutter application and stablishes
```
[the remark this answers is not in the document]
kf6gpe: Good catch. Will fix.

RedBrogdon:
Discussed this offline with @kf6gpe a few days ago, and agreed to move this to flutter/packages.

PR #22, opened 2018-10-29
Add stub reviews to the place details sheet. 

kenzieschmoll:
Offers a better demonstration of the map in a list view.

### DIFF
--- a/place_tracker/lib/place_details.dart
+++ b/place_tracker/lib/place_details.dart
@@ -8,9 +8,9 @@ import 'stub_data.dart';
 
 class PlaceDetails extends StatefulWidget {
   const PlaceDetails({
-    Key key,
     @required this.place,
     @required this.onChanged,
+    Key key,
   }) : assert(place != null),
        assert(onChanged != null),
        super(key: key);
@@ -39,10 +39,8 @@ class PlaceDetailsState extends State<PlaceDetails> {
   }
 
   void _onMapCreated(GoogleMapController controller) {
-    setState(() {
-      _mapController = controller;
-      _mapController.addMarker(MarkerOptions(position: _place.latLng));
-    });
+    _mapController = controller;
+    _mapController.addMarker(MarkerOptions(position: _place.latLng));
   }
 
   Widget _detailsBody() {
@@ -114,9 +112,9 @@ class PlaceDetailsState extends State<PlaceDetails> {
 
 class _NameTextField extends StatelessWidget {
   const _NameTextField({
-    Key key,
     @required this.controller,
     @required this.onChanged,
+    Key key,
   }) : assert(controller != null),
        assert(onChanged != null),
        super(key: key);
@@ -146,9 +144,9 @@ class _NameTextField extends StatelessWidget {
 
 class _DescriptionTextField extends StatelessWidget {
   const _DescriptionTextField({
-    Key key,
     @required this.controller,
     @required this.onChanged,
+    Key key,
   }) : assert(controller != null),
        assert(onChanged != null),
        super(key: key);
@@ -179,9 +177,9 @@ class _DescriptionTextField extends StatelessWidget {
 
 class _StarBar extends StatelessWidget {
   const _StarBar({
-    Key key,
     @required this.rating,
     @required this.onChanged,
+    Key key,
   }) : assert(rating != null && rating >= 0 && rating <= 5),
        assert(onChanged != null),
        super(key: key);
@@ -210,10 +208,10 @@ class _StarBar extends StatelessWidget {
 
 class _Map extends StatelessWidget {
   const _Map({
-    Key key,
     @required this.center,
     @required this.mapController,
     @required this.onMapCreated,
+    Key key,
   }) : assert(center != null),
        assert(onMapCreated != null),
        super(key: key);
@@ -250,7 +248,7 @@ class _Map extends StatelessWidget {
 
 class _Reviews extends StatelessWidget {
 
-  Widget _review(String reviewText) {
+  Widget _buildSingleReview(String reviewText) {
     return Column(
       children: <Widget>[
         Padding(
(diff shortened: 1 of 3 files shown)


place_tracker/lib/place_details.dart
```diff
@@ -140,10 +145,13 @@ class _NameTextField extends StatelessWidget {
 }
 
 class _DescriptionTextField extends StatelessWidget {
-  _DescriptionTextField({
+  const _DescriptionTextField({
+    Key key,
```
RedBrogdon: nit: I generally like to put the required ones first, then the others (particularly key, which is just getting passed through).

place_tracker/lib/place_details.dart
```diff
@@ -235,4 +246,88 @@ class _Map extends StatelessWidget {
...
+                fontSize: 24.0,
+                fontWeight: FontWeight.bold,
+                decoration: TextDecoration.underline,
+                color: Colors.black87,
+              ),
+            ),
+          ),
+        ),
+        Column(
+          children: StubData.reviews.map((String reviewText) => _review(reviewText)).toList(),
```
RedBrogdon: another nit: While flutter/flutter requires types on lambda parameters, I prefer to leave them out in our samples.

place_tracker/lib/place_details.dart
```diff
@@ -235,4 +246,88 @@ class _Map extends StatelessWidget {
...
+        ),
+      ],
+    );
+  }
 }
```
RedBrogdon: nit: newline needed.

place_tracker/lib/stub_data.dart
```diff
@@ -96,4 +96,10 @@ class PlaceStubData {
         category: PlaceCategory.wantToGo,
         starRating: 4),
   ];
+
+  static const List<String> reviews = [
```
RedBrogdon: `reviews` isn't that specific a property name since this holds not whole reviews, but just the text for them. It would likely be better as `reviewStrings`, `reviewTexts`, or something similar.

place_tracker/lib/place_details.dart
```diff
@@ -235,4 +246,88 @@ class _Map extends StatelessWidget {
       ),
     );
   }
+}
+
+class _Reviews extends StatelessWidget {
+
+  Widget _review(String reviewText) {
```
RedBrogdon: Consider a more specific name here. Some of our other examples use things like `_buildSingleReview` or `_buildReviewWidget` to more clearly indicate what's going on.

place_tracker/lib/place_details.dart
kenzieschmoll: Done.

place_tracker/lib/place_details.dart
```diff
@@ -4,12 +4,13 @@ import 'package:flutter/rendering.dart';
...
 
 import 'place.dart';
+import 'stub_data.dart';
 
 class PlaceDetails extends StatefulWidget {
   const PlaceDetails({
-    Key key,
     @required this.place,
     @required this.onChanged,
+    Key key,
```
HansMuller: Not that it really matters, but conventionally the key argument goes first, and if there's a child or children argument, they go last.

place_tracker/lib/place_details.dart
```diff
@@ -77,6 +76,7 @@ class PlaceDetailsState extends State<PlaceDetails> {
           mapController: _mapController,
           onMapCreated: _onMapCreated,
         ),
+        _Reviews(),
```
HansMuller: It would be slightly more efficient to use `const _Reviews()` here.

place_tracker/lib/place_details.dart
```diff
@@ -171,10 +177,11 @@ class _DescriptionTextField extends StatelessWidget {
 
 class _StarBar extends StatelessWidget {
   const _StarBar({
-    Key key,
     @required this.rating,
     @required this.onChanged,
+    Key key,
   }) : assert(rating != null && rating >= 0 && rating <= 5),
```
HansMuller: Might be a little better as `rating <= maxStars`

place_tracker/lib/place_details.dart
```diff
@@ -235,4 +244,88 @@ class _Map extends StatelessWidget {
       ),
     );
   }
-}
\ No newline at end of file
+}
+
+class _Reviews extends StatelessWidget {
+
```
HansMuller: Add a const constructor here to enable `const _Reviews()`.

place_tracker/lib/place_details.dart
```diff
@@ -235,4 +244,88 @@ class _Map extends StatelessWidget {
...
+                    color: Colors.grey,
+                  ),
+                ),
+                child: Row(
+                  mainAxisAlignment: MainAxisAlignment.center,
+                  children: <Widget>[
+                    const Text(
+                      '5',
+                      style: TextStyle(
+                          fontSize: 24.0,
```
HansMuller: two-space indent

place_tracker/lib/place_details.dart
```diff
@@ -235,4 +244,88 @@ class _Map extends StatelessWidget {
...
+                decoration: BoxDecoration(
+                  borderRadius: BorderRadius.circular(40.0),
+                  border: Border.all(
+                    width: 3.0,
+                    color: Colors.grey,
+                  ),
+                ),
+                child: Row(
+                  mainAxisAlignment: MainAxisAlignment.center,
+                  children: <Widget>[
```
HansMuller: It looks like this could be `const <Widget>[`

place_tracker/lib/place_details.dart
HansMuller: You-all are welcome to order parameters by whatever convention you like :-). 

If you look at the flutter sources, 
say https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/widgets/basic.dart, you'll see that we're pretty consistent.

PR #23, opened 2018-10-31
Add map type toggle to place_tracker demo.

kenzieschmoll:
Added a button that toggles between different map types: satellite, terrain, hybrid, and normal.

### DIFF
--- a/place_tracker/lib/place_map.dart
+++ b/place_tracker/lib/place_map.dart
@@ -238,12 +238,6 @@ class PlaceMapState extends State<PlaceMap> {
     MapType nextType =
       MapType.values[(mapController.options.mapType.index + 1) % MapType.values.length];
 
-    // We do not want to show MapType.none, so skip over it.
-    if (nextType == MapType.none) {
-      nextType =
-        MapType.values[(mapController.options.mapType.index + 2) % MapType.values.length];
-    }
-
     mapController.updateMapOptions(
       GoogleMapOptions(mapType: nextType),
     );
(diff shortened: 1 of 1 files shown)


kenzieschmoll:
screenshot_1541021563

place_tracker/lib/place_map.dart
```diff
@@ -234,6 +234,21 @@ class PlaceMapState extends State<PlaceMap> {
...
   }
 
+  void _onToggleMapTypePressed() {
+    MapType nextType =
+      MapType.values[(mapController.options.mapType.index + 1) % MapType.values.length];
+
+    // We do not want to show MapType.none, so skip over it.
+    if (nextType == MapType.none) {
+      nextType =
+        MapType.values[(mapController.options.mapType.index + 2) % MapType.values.length];
```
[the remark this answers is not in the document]
kenzieschmoll: No, the +2 will cause MapType.none to be skipped. But okay, I'll just keep MapType.none in the cycle.

place_tracker/lib/place_map.dart
```diff
@@ -278,9 +293,10 @@ class PlaceMapState extends State<PlaceMap> {
                 onSavePressed: () => _confirmAddPlace(context),
                 onCancelPressed: _cancelAddPlace,
               ),
-              _AddPlaceFab(
+              _MapFabs(
```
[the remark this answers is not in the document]
kenzieschmoll: This is possible, but then when the snackbar comes up saying "New placed added", it slides the fabs up with it, which looks weird. If I add the fabs to the stack, this is not a problem.

place_tracker/lib/place_map.dart
```diff
@@ -234,6 +234,15 @@ class PlaceMapState extends State<PlaceMap> {
     }
   }
 
+  void _onToggleMapTypePressed() {
+    MapType nextType =
```
HansMuller: final

place_tracker/lib/place_map.dart
```diff
@@ -400,33 +410,49 @@ class  _AddPlaceButtonBar extends StatelessWidget {
...
-          child: FloatingActionButton(
-            onPressed: onPressed,
-            materialTapTargetSize: MaterialTapTargetSize.padded,
-            backgroundColor: Colors.green,
-            child: const Icon(Icons.add_location, size: 36.0),
-          ),
+    return Container(
+      alignment: Alignment.topRight,
+      margin: const EdgeInsets.only(top: 12.0, right: 12.0),
+      child: Opacity(
```
HansMuller: The Visibility widget would be a little better here.

place_tracker/lib/place_map.dart
```diff
@@ -400,33 +410,49 @@ class  _AddPlaceButtonBar extends StatelessWidget {
...
+        opacity: visible ? 1.0 : 0.0,
+        child: Column(
+          children: <Widget>[
+            FloatingActionButton(
+              heroTag: "add_place_btn",
```
HansMuller: We don't usually abbreivate. Typically strings are `'foo'` rather than `"foo"`.

PR #28, opened 2018-11-07
Fixes Future<void> issue in Shrine.

RedBrogdon:
Also adds a new project file created by Xcode.


RedBrogdon:
Added a change of dependencies in jsonexample to fix a break introduced by yesterday's beta rollout. This unbreaks Travis and the sample.

PR #34, opened 2018-12-06
Adds some Shrine updates from Material team.

RedBrogdon:
@willlarche, you may recall the changes you helped me with just prior to #AskFlutter. This PR adds them to the actual repo for posterity

### DIFF
--- a/shrine/lib/login.dart
+++ b/shrine/lib/login.dart
@@ -22,33 +22,33 @@ class LoginPage extends StatefulWidget {
 }
 
 class _LoginPageState extends State<LoginPage> {
-  final _usernameController = TextEditingController();
-  final _passwordController = TextEditingController();
+  final TextEditingController _usernameController = TextEditingController();
+  final TextEditingController _passwordController = TextEditingController();
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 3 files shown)


shrine/lib/login.dart
```diff
@@ -58,32 +58,36 @@ class _LoginPageState extends State<LoginPage> {
...
               child: TextField(
                 controller: _passwordController,
-                decoration: InputDecoration(
+                decoration: const InputDecoration(
                   labelText: 'Password',
                 ),
               ),
             ),
-            ButtonBar(
+            Wrap(
```
willlarche: What does this change do?

RedBrogdon: Hmm. This was probably introduced for the flutter/flutter version of Shrine, which has an extra button (exit).  It's not serving much of a purpose here, other than keeping the code in sync with the other copy.

I can take it out.

willlarche: That exit button don't look good anyway. Should be a back button at the top.

RedBrogdon: In its defense, it was hacked into the code at the last minute by an engineer with no eye for the visual. 😄

PR #41, opened 2018-12-20
Add the ScopedModel Counter sample

filiph:
Adds sample app that mimics the beloved starter app, but with ScopedModel for state management.

Contains a smoke test and is set up to run it in `travis_script.sh`.


scoped_model_counter/README.md
```diff
@@ -0,0 +1,26 @@
...
+state synchronized over network) and which needs to be accessed and changed
+from different parts of your app.
+
+## Getting Started
+
+The only important part of the app is the `lib/main.dart` file. It has comments that will walk you
+through it.
+
+For more information on the `scoped_model` package (where `ScopedModel` comes from), please
+[read its documentation](https://pub.dartlang.org/packages/scoped_model).
```
chalin: For more information on the `scoped_model` package (...), <del>please [read its documentation](https://pub.dartlang.org/packages/scoped_model)</del>.

-> 

For more information on the `scoped_model` package (...), <ins>see the [package documentation](https://pub.dartlang.org/packages/scoped_model).</ins>

scoped_model_counter/lib/main.dart
```diff
@@ -0,0 +1,80 @@
...
+
+import 'package:flutter/material.dart';
+import 'package:scoped_model/scoped_model.dart';
+
+void main() {
+  // Initialize the model. Can be done outside a widget, like here.
+  var counter = Counter();
+
+  // We can now interact with the model from outside of the Flutter widget tree.
+  // Here, we wait 5 seconds and increment the counter.
```
chalin: Maybe for newbies it might be clearer to state something along the lines of the following:

// Setup a delayed interaction with the model (in 5 seconds time),
// outside of the Flutter widget tree

scoped_model_counter/lib/main.dart
```diff
@@ -0,0 +1,80 @@
...
+  Timer(
+    const Duration(seconds: 5),
+    () => counter.increment(),
+  );
```
chalin: Maybe this could be setup as a repeated interaction that occurs every 5 sec?

scoped_model_counter/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
+name: scoped_model_counter
+description: >
+  The starter Flutter application, but using ScopedModel to manage state.
+
+version: 1.0.0+1
+
+environment:
+  sdk: ">=2.0.0-dev.68.0 <3.0.0"
```
chalin: I guess that this comes from a template somewhere. Maybe it should be ">=2.0.0 <3.0.0"?

scoped_model_counter/test/widget_test.dart
```diff
@@ -0,0 +1,28 @@
...
+    expect(find.text('1'), findsNothing);
+
+    // Tap the '+' icon and trigger a frame.
+    await tester.tap(find.byIcon(Icons.add));
+    await tester.pump();
+
+    // Verify that our counter has incremented.
+    expect(find.text('0'), findsNothing);
+    expect(find.text('1'), findsOneWidget);
+  });
```
chalin: Given the speed of tests, the 5 sec outside interaction hasn't happened yet.

Maybe we could add a delay here and check that the counter gets updated from out-of-band?

scoped_model_counter/lib/main.dart
```diff
@@ -0,0 +1,80 @@
...
+    );
+  }
+}
+
+class MyHomePage extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text('Flutter Demo Home Page'),
```
chalin: Did you want to minimize diffs with the base counter app? If so, consider using `widget.title`.

scoped_model_counter/README.md
```diff
@@ -0,0 +1,26 @@
+# ScopedModel Counter
+
+The starter Flutter application, but using ScopedModel to manage state.
+
+This app is a direct counterpart to the [simple counter application][] that you get when you create
```
RedBrogdon: We've done all the other URLs in these files as inlines, which I'd prefer to maintain for consistency's sake if nothing else.

scoped_model_counter/lib/main.dart
```diff
@@ -0,0 +1,80 @@
...
+      ),
+      body: Center(
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            Text('You have pushed the button this many times:'),
+            // Use the model and rebuild this part of tree when it changes.
```
RedBrogdon: I would consider expanding this comment a little bit, since it's such a critical piece of the code for people to understand:

"ScopedModelDescendant looks for an ancestor `ScopedModel` widget and retrieves its model (`Counter`, in this case). Then it uses that model to build widgets, and will trigger rebuilds if the model is updated."

scoped_model_counter/lib/main.dart
```diff
@@ -0,0 +1,80 @@
...
+              builder: (context, child, counter) => Text(
+                    '${counter.value}',
+                    style: Theme.of(context).textTheme.display1,
+                  ),
+            ),
+          ],
+        ),
+      ),
+      floatingActionButton: FloatingActionButton(
+        // Access the model. This won't rebuild anything here when it changes.
```
RedBrogdon: Consider expanding this one as well to something like:

"`ScopedModel.of` is another way to access the model object held by an ancestor `ScopedModel`. By default, it just returns the current model and doesn't automatically trigger rebuilds. Since this button always looks the same, though, no rebuilds are needed".

scoped_model_counter/test/widget_test.dart
RedBrogdon: +1 to this, mainly so I can see another example of how to do async properly in tests. 😄

scoped_model_counter/README.md
filiph: Done.

scoped_model_counter/lib/main.dart
filiph: Done.

scoped_model_counter/test/widget_test.dart
filiph: The outside interaction isn't tested here (it's not set up, we're just pumping the site).

The reason for that is that I would have to extract the outside interaction to something that could be added here. Which adds complexity to an otherwise simple example. 

If I don't extract, then I need to duplicate the timer code inside the test, which defeats the purpose of the test.

Happy to add an async sample to some other test in this repo.

scoped_model_counter/lib/main.dart
filiph: Done.

scoped_model_counter/lib/main.dart
filiph: Done.

filiph:
Thank you both! I implemented most of the suggestions, and explained which I didn't. Let me know.

scoped_model_counter/pubspec.yaml
filiph: This comes directly from the "new app" template. I'm gonna keep this for now.

scoped_model_counter/test/widget_test.dart
chalin: > The outside interaction isn't tested here (it's not set up, we're just pumping the site).

Ah, right! :( ... hmm. 

It might not be much more complex to factor out the model creation and timer setup code into a function, maybe something like this:

```dart
Widget MyAppWithScoppedModel() {
  // Initialize the model. Can be done outside a widget, like here.
  var counter = Counter();

  // Setup a delayed interaction with the model (increment each 5 seconds),
  // outside of the Flutter widget tree.
  Timer.periodic(
    const Duration(seconds: 5),
    (timer) => counter.increment(),
  );

  // Provide the model to all the app's widgets
  return ScopedModel<Counter>(
    model: counter,
    child: MyApp(),
  );
}

// Run the app using a scoped model:
void main() => runApp(MyAppWithScoppedModel());
```

Then you can `pumpWidget(MyAppWithScoppedModel())`. WDYT?

RedBrogdon: Now that I've had a chance to think more about this, I'm inclined toward Filip's "let's just keep it simple" approach. IMHO, What's being tested here is:

* Whether the app correctly builds and rebuilds widgets based on a ScopedModel.
* Whether the app can correctly update a ScopedModel instance based on a UI event.

I think that covers what's needed. Adding the Timer into the test would get you:

* Whether the Timer code in the app is correctly updating the ScopedModel.
* Whether ScopedModel is correctly notifying its listeners when updated asynchronously by a Timer.

The first of those is within the domain of the app, but not all that interesting. The second is more of a ScopedModel concern than something to specific to this app.

chalin: 👍

PR #42, opened 2018-12-21
Updates travis config to include place_tracker and platform_view_swift

RedBrogdon:
Updates the Travis config for the repo, and makes small changes to place_tracker and platform_view_swift for formatting and analyzer issues.

### DIFF
--- a/place_tracker/lib/app_model.dart
+++ b/place_tracker/lib/app_model.dart
@@ -1,11 +1,8 @@
 import 'package:flutter/material.dart';
 
 class _AppModelScope<T> extends InheritedWidget {
-  const _AppModelScope({
-    Key key,
-    this.appModelState,
-    Widget child
-  }) : super(key: key, child: child);
+  const _AppModelScope({Key key, this.appModelState, Widget child})
+      : super(key: key, child: child);
 
   final _AppModelState<T> appModelState;
 
@@ -18,7 +15,7 @@ class AppModel<T> extends StatefulWidget {
     Key key,
     @required this.initialState,
     this.child,
-  }) : assert(initialState != null),
+  })  : assert(initialState != null),
         super(key: key);
 
   final T initialState;
@@ -30,13 +27,15 @@ class AppModel<T> extends StatefulWidget {
 
   static T of<T>(BuildContext context) {
     final Type appModelScopeType = _typeOf<_AppModelScope<T>>();
-    final _AppModelScope<T> scope = context.inheritFromWidgetOfExactType(appModelScopeType);
+    final _AppModelScope<T> scope =
+        context.inheritFromWidgetOfExactType(appModelScopeType);
     return scope.appModelState.currentState;
   }
 
   static void update<T>(BuildContext context, T newState) {
     final Type appModelScopeType = _typeOf<_AppModelScope<T>>();
-    final _AppModelScope<T> scope = context.inheritFromWidgetOfExactType(appModelScopeType);
+    final _AppModelScope<T> scope =
+        context.inheritFromWidgetOfExactType(appModelScopeType);
     scope.appModelState.updateState(newState);
   }
 }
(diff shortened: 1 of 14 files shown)


place_tracker/lib/place_map.dart
```diff
@@ -152,16 +151,17 @@ class PlaceMapState extends State<PlaceMap> {
     _showPlacesForSelectedCategory(category);
   }
 
-  void _showPlacesForSelectedCategory(PlaceCategory category) async {
-    await _markedPlaces.forEach((Marker marker, Place place) {
-      mapController.updateMarker(
+  Future<void> _showPlacesForSelectedCategory(PlaceCategory category) async {
+    await Future.forEach(_markedPlaces.keys, (Marker marker) async {
```
efortuna: I think a better way to write this is:

await Future.wait(_markedPlaces.forEach(....).toList());

The way in the code currently will wait for each one to return before proceeding to the next one, and I think what you want is to only proceed once all of these have completed, but one doesn't need to finish before the next.

place_tracker/lib/place_map.dart
```diff
@@ -272,34 +270,36 @@ class PlaceMapState extends State<PlaceMap> {
...
+          _configuration.selectedCategory !=
+              newConfiguration.selectedCategory) {
         // If the configuration change is only a category change, just update
         // the marker visibilities.
         _showPlacesForSelectedCategory(newConfiguration.selectedCategory);
       } else {
         // At this point, we know the places have been updated from the list view.
         // We need to reconfigure the map to respect the updates.
-        await newConfiguration.places.forEach((Place value) {
+        await Future.forEach(newConfiguration.places, (Place value) async {
```
efortuna: same here. Future.wait()

place_tracker/lib/place_map.dart
efortuna: and actually, you don't even need the .toList(). Just:

await Future.wait(_markedPlaces.forEach(....));

place_tracker/lib/app_model.dart
```diff
@@ -18,7 +18,7 @@ class AppModel<T> extends StatefulWidget {
     Key key,
     @required this.initialState,
     this.child,
-  }) : assert(initialState != null),
+  })  : assert(initialState != null),
```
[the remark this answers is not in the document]
RedBrogdon: `flutter format` did it, yeah.

place_tracker/lib/place_map.dart
RedBrogdon: Ooh, you're right. I hadn't taken the serial vs parallel part into account.

I don't think the snippet in your last comment will work, though, because Future.wait wants an iterable of Futures, and _markedPlaces.forEach is going to return void (though your Dart-fu obviously exceeds mine).

I'll see if I can combine your stuff and my stuff and get something to do the job.

filiph: I think you want

```dart
Future.wait(_markedPlaces.map((marker, place) => getTheFuture(marker, place)));
```

platform_view_swift/test/widgets_test.dart
```diff
@@ -0,0 +1,11 @@
+// This is a basic Flutter widget test.
+// To perform an interaction with a widget in your test, use the WidgetTester utility that Flutter
+// provides. For example, you can send tap and scroll gestures. You can also use WidgetTester to
+// find child widgets in the widget tree, read text, and verify that the values of widget properties
+// are correct.
+
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  testWidgets('This test always passes', (WidgetTester tester) async {});
```
filiph: :sad filip:

Let's have a fixit one day where we go through all of these tests of samples and actually write them. Just one smoke test per sample is all we need.

RedBrogdon: Adding issues for this now.

PR #43, opened 2018-12-23
Fleshes out settings screen with groups, items, and navigation

RedBrogdon:
Adds refactored versions of SettingsItem and SettingsGroup from @xster's demo, as well as a functional main settings screen and two routes for settings details.

### DIFF
--- a/veggieseasons/lib/screens/settings.dart
+++ b/veggieseasons/lib/screens/settings.dart
@@ -3,19 +3,203 @@
 // found in the LICENSE file.
 
 import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter/widgets.dart';
+import 'package:scoped_model/scoped_model.dart';
+import 'package:veggieseasons/data/preferences.dart';
+import 'package:veggieseasons/data/veggie.dart';
 import 'package:veggieseasons/styles.dart';
+import 'package:veggieseasons/widgets/settings_group.dart';
+import 'package:veggieseasons/widgets/settings_item.dart';
 
-class SettingsScreen extends StatelessWidget {
+class VeggieCategorySettingsScreen extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
+    final model = ScopedModel.of<Preferences>(context, rebuildOnChange: true);
+    final currentPrefs = model.preferredCategories;
+
     return CupertinoPageScaffold(
       navigationBar: CupertinoNavigationBar(
-        middle: Text('Settings'),
+        middle: Text('Preferred Categories'),
+        previousPageTitle: 'Settings',
       ),
       backgroundColor: Styles.scaffoldBackground,
-      child: Center(
-        child: Text('Not yet implemented.'),
+      child: FutureBuilder<Set<VeggieCategory>>(
+        future: currentPrefs,
+        builder: (context, snapshot) {
+          final items = <SettingsItem>[];
+
+          for (final category in VeggieCategory.values) {
+            CupertinoSwitch toggle;
+
+            // It's possible that category data hasn't loaded from shared prefs
+            // yet, so display it if possible, and fall back to disabled
+            // switches otherwise.
+            if (snapshot.hasData) {
+              toggle = CupertinoSwitch(
+                value: snapshot.data.contains(category),
+                onChanged: (isOn) {
+                  if (isOn) {
+                    model.addPreferredCategory(category);
+                  } else {
+                    model.removePreferredCategory(category);
+                  }
+                },
+              );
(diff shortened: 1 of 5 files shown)


xster:
Add a code comment note in the table view related files that they're going to be replaced by actual widgets in Cupertino once https://github.com/flutter/flutter/projects/29 is done.

RedBrogdon:
> Add a code comment note in the table view related files that they're going to be replaced by actual widgets in Cupertino once https://github.com/flutter/flutter/projects/29 is done.

No problem. Which files do you mean, just so I can get them all in one go? Anything that's presenting a large, scrolling list of items like the home, search, and favorite screens?

xster:
settings_group and settings_item

veggieseasons/lib/screens/settings.dart
```diff
@@ -3,19 +3,199 @@
...
+class VeggieCategorySettingsScreen extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    final model = ScopedModel.of<Preferences>(context, rebuildOnChange: true);
+    final currentPrefs = model.preferredCategories;
+
+    return CupertinoPageScaffold(
+      navigationBar: CupertinoNavigationBar(
+        middle: Text('Preferred Categories'),
+        previousPageTitle: 'Settings',
```
xster: You might not need this if your previous CupertinoPageRoute has a `title` field.

veggieseasons/lib/screens/settings.dart
```diff
@@ -3,19 +3,199 @@
...
+                    model.addPreferredCategory(category);
+                  } else {
+                    model.removePreferredCategory(category);
+                  }
+                },
+              );
+            } else {
+              toggle = CupertinoSwitch(
+                value: false,
+                onChanged: (value) {},
```
xster: This is not disabled though right?

veggieseasons/lib/screens/settings.dart
```diff
@@ -3,19 +3,199 @@
...
 
-class SettingsScreen extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
+    final model = ScopedModel.of<Preferences>(context, rebuildOnChange: true);
+
     return CupertinoPageScaffold(
       navigationBar: CupertinoNavigationBar(
-        middle: Text('Settings'),
+        middle: Text('Target Calorie Level'),
```
xster: Seems long (e.g. < Settings   Target Calorie Level). Are you sure this fits on an iPhone SE?

veggieseasons/lib/screens/settings.dart
```diff
@@ -3,19 +3,199 @@
...
+    return SettingsItem(
+      label: 'Target Calorie Intake',
+      icon: SettingsIcon(
+        backgroundColor: Styles.iconBlue,
+        icon: Styles.calorieIcon,
+      ),
+      content: FutureBuilder<int>(
+        future: prefs.desiredCalories,
+        builder: (context, snapshot) => Text(snapshot.data?.toString() ?? ''),
+      ),
```
xster: I may not visualize correctly but all table cell items that navigates needs the > chevron at the end

veggieseasons/lib/screens/settings.dart
RedBrogdon: Coming from Material-based stuff, I didn't realize that property existed. Updated.

veggieseasons/lib/screens/settings.dart
RedBrogdon: It fits on an SE, but looks crowded. Revised to be 'Calorie Target' here and elsewhere.

veggieseasons/lib/screens/settings.dart
RedBrogdon: No. It should be `onChanged: null`.

Out of curiosity, if null is an acceptable value, why are the onChanged callbacks for switches `@required` (both here and in material.dart)?

veggieseasons/lib/screens/settings.dart
RedBrogdon: Added. Should I pull the preview of the selected value as well, or include both?

RedBrogdon:
Comments addressed, also added disclaimers to the settings_group and settings_item files. PTAL!

veggieseasons/lib/screens/settings.dart
xster: Just to make it explicit

veggieseasons/lib/screens/settings.dart
xster: You can have both

xster:
LGTM

PR #47, opened 2019-02-17
Update INDEX.md

gvsharma:
removed 'because' repeated twice

PR #48, opened 2019-02-22
Update INDEX.md

rktv:


RedBrogdon:
It looks like this is a duplicate of #47, so I'm going to go ahead and close it. Thanks for submitting the fix, though, and let us know if you spot anything else!

PR #52, opened 2019-03-02
Adding some lints, and minor cleanup

domesticmouse:


shrine/lib/colors.dart
```diff
@@ -14,15 +14,15 @@
 
 import 'package:flutter/material.dart';
 
-const kShrinePink50 = const Color(0xFFFEEAE6);
-const kShrinePink100 = const Color(0xFFFEDBD0);
-const kShrinePink300 = const Color(0xFFFBB8AC);
-const kShrinePink400 = const Color(0xFFEAA4A4);
+const kShrinePink50 = Color(0xFFFEEAE6);
```
RedBrogdon: Will Dart still infer these to be constants?

veggieseasons/lib/widgets/settings_group.dart
```diff
@@ -69,7 +69,7 @@ class SettingsGroup extends StatelessWidget {
     this.header,
     this.footer,
   })  : assert(items != null),
-        assert(items.length > 0);
+        assert(items.isNotEmpty);
```
RedBrogdon: Do people use these properties (`isEmpty`, `isNotEmpty`) consistently? I have nothing more than personal taste to back up my opinion, but for some reason I don't like them.

shrine/lib/colors.dart
domesticmouse: The second `const` is not needed in Dart 2 IIUC, and this change is encouraged by the `unnecessary_const` lint check.

veggieseasons/lib/widgets/settings_group.dart
domesticmouse: I prefer them as the code reads better, less decoding overhead. As to usage, the `prefer_is_empty` and `prefer_is_not_empty` lint checks will educate and encourage developers. This is why I'm on my add lint checks to everything spree, I witnessed how much value the Go community got out of having one blessed code style.

willlarche:
@RedBrogdon , if this is not materially different from what's in the flutter gallery, we should consider removing it here and bringing it down to 2 places.

domesticmouse:
Are we going to land this PR, or should I close it out?

PR #57, opened 2019-03-14
Adding "no results" display to search screen.

RedBrogdon:
Also renames some properties to bring them in line with conventions elsewhere in the app.


xster:
LGTM

PR #59, opened 2019-03-19
Update Place Tracker to use latest version of plugin.

kenzieschmoll:
Updated the Google Maps article this morning to use the latest version of the google_maps_flutter plugin. While I was doing this, wanted to do a quick re-vamp on the sample app as well.

### DIFF
--- a/place_tracker/analysis_options.yaml
+++ b/place_tracker/analysis_options.yaml
@@ -1,0 +1,10 @@
+analyzer:
+  errors:
+    # treat missing required parameters as a warning (not a hint)
+    missing_required_param: warning
+    # treat missing returns as a warning (not a hint)
+    missing_return: warning
+
+linter:
+  rules:
+    - unawaited_futures
--- a/place_tracker/lib/place_map.dart
+++ b/place_tracker/lib/place_map.dart
@@ -41,49 +41,55 @@ class PlaceMapState extends State<PlaceMap> {
     return places.where((Place place) => place.category == category).toList();
   }
 
-  GoogleMapController mapController;
+  Completer<GoogleMapController> mapController = Completer();
+
+  MapType _currentMapType = MapType.normal;
+
+  LatLng _lastMapPosition;
+
   Map<Marker, Place> _markedPlaces = Map<Marker, Place>();
+
+  final Set<Marker> _markers = {};
+
   Marker _pendingMarker;
+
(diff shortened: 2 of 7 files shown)


place_tracker/lib/place_map.dart
```diff
@@ -331,13 +350,13 @@ class PlaceMapState extends State<PlaceMap> {
           children: <Widget>[
             GoogleMap(
               onMapCreated: onMapCreated,
-              options: GoogleMapOptions(
-                trackCameraPosition: true,
-                cameraPosition: CameraPosition(
-                  target: widget.center,
-                  zoom: 11.0,
-                ),
+              initialCameraPosition: CameraPosition(
```
RedBrogdon: This isn't part of the change, but is the Builder above still necessary?

RedBrogdon:
Thanks for doing this so quickly. Just got a couple questions.

place_tracker/lib/place_map.dart
kenzieschmoll: Yes, because we still show the snack bar.

place_tracker/analysis_options.yaml
```diff
@@ -0,0 +1,10 @@
+analyzer:
+  errors:
+    # treat missing required parameters as a warning (not a hint)
+    missing_required_param: warning
+    # treat missing returns as a warning (not a hint)
+    missing_return: warning
+
+linter:
+  rules:
+    - unawaited_futures
```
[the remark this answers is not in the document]
kenzieschmoll: done.

place_tracker/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -160,6 +179,7 @@
 				TargetAttributes = {
 					97C146ED1CF9000F007C117D = {
 						CreatedOnToolsVersion = 7.3.1;
+						DevelopmentTeam = WECUAV287B;
```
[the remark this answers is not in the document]
kenzieschmoll: removed all the workspace files from this pr

place_tracker/lib/place_map.dart
RedBrogdon: I'll take your word for it. Normally when I see this pattern, there's a Scaffold constructor called in between the outer build method and the Builder constructor, which makes the need for the Builder obvious.

PR #60, opened 2019-03-19
Refactors detail view in advance of trivia

RedBrogdon:
This removes the seasons circles and some other kruft from the details view in favor or a more traditional hierarchy. Four new data fields have been added to each veggie record so there's more to display. A segmented control has been added to the details page as well, so users can flip between info and trivia questions.

The trivia bit is forthcoming. If you have an amusing question about artichokes, for example, please suggest it.


RedBrogdon:
Preview of new UI:

preview2

RedBrogdon:
@willlarche, do you mind taking a look at that screenshot and letting me know if I'm using the segmented control in an appropriate manner?

xster:
tangential: it seems kinda weird hierarchically that the save to garden is at the bottom of the screen. Perhaps add it to the top banner's upper right corner and use https://github.com/flutter/flutter/pull/29630?

RedBrogdon:
I can't use anything not in the current stable release, so it'd be awhile before I could import that icon. I'm very open to changing how things are marked as favorites, though. I don't like how that switch looks tacked on at the end.

xster:
Ah ya, that's a good point. Please disregard then.

veggieseasons/lib/screens/details.dart
```diff
@@ -6,68 +6,222 @@ import 'package:flutter/cupertino.dart';
...
+      future: targetCalories,
+      builder: (context, snapshot) {
+        final target = snapshot?.data ?? 2000;
+        final percent = standardPercentage * 2000 ~/ target;
 
-  String get _firstChars {
-    return '${season.toString().substring(7, 8).toUpperCase()}'
-        '${season.toString().substring(8, 9)}';
+        return Text(
+          '$percent% DV',
```
xster: Is DV a common word people know?

veggieseasons/lib/screens/details.dart
```diff
@@ -6,68 +6,222 @@ import 'package:flutter/cupertino.dart';
...
+              style: Styles.detailsServingHeaderText,
             ),
           ),
-        ),
+          Container(
+            decoration: BoxDecoration(
+              border: Border.all(color: Styles.servingInfoBorderColor),
+            ),
+            padding: const EdgeInsets.all(8.0),
+            child: Column(
```
xster: This build is kinda long. Maybe break this into another widget?

willlarche:
LGTM!

veggieseasons/lib/screens/details.dart
RedBrogdon: I've seen it in multiple sources while looking up nutrition facts. I'll update the note below the values to provide a hint, though.

veggieseasons/lib/screens/details.dart
RedBrogdon: Tables add so much rightward drift. :(

I'll break it out.

PR #61, opened 2019-03-20
Adds trivia to Veggie details screen

RedBrogdon:
Adds a finished trivia widget that appears alternately with veggie facts and info on the details page. Right now there are questions for Apples alone. I'll add more questions once we've verified that the UI/UX works this way.


xster:
LG

PR #66, opened 2019-04-08
add profile xcode configuration

xster:
you're gonna need this to profile the app (since you created the project earlier on)


xster:
I don't actually remember why we need this. We don't really use the Xcode Instruments for anything. Flutter tools just now expects each Dart build configuration to have a matching 'Xcode build configuration' on the Xcode side. I don't remember why. 

So here, it's just to use Dart's profiling mode (which is the same as release mode but has minimal tracing stuff and the observatory left in)

PR #67, opened 2019-04-08
Make in-season and not-in-season cards different

xster:
ezgif com-video-to-gif

You'll have to exercise some artistic taste on top of this but it looks like this

Fixes https://github.com/flutter/flutter/issues/30737

Tangentially fixes https://github.com/flutter/flutter/issues/30511

### DIFF
--- a/veggieseasons/lib/screens/list.dart
+++ b/veggieseasons/lib/screens/list.dart
@@ -13,7 +13,11 @@ import 'package:veggieseasons/styles.dart';
 import 'package:veggieseasons/widgets/veggie_card.dart';
 
 class ListScreen extends StatelessWidget {
-  List<Widget> _generateVeggieRows(List<Veggie> veggies, Preferences prefs) {
+  List<Widget> _generateVeggieRows(
+    List<Veggie> veggies,
+    Preferences prefs,
+    { bool inSeason = true }
+  ) {
     final cards = List<Widget>();
 
     for (Veggie veggie in veggies) {
@@ -23,7 +27,7 @@ class ListScreen extends StatelessWidget {
             future: prefs.preferredCategories,
             builder: (context, snapshot) {
               final data = snapshot.data ?? Set<VeggieCategory>();
-              return VeggieCard(veggie, data.contains(veggie.category));
+              return VeggieCard(veggie, inSeason, data.contains(veggie.category));
             }),
       ));
     }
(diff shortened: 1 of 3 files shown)


veggieseasons/lib/screens/list.dart
```diff
@@ -65,12 +69,14 @@ class ListScreen extends StatelessWidget {
...
         );
 
-        rows.addAll(_generateVeggieRows(appState.unavailableVeggies, prefs));
+        rows.addAll(_generateVeggieRows(appState.unavailableVeggies, prefs, inSeason: false));
 
         return DecoratedBox(
           decoration: BoxDecoration(color: Color(0xffffffff)),
-          child: ListView(
-            children: rows,
+          child: CupertinoScrollbar(
```
RedBrogdon: Any reason this isn't in the Cupertino Widget Gallery? I didn't know it existed.

https://flutter.dev/docs/development/ui/widgets/cupertino

veggieseasons/lib/widgets/veggie_card.dart
```diff
@@ -119,10 +123,23 @@ class VeggieCard extends StatelessWidget {
...
+            label: 'A card background featuring ${veggie.name}',
+            child: Container(
+              height: isInSeason ? 350 : 150,
+              decoration: BoxDecoration(
+                image: DecorationImage(
+                  fit: BoxFit.cover,
+                  colorFilter: isInSeason
+                      ? null
+                      // 222222 is a random color that has low color saturation.
+                      : ColorFilter.mode(Color(0xFF222222), BlendMode.saturation),
```
RedBrogdon: I've been sticking random colors like this in `styles.dart`. Actually, ColorFilter has a const constructor, so you can probably just put the whole filter in there.

veggieseasons/lib/screens/list.dart
xster: Oops, I didn't mean to keep this in. We'd need to fix https://github.com/flutter/flutter/issues/13253 first. I'll add it to the catalog tomorrow though.

veggieseasons/lib/widgets/veggie_card.dart
xster: ah sure. Done

xster:
> How does this affect later plans to use the veggie cards as heroes on the way to the detail screen? 

Heros can go from whatever dimension to whatever dimension showing whatever we want in the process. 

> Is there a way to reach into the cards and lerp the saturation as the transition progresses?

Yes, we'd have to do a simple framework tweak on DecorationImage (could be your PR :D) though maybe it'd be cool to leave it desaturated too for off-season veggies.

PR #70, opened 2019-04-19
Small text correction

legalcodes:
See https://www.californiaavocado.com/blog/avocado-fruit-or-vegetable


legalcodes:
That categorization is dead on -- perhaps its stealthy taxonomy should be a part of its description in the next release

PR #72, opened 2019-05-01
Update jsonexample

kevmoo:
- use latest dependencies
- update build.yaml for removed option
- rebuilt generated code

### DIFF
--- a/jsonexample/build.yaml
+++ b/jsonexample/build.yaml
@@ -14,13 +14,3 @@ targets:
       json_serializable|json_serializable:
         generate_for:
         - lib/json_serializable/*.dart
-        options:
-          header: |
-            // Copyright 2018 The Chromium Authors. All rights reserved.
-            // Use of this source code is governed by a BSD-style license that can be
-            // found in the LICENSE file.
-
-            // GENERATED CODE - DO NOT MODIFY BY HAND
-
-
-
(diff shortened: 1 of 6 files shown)


jsonexample/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
 dev_dependencies:
   build_runner: ^1.0.0
   built_value_generator: ^6.1.5
-  json_serializable: ^1.0.0
+  json_serializable: ^2.2.1
```
RedBrogdon: Whoa, yeah, thanks for catching this. I need to figure out how to hook into pub's "There's a new version you're not getting" warnings.

jsonexample/build.yaml
```diff
@@ -14,13 +14,3 @@ targets:
       json_serializable|json_serializable:
         generate_for:
         - lib/json_serializable/*.dart
-        options:
```
RedBrogdon: Did the header option get removed from source_gen/builder_runner?

kevmoo: yup

jsonexample/pubspec.yaml
kevmoo: Yeah, you could look at the output from `pub upgrade` – but flutter hides it TOO well.

kevmoo:
@RedBrogdon – also confirmed I can land PRs here – good to know!

PR #81, opened 2019-05-16
Migrate from scoped_model to provider

filiph:
This is the minimal possible change. It is meant to be readable as a standalone Pull Request.

This does not touch comments. Additionally, it doesn’t use the niceties of provider, like automatic disposal of the model. Lastly, it doesn’t even rename the project from `scoped_model_counter`. All that will come in a follow-up PR.

### DIFF
--- a/scoped_model_counter/test/widget_test.dart
+++ b/scoped_model_counter/test/widget_test.dart
@@ -1,14 +1,14 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
-import 'package:scoped_model/scoped_model.dart';
+import 'package:provider/provider.dart';
 import 'package:scoped_model_counter/main.dart';
 
 void main() {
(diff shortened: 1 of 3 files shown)


scoped_model_counter/test/widget_test.dart
```diff
@@ -1,14 +1,14 @@
...
 import 'package:scoped_model_counter/main.dart';
 
 void main() {
   testWidgets('Counter increments smoke test', (WidgetTester tester) async {
     // Build our app, provide it with a model, and trigger a frame.
     await tester.pumpWidget(
-      ScopedModel(
-        model: Counter(),
+      ChangeNotifierProvider(
+        builder: (_) => Counter(),
```
RedBrogdon: nit: There are some other spots in here where the underscore for "param I don't care about" could be used but isn't (line 67 in `main.dart` for example). I personally prefer to avoid underscores, but really all that matters is that the code picks one way and sticks with it.

PR #82, opened 2019-05-16
Move scoped_model_counter to provider_counter

filiph:
I used `flutter create` anew, so the project structure now has additional 2 `AndroidManifest.xml` files, and some other changes that are now defaults for new Flutter projects. Everything else stayed the same, so from git’s perspective, it’s mostly just moved files.

I did fix `INDEX.md` and `travis_script.sh`.

WIP: This depends on #81 landing.

Once this lands, @filiph to fix the link to scoped_model_counter at https://flutter.dev/docs/development/data-and-backend/state-mgmt/simple. (According to one backlink checker, that's the only site on the web that links directly to https://github.com/flutter/samples/tree/master/scoped_model_counter.

PR #83, opened 2019-05-16
Finish the migration from scoped_model to provider

filiph:
After this change, we’re managing Counter’s lifecycle with ChangeNotifierProvider. All the comments are referring to Provider. And the migration from ScopedModel to Provider is complete.

WIP because this depends on #81 and #82.

This removes the periodic Timer. Not only does that simplify the example and makes it closer to the original, it also prevents a leaking timer (though, in this case, it’s not an issue, since the timer is needed for the whole duration of the app). I experimented with a more robust approach (having or injecting a Timer/StreamController into the Counter, and disposing of it there) but that seemed overly complex for such a simple example. This whole problem will get significantly easier with https://github.com/rrousselGit/provider/issues/46, at which point I could reintroduce this. I will also think about a more complex Provider example, something like the venerable `bloc_complex`, with infinite-scrolling pagination and all that.

### DIFF
--- a/provider_counter/lib/main.dart
+++ b/provider_counter/lib/main.dart
@@ -1,31 +1,29 @@
-import 'dart:async';
-
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 
 void main() {
-  // Initialize the model. Can be done outside a widget, like here.
-  var counter = Counter();
-
-  // Setup a delayed interaction with the model (increment each 5 seconds),
-  // outside of the Flutter widget tree.
-  Timer.periodic(
-    const Duration(seconds: 5),
-    (timer) => counter.increment(),
-  );
-
-  // Now we're ready to run the app...
   runApp(
-    // ... and provide the model to all widgets within.
-    ChangeNotifierProvider<Counter>.value(
-      notifier: counter,
+    // Provide the model to all widgets within the app. We're using
+    // ChangeNotifierProvider because that's a simple way to rebuild
+    // widgets when a model changes. We could also just use
+    // Provider, but then we would have to listen to Counter ourselves.
+    //
+    // Read Provider's docs to learn about all the available providers.
+    ChangeNotifierProvider(
+      // Initialize the model in the builder. That way, Provider
+      // can own Counter's lifecycle, making sure to call `dispose`
+      // when not needed anymore.
+      builder: (context) => Counter(),
       child: MyApp(),
     ),
   );
 }
 
 /// Simplest possible model, with just one field.
-class Counter extends ChangeNotifier {
+///
+/// [ChangeNotifier] is a class in `flutter:foundation`. [Counter] does
+/// _not_ depend on Provider.
+class Counter with ChangeNotifier {
   int value = 0;
 
   void increment() {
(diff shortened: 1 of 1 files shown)


filiph:
Fixes https://github.com/flutter/flutter/issues/32506.

bizz84:
Hi @filiph.

I just landed here while reading about `ProxyProvider`: https://github.com/rrousselGit/provider/issues/46

This looks like an interesting alternative to the basic counter example, and shows well how to use `ChangeNotifier`.

The very same example could be implemented with `ValueNotifier` + `ValueListenableBuilder`. I believe this would require less code as the custom `Counter` class could be replaced with `ValueNotifier<int>`.

Is there any reason this code uses `ChangeNotifier` when `ValueNotifier` may be simpler? However if the point of this was to show how to use `ChangeNotifier` you can ignore my comment.

Many Thanks.

filiph:
Good question. This sample is trying really hard to be as simple as possible, while using a state management approach that will scale beyond one value. See the scoped_model example that this is replacing for comparison.

As you say, if the counter app was an end result, it would be better served by a ValueNotifier. Or, to be honest, I think just setState works okay, given the complete lack of complexity. But we've heard from many newcomers to Flutter that after that, they don't know how to continue. With ChangeNotifier, it's hopefully pretty clear how to add more variables to track. And with Provider, it's hopefully pretty clear how to add more things to provide.

Do you think this needs documenting? I don't want to over-explain things, especially since this is really meant mostly for newcomers who just want to get to coding.

bizz84:
From what I can gather from comments on Reddit, newcomers did get the message that `Provider` is the way to go.

As we know, Provider is a dependency injection system, not a state management technique. But this still doesn't seem very clear to everyone.

I feel that `ValueNotifier` and `ChangeNotifier` are still relatively unknown (certainly when compared to `setState` and `BLoC`, to name two popular alternatives).

So it is very useful to have sample code showing how to use them. Equally important is for people to discover where the sample code is (I didn't know about Flutter samples until yesterday).

I may write an article comparing `setState`, `ValueNotifier`, `ChangeNotifier` and `Bloc`, with emphasis on using the simplest solution that works for various use cases. 

At the same time, having 4 different techniques (or 3 and a half, since ValueNotifier and ChangeNotifier are based on the same concept) may still be overwhelming for people getting started and trying to choose the "right" one.

Anyway, just some thoughts.

bizz84:
@filiph On a separate note, I think it's also useful to show how to use these techniques on production code, and not just samples.

This week I shared this project, which goes on this direction:

https://github.com/bizz84/firebase_auth_demo_flutter

I'm planning a series of state-management articles on the back of this. Would be cool to get your thoughts on this effort ;)

filiph:
Thanks @bizz84! I agree on all fronts. We need more samples, with more variety, and we need to hammer it down that no single approach is there to rule them all. This is a larger discussion than this PR. I'll have a look at your example.

For this particular sample, I'm using the KISS principle. This code is really only about a simple scenario, and there is really only one concept to learn. Many people prefer to learn from samples like that. Some people even like to see sample code that doesn't necessarily compile, but is stripped to the basics of one API/approach. They can always figure out later how to make it compile and run.

Many other people prefer to learn from a full fledged app with many moving parts. Which is where your work comes in. Thanks again for doing it!

filiph:
Sorry for the force-push, that was a stupid mistake!

rrousselGit:
> As we know, Provider is a dependency injection system, not a state management technique.

I'd agree with that a few weeks ago. But with `ProxyProvider`,  `provider` is starting to cross that line. 

I've been working on many variations of the same application using `provider` + something. I tried it on Mobx, Redux, BLoC, flutter_bloc, and ChangeNotifier.

And I realized that `ProxyProvider` forced all of these to have a uni-directional data-flow with a declarative API.

I think that's very good for maintainability! But that means we can't just advertise it as purely dependency injection anymore.

PR #85, opened 2019-05-21
[WIP] Add provider_shopper

filiph:
This is the code for the sample at https://flutter.dev/docs/development/data-and-backend/state-mgmt/simple (when that's rewritten to use `provider` instead of `scoped_model`).


filiph:
This becomes https://github.com/flutter/samples/pull/87 (solely because of name change of the branch from `provider-shopper-final` to `provider-shopper`.

PR #88, opened 2019-06-07
Updates veggieseasons to Dart 2.3, adds some minor UI updates

RedBrogdon:
Now that I'm looking at this, it probably should have been two PRs.

The changes are the addition of UI as Code operators in some collections, the dropping of ".0" on a bunch of doubles, and the inclusion of a few UI tweaks @xster did for the I/O '19 iOS sample.


veggieseasons/lib/screens/list.dart
```diff
@@ -15,65 +15,53 @@ import 'package:veggieseasons/widgets/veggie_card.dart';
...
+                padding: const EdgeInsets.fromLTRB(16, 24, 16, 16),
+                child: Column(
+                  crossAxisAlignment: CrossAxisAlignment.start,
+                  children: [
+                    Text(dateString.toUpperCase(), style: Styles.minorText),
+                    Text('In season today', style: Styles.headlineText),
+                  ],
+                ),
+              ),
+              ..._generateVeggieRows(appState.availableVeggies, prefs),
```
xster: You don't have to do it now but you'll want to change this to ListView.builder at some point

RedBrogdon: No, you're right that I should switch it. I've made the change in a subsequent commit.

Is there an easier way to do a ListView.builder with contents this varied other than manually doing a bunch of index math?

xster: Nothing I can think of unfortunately. Maybe some clever tricks with generators?

PR #89, opened 2019-06-10
Reformats platform_design to match repo conventions, fixes missing icon

RedBrogdon:


platform_design/lib/songs_tab.dart
```diff
@@ -138,7 +139,7 @@ class _SongsTabState extends State<SongsTab> {
         CupertinoSliverNavigationBar(
           trailing: CupertinoButton(
             padding: EdgeInsets.zero,
-            child: Icon(CupertinoIcons.shuffle),
+            child: Icon(CupertinoIcons.shuffle_thick),
```
xster: Can we add a todo here to change this back when ready?

RedBrogdon: Sure thing.

PR #94, opened 2019-06-18
Added the main animation for page 1

bratay:
A smooth animation that will run constantly to show the performance of the first page.


RedBrogdon:
Looks like travis failed due to a formatting issue. You can use the alias I gave you or Cmd-Opt-L to format the files, and push the results.

isolate_example/lib/page_one.dart
```diff
@@ -0,0 +1,101 @@
...
+    with TickerProviderStateMixin {
+  AnimationController _controller;
+  var borderRadius;
+
+  @override
+  void initState() {
+    super.initState();
+
+    _controller =
+        AnimationController(duration: const Duration(seconds: 1), vsync: this)
```
RedBrogdon: Putting a trailing comma after "vsync: this" should cause the formatter to better format this.

isolate_example/lib/page_one.dart
```diff
@@ -0,0 +1,101 @@
...
+    );
+
+    _controller.forward();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return AnimatedBuilder(
+      animation: borderRadius,
+      builder: (BuildContext context, Widget child) {
```
RedBrogdon: In general, we don't type the parameters for anonymous functions like this one. `builder: (context, child) {}` should be enough for the reader.

PR #97, opened 2019-06-25
Secondpage

bratay:
A second page that runs computation on a second isolate that the user can control

### DIFF
--- a/isolate_example/lib/page_two.dart
+++ b/isolate_example/lib/page_two.dart
@@ -19,7 +19,7 @@ import 'dart:math';
 
 class InfiniteProcessPageStarter extends StatelessWidget {
   @override
-  Widget build(BuildContext context) {
+  Widget build(context) {
     return ChangeNotifierProvider(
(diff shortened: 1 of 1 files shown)


isolate_example/lib/page_two.dart
```diff
@@ -40,14 +40,12 @@ class InfiniteProcessPage extends StatelessWidget {
...
+                children: [newButtons(context)],
               ),
               Row(
                 mainAxisAlignment: MainAxisAlignment.center,
                 crossAxisAlignment: CrossAxisAlignment.center,
-                children: <Widget>[
+                children: [
                   Switch(
                     value:
                         !(Provider.of<IsolateController>(context, listen: true)
```
RedBrogdon: This code would be tidier if there was a single call to get the provider at the top of the build method. Something like:

```
final controller = Provider.of<IsolateController>(context, listen: true);
```

Then you can set `value` and `onChanged` here using `!controller.paused` and `controller.pausedSwitch()`

PR #98, opened 2019-06-26
Adds analysis options to Veggie Seasons

RedBrogdon:
We don't have a standard analysis options file for the repo, and I'd like to introduce one. I'm using Veggie Seasons as a test case, with the plan of replicating the change to all the other apps once we've agreed on a set of linter rules.

All I've added so far is the list from pedantic and a handful of additional rules that I think we can all agree on.

@domesticmouse, @johnpryan: Are there any other rules you'd like in here?


veggieseasons/analysis_options.yaml
```diff
@@ -0,0 +1,9 @@
+include: package:pedantic/analysis_options.yaml
+
+linter:
+  rules:
+    - avoid_types_on_closure_parameters
+    - camel_case_types
+    - constant_identifier_names
+    - non_constant_identifier_names
+    - unnecessary_new
```
johnpryan: I think we could also consider adding the additional Flutter lints marked as such on http://dart-lang.github.io/linter/lints/: 

- cancel_subscriptions
- close_sinks
- control_flow_in_finally
- empty_statements
- hash_and_equals
- test_types_in_equals
- throw_in_finally
- unnecessary_statements
- await_only_futures
- implementation_imports
- package_api_docs
- package_prefixed_library_names
- unnecessary_brace_in_string_interps
- unnecessary_getters_setters
- package_names

RedBrogdon: That's a longer list than I was hoping to add, but I don't see anything in there I disagree with. @domesticmouse, thoughts?

RedBrogdon: Based on a conversation I just had with @bratay, I'd like to nominate avoid_void_async as well:

https://dart-lang.github.io/linter/lints/avoid_void_async.html

domesticmouse: I'm all for adding more lints, as it makes our code more consistent and predictable. I've spent enough time in go-land to enjoy the predictability that a strict compiler and formatter can bring to the community's code.

veggieseasons/analysis_options.yaml
```diff
@@ -0,0 +1,9 @@
+include: package:pedantic/analysis_options.yaml
```
domesticmouse: Please change this to include: package:pedantic/analysis_options.1.7.0.yaml

That means we can adopt newer versions of pedantic's lints on our time schedule, not when new versions of pedantic drop. This will stop false positive travis failures.

veggieseasons/pubspec.yaml
```diff
@@ -20,6 +20,7 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
 
+  pedantic: 1.5.0
```
domesticmouse: We can take this to 1.7.0 once the next version of Flutter drops, yeah?

veggieseasons/analysis_options.yaml
domesticmouse: Er, I've spent too long in dart land. Closing this out

veggieseasons/pubspec.yaml
domesticmouse: And nope. 

/me does sadface

veggieseasons/analysis_options.yaml
```diff
@@ -0,0 +1,9 @@
...
+
+linter:
```
domesticmouse: Consider adding:

analyzer:
  strong-mode:
    implicit-casts: false

https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks

RedBrogdon: Done. This has caused the most actual changes to the code, as it turns out.

RedBrogdon:
I've updated the options with everything that's been suggested at this point. If no one has any objections I'll merge it on Monday, so speak now or forever hold your peace. :)

PR #99, opened 2019-06-28
flutter pub upgrade && flutter format

domesticmouse:
for pubspec in `find . -name pubspec.yaml`
  do (cd `dirname $pubspec` &&
        flutter pub upgrade)
done
flutter format .


RedBrogdon:
Dude, I love the idea behind this and really appreciate that you took the time, but this PR is so large as to be unreviewable. 😄 

Our tests aren't sophisticated that we can trust them to check all of this (the ChromeOS sample and web samples aren't even included in travis right now), and I'm assuming you didn't have time to test each of the apps manually to verify that they compile and run correctly?

How about this as a strategy:

- Finish working out the analyzer options for the repo.
- Go one-by-one and update each of the non-web samples to match those options, bump everything in their pubspecs and so on. Should be easy to do a couple apps at a time in a single PR.
- Check with @kevmoo about what to do with the Flutter for web samples. Kevin, are the samples mirrored anywhere else at this point? If not, are you cool with us updating these to match our style/analyzer options for the repo and pushing new versions with your peanut script?

domesticmouse:
The vast majority of the delta is removed `new` keywords 🤣

domesticmouse:
I can split this PR into per sample PRs easily enough. Would that help?

RedBrogdon:
> I can split this PR into per sample PRs easily enough. Would that help?

Yeah, that'd be great. We've also got the analysis options worked out now, so they can be replicated across the mobile samples.

PR #100, opened 2019-06-28
Secondpage Add-ons

bratay:
Some extra add-ons for the second page mostly for looks


isolate_example/lib/page_two.dart
```diff
@@ -35,6 +35,13 @@ class InfiniteProcessPage extends StatelessWidget {
     return Column(
       mainAxisSize: MainAxisSize.min,
       children: [
+        Container(
+          child: Text(
+            'Summation Results',
+            style: TextStyle(fontSize: 25),
```
RedBrogdon: Odds are that you can find a text style in the app's existing theme that will work here:

```
style: Theme.of(context).textTheme.pickAFontStyleFromTheListHere
```

If you can't find one that's exactly what you want, you can modify it with `copyWith`, which should be available on all the text styles in the theme.

isolate_example/lib/page_two.dart
```diff
@@ -35,6 +35,13 @@ class InfiniteProcessPage extends StatelessWidget {
...
+          ),
+          padding: new EdgeInsets.all(8),
```
RedBrogdon: No need to use a `Container` if all you're looking for is Padding. Just wrap the `Text` in a `Padding` widget.

isolate_example/lib/page_two.dart
```diff
@@ -157,17 +164,28 @@ class RunningList extends StatelessWidget {
...
       ),
       child: new ListView.builder(
         itemCount: sums.length,
         itemBuilder: (context, index) {
-          return ListTile(
-            title: Text((sums.length - index).toString() +
-                '. ' +
-                sums[index].toString()),
+          return Column(
+            children: <Widget>[
```
RedBrogdon: Don't need the `&lt;Widget&gt;` here, since the list has widgets of different types.

isolate_example/lib/page_two.dart
```diff
@@ -157,17 +164,28 @@ class RunningList extends StatelessWidget {
...
+              Card(
+                child: ListTile(
+                  leading: Text((sums.length - index).toString() + '. '),
```
RedBrogdon: Have we talked about Dart's string interpolation yet?

https://dart.dev/guides/language/language-tour#strings

You can rewrite this as:

```
Text(${sums.length - index}.')
```

The one on line 177 can also be done with interpolation.

isolate_example/lib/page_two.dart
```diff
@@ -157,17 +164,28 @@ class RunningList extends StatelessWidget {
...
+                  title: Text(sums[index].toString()),
+                ),
+                color:
+                    (controller.running == true && controller.paused == false)
```
RedBrogdon: `(controller.running && !controller.paused)`

isolate_example/lib/page_two.dart
```diff
@@ -157,17 +164,28 @@ class RunningList extends StatelessWidget {
...
+                        ? Colors.lightGreenAccent
+                        : Colors.deepOrangeAccent,
+              ),
+              Divider(
```
RedBrogdon: Nice use of `Divider`. Not that many people even know it exists. 😄

isolate_example/lib/page_two.dart
```diff
@@ -157,17 +164,27 @@ class RunningList extends StatelessWidget {
...
-          return ListTile(
-            title: Text((sums.length - index).toString() +
-                '. ' +
-                sums[index].toString()),
+          return Column(
+            children: [
+              Card(
+                child: ListTile(
+                  leading: Text('${sums.length - index}.'),
+                  title: Text(sums[index].toString()),
```
RedBrogdon: This one can be interpolated as well:

`Text('${sums[index]}'),`

All it'd really be doing is calling `toString` for you, but it's generally considered more idiomatic to use interpolation and avoid using `toString`.

PR #104, opened 2019-07-02
Add analysis_options, format and fix up

domesticmouse:
PTAL @RedBrogdon, looking at the deltas in this sample, we may want to back out the `strong-mode` requirements.


RedBrogdon:
These PRs are all great, but don't feel pressured to do all the grunt work yourself. I know you have plenty of other things on your plate right now, though it seems like you really enjoy tidying up old code.

PR #107, opened 2019-07-02
Analysis options, fixes and format

domesticmouse:


platform_design/analysis_options.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - non_constant_identifier_names
+    - package_api_docs
+    - package_names
+    - package_prefixed_library_names
+    - test_types_in_equals
+    - throw_in_finally
+    - unnecessary_brace_in_string_interps
+    - unnecessary_getters_setters
+    - unnecessary_new
+    - unnecessary_statements
```
RedBrogdon: nit: newline

xster:
Can this just be in the root directory or it needs to be a sibling with pubspec?

domesticmouse:
At this point we are rolling out this analysis options across all the samples, but it will take a lot of work to get them all on the same one. There are samples under the web directory from all over the place, and we shouldn't deviate these downstream versions too far from their upstreams, just to make merging sane.

xster:
SG

PR #115, opened 2019-07-09
Updated README.md file

bratay:
Updated README.md

### DIFF
--- a/isolate_example/README.md
+++ b/isolate_example/README.md
@@ -25,10 +25,10 @@ and change how many time the result will be multiplied by.
 
 ### `page_three.dart`
 
-Shows how expensive it is to move large amounts of data between isolates. This will 
-be done by either creating large amounts of randomly generated numbers on the main isolate
-to send to the secondary isolate to be summated or simply generating the number on the 
-second isolate.
+A demonstration of how expensive it is to move large amounts of data between isolates and 
+an alternative using transferableTypeData. You can create large amounts of randomly generated
+numbers on the main isolate to send to the secondary isolate to be summated, simply
+generating the number on the second isolate, or transfer the data using transferableTypeData.
 
 ## Questions/issues
 
--- a/isolate_example/README.md
+++ b/isolate_example/README.md
@@ -9,6 +9,27 @@ A sample application that demonstrate best practices when using [`isolates`](htt
 * Demonstrate how to initialize and use an isolate.
 * Show the cost of moving data between isolates and provide alternatives.
 
+## The important bits
+
+### `page_one.dart`
+
+Compares running a large computation on the main isolate vs on a secondary isolate. 
+The `SmoothAnimationWidget` will lower frames when the performance drops.
+
+### `page_two.dart`
+
+An infinite process will be run on the secondary isolate.  The process will be an 
+endless loop of the summation of 100 million randomly generated numbers in increments
+of 10 with the results being printed. The user can start, terminate, pause, resume,
+and change how many time the result will be multiplied by.
+
+### `page_three.dart`
+
+Shows how expensive it is to move large amounts of data between isolates. This will 
+be done by either creating large amounts of randomly generated numbers on the main isolate
+to send to the secondary isolate to be summated or simply generating the number on the 
+second isolate.
+
 ## Questions/issues
 
 If you have a general question about any of the techniques you see in
--- a/isolate_example/README.md
+++ b/isolate_example/README.md
@@ -13,19 +13,26 @@ A sample application that demonstrate best practices when using [`isolates`](htt
 
 ### `page_one.dart`
 
-Compares running a large computation on the main isolate with running the same calculation on a second isolate.
-When the main isolate is used, Flutter is unable to render new frames, and the `SmoothAnimationWidget`'s animation freezes.
+Compares running a large computation on the main isolate with running the same calculation
+on a second isolate.When the main isolate is used, Flutter is unable to render new frames, and
+the `SmoothAnimationWidget`'s animation freezes.
 ### `page_two.dart`
 
-Creates an isolate used to compute an infinate loop of sum of 100M randomly generated numbers while reporting progress. 
-Users can start, terminate, pause, and resume the isolate, as well as modify how the calculation is performed.
+Creates an isolate used to run an infinite loop that sums batches of 100M randomly generated
+numbers at a time. Users can start, terminate, pause, and resume the isolate, as well as modify
+how the calculation is performed.
 
 ### `page_three.dart`
 
-Demonstration how expensive it is to move large amounts of data between isolates and 
-an alternative using transferableTypeData. You can create large amounts of randomly generated
-numbers on the main isolate to send to the secondary isolate to be summated, simply
-generating the number on the second isolate, or transfer the data using transferableTypeData.
+Demonstrates how expensive it is to move large amounts of data between isolates and 
+better a alternative. This page creates an isolate that can add up a list of numbers and
+gives users three options for how to provide it with input:
+
+* Send values normally using a List
+* Send the values using TransferableTypedData
+* Generate the values on the second isolate, so no copying is necessary
+
+Users can then compare the performance of each approach using the displayed timestamps.
 
 ## Questions/issues
 
--- a/isolate_example/README.md
+++ b/isolate_example/README.md
@@ -14,7 +14,7 @@ A sample application that demonstrate best practices when using [`isolates`](htt
 ### `page_one.dart`
 
 Compares running a large computation on the main isolate with running the same calculation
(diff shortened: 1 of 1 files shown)


isolate_example/README.md
```diff
@@ -9,6 +9,27 @@ A sample application that demonstrate best practices when using [`isolates`](htt
 * Demonstrate how to initialize and use an isolate.
 * Show the cost of moving data between isolates and provide alternatives.
 
+## The important bits
+
+### `page_one.dart`
+
+Compares running a large computation on the main isolate vs on a secondary isolate.
```
johnpryan: nit: consider changing 'secondary' to 'second'

isolate_example/README.md
```diff
@@ -9,6 +9,27 @@ A sample application that demonstrate best practices when using [`isolates`](htt
 * Demonstrate how to initialize and use an isolate.
 * Show the cost of moving data between isolates and provide alternatives.
 
+## The important bits
+
+### `page_one.dart`
+
+Compares running a large computation on the main isolate vs on a secondary isolate. 
+The `SmoothAnimationWidget` will lower frames when the performance drops.
```
johnpryan: Use [present tense](https://developers.google.com/style/tense) here and in the following paragraphs

isolate_example/README.md
```diff
@@ -9,6 +9,27 @@ A sample application that demonstrate best practices when using [`isolates`](htt
...
+
+### `page_two.dart`
+
+An infinite process will be run on the secondary isolate.  The process will be an 
+endless loop of the summation of 100 million randomly generated numbers in increments
+of 10 with the results being printed. The user can start, terminate, pause, resume,
```
johnpryan: consider "that adds 100 million..." or "that sums 100 million"

isolate_example/README.md
```diff
@@ -9,6 +9,27 @@ A sample application that demonstrate best practices when using [`isolates`](htt
...
+and change how many time the result will be multiplied by.
+
+### `page_three.dart`
+
+A demonstration of how expensive it is to move large amounts of data between isolates and
```
johnpryan: consider using [active voice](https://developers.google.com/style/voice) e.g. "This sample moves large amounts of data..."

isolate_example/README.md
RedBrogdon: "Compares running a large computation on the main isolate with running the same calculation on a second isolate."

isolate_example/README.md
RedBrogdon: "When the main isolate is used, Flutter is unable to render new frames, and the `SmoothAnimationWidget`'s animation freezes."

isolate_example/README.md
```diff
@@ -9,6 +9,27 @@ A sample application that demonstrate best practices when using [`isolates`](htt
...
+## The important bits
+
+### `page_one.dart`
+
+Compares running a large computation on the main isolate vs on a secondary isolate. 
+The `SmoothAnimationWidget` will lower frames when the performance drops.
+
+### `page_two.dart`
+
+An infinite process will be run on the secondary isolate.  The process will be an
```
RedBrogdon: Technically the process isn't infinite, right? Consider:

"Creates an isolate and uses it to compute the sum of 100M randomly generated numbers while reporting progress. Users can start, terminate, pause, and resume the isolate, as well as modify how the calculation is performed."

isolate_example/README.md
RedBrogdon: +1. For parallelism's sake, I'd recommend starting this off with the verb ("Demonstrates the expense of moving large...")

isolate_example/README.md
```diff
@@ -9,6 +9,24 @@ A sample application that demonstrate best practices when using [`isolates`](htt
 * Demonstrate how to initialize and use an isolate.
 * Show the cost of moving data between isolates and provide alternatives.
 
+## The important bits
+
+### `page_one.dart`
+
+Compares running a large computation on the main isolate with running the same calculation on a second isolate.
+When the main isolate is used, Flutter is unable to render new frames, and the `SmoothAnimationWidget`'s animation freezes.
```
RedBrogdon: nit: Need a line break between 17 & 18

isolate_example/README.md
```diff
@@ -9,6 +9,24 @@ A sample application that demonstrate best practices when using [`isolates`](htt
...
+### `page_two.dart`
+
+Creates an isolate used to compute an infinate loop of sum of 100M randomly generated numbers while reporting progress.
```
RedBrogdon: Is "infinite loop" accurate here? It's a long-running loop, certainly, but it does eventually exit.

Consider "...used to compute the sum of 100M..." instead.

isolate_example/README.md
```diff
@@ -9,6 +9,24 @@ A sample application that demonstrate best practices when using [`isolates`](htt
...
+Compares running a large computation on the main isolate with running the same calculation on a second isolate.
+When the main isolate is used, Flutter is unable to render new frames, and the `SmoothAnimationWidget`'s animation freezes.
+### `page_two.dart`
+
+Creates an isolate used to compute an infinate loop of sum of 100M randomly generated numbers while reporting progress. 
+Users can start, terminate, pause, and resume the isolate, as well as modify how the calculation is performed.
+
+### `page_three.dart`
+
+Demonstration how expensive it is to move large amounts of data between isolates and
```
RedBrogdon: The other two page descriptions begin with present tense verbs ("Compares," "Creates"), so for parallelism's sake this one should be changed to "demonstrates" (or the other two should be changed).

isolate_example/README.md
```diff
@@ -9,6 +9,24 @@ A sample application that demonstrate best practices when using [`isolates`](htt
...
+When the main isolate is used, Flutter is unable to render new frames, and the `SmoothAnimationWidget`'s animation freezes.
+### `page_two.dart`
+
+Creates an isolate used to compute an infinate loop of sum of 100M randomly generated numbers while reporting progress. 
+Users can start, terminate, pause, and resume the isolate, as well as modify how the calculation is performed.
+
+### `page_three.dart`
+
+Demonstration how expensive it is to move large amounts of data between isolates and 
+an alternative using transferableTypeData. You can create large amounts of randomly generated
```
RedBrogdon: Because the text is presenting three alternatives (copy numbers over with a typical list, copy them using TransferableTypedData, or generate them on the second isolate), it's probably better just to list them out plainly. Something like:

Demonstrates how expensive it is to move large amounts of data between isolates. This page creates an isolate that can add up a list of numbers and gives users three options for how to provide it with input:

* Send values using a List<int>
* Send the values using TransferableTypedData
* Generate the values on the second isolate, so no copying is necessary

Users can then compare the performance of each approach.

isolate_example/README.md
```diff
@@ -19,4 +37,4 @@ the sample, the best places to go are:
 * [StackOverflow](https://stackoverflow.com/questions/tagged/flutter)
 
 If you run into an issue with the sample itself, please file an issue
-in the [main Flutter repo](https://github.com/flutter/flutter/issues).
\ No newline at end of file
+in the [main Flutter repo](https://github.com/flutter/flutter/issues).
```
RedBrogdon: Not sure why this line is lit up as a change. It may have lost or gained a line break.

isolate_example/README.md
bratay: I believe that infinite loop is accurate since it will run forever unless the user presses terminate

RedBrogdon: Discussed offline, and "infinite loop" is indeed more accurate.

isolate_example/README.md
```diff
@@ -9,6 +9,31 @@ A sample application that demonstrate best practices when using [`isolates`](htt
 * Demonstrate how to initialize and use an isolate.
 * Show the cost of moving data between isolates and provide alternatives.
 
+## The important bits
+
+### `page_one.dart`
+
+Compares running a large computation on the main isolate with running the same calculation
+on a second isolate.When the main isolate is used, Flutter is unable to render new frames, and
```
RedBrogdon: Need a space before "when."

isolate_example/README.md
```diff
@@ -9,6 +9,31 @@ A sample application that demonstrate best practices when using [`isolates`](htt
...
+### `page_two.dart`
+
+Creates an isolate used to run an infinite loop that sums batches of 100M randomly generated
+numbers at a time. Users can start, terminate, pause, and resume the isolate, as well as modify
+how the calculation is performed.
+
+### `page_three.dart`
+
+Demonstrates how expensive it is to move large amounts of data between isolates and 
+better a alternative. This page creates an isolate that can add up a list of numbers and
```
RedBrogdon: It looks like there's a cut-and-paste error in here or something: "...between isolates and better a alternative" doesn't parse grammatically.

PR #116, opened 2019-07-09
TransferableTypeData added

bratay:
New feature added button for sending numbers with TransferableTypeData!

### DIFF
--- a/isolate_example/lib/page_three.dart
+++ b/isolate_example/lib/page_three.dart
@@ -11,6 +11,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+import 'dart:typed_data';
+
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 import 'dart:isolate';
@@ -105,23 +107,24 @@ class DataTransferIsolateController extends ChangeNotifier {
     if (running) return;
 
     running = true;
-    _timer.reset();
     currentProgress.clear();
 
+    _timer = Stopwatch();
     _timer.start();
     _newIceSP.send('start');
 
     notifyListeners();
   }
 
-  Future<void> generateRandomNumbers() async {
+  Future<void> generateRandomNumbers(bool ttd) async {
     if (running) return;
 
     running = true;
     Random rng = Random();
-    _timer = Stopwatch();
 
     currentProgress.clear();
+
+    _timer.reset();
     _timer.start();
 
     List<int> randNums = [];
@@ -131,11 +134,18 @@ class DataTransferIsolateController extends ChangeNotifier {
       for (int j = 0; j < 1000000; j++) {
         randNums.add(rng.nextInt(100));
       }
-      await sendNumbers(randNums);
+
+      if (ttd) {
+        final transferable =
+            TransferableTypedData.fromList([Int32List.fromList(randNums)]);
+        await sendNumbers(transferable);
+      } else {
+        await sendNumbers(randNums);
+      }
     }
   }
 
-  Future<void> sendNumbers(List<int> numList) async {
+  Future<void> sendNumbers(dynamic numList) async {
     return Future<void>(() {
       _newIceSP.send(numList);
     });
@@ -186,17 +196,25 @@ class RunningList extends StatelessWidget {
 Future<void> _secondIsolateEntryPoint(SendPort callerSP) async {
   ReceivePort newIceRP = ReceivePort();
   callerSP.send(newIceRP.sendPort);
+  int length = 1;
 
   newIceRP.listen(
     (dynamic message) async {
       if (message is String && message == 'start') {
-        await generateAndSum(callerSP, createNums());
+        await generateAndSum(callerSP, createNums(), length);
 
         callerSP.send('done');
-      } else {
-        await generateAndSum(callerSP, message as List<int>);
+      } else if (message is TransferableTypedData) {
+        await generateAndSum(callerSP, message.materialize().asInt32List(), length);
+        length++;
+      } else if (message is List<int>) {
+        await generateAndSum(callerSP, message, length);
+        length++;
+      }
 
+      if (length == 101) {
         callerSP.send('done');
+        length = 1;
       }
     },
   );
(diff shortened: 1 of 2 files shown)


isolate_example/lib/page_three.dart
```diff
@@ -105,23 +107,24 @@ class DataTransferIsolateController extends ChangeNotifier {
...
 
+    _timer = Stopwatch();
     _timer.start();
     _newIceSP.send('start');
 
     notifyListeners();
   }
 
-  Future<void> generateRandomNumbers() async {
+  Future<void> generateRandomNumbers(bool ttd) async {
```
Piinks: What is `ttd` represent here?

isolate_example/lib/page_three.dart
```diff
@@ -131,11 +134,18 @@ class DataTransferIsolateController extends ChangeNotifier {
       for (int j = 0; j < 1000000; j++) {
         randNums.add(rng.nextInt(100));
       }
-      await sendNumbers(randNums);
+
+      if (ttd) {
```
Piinks: Same comment for `ttd` here

isolate_example/lib/page_three.dart
```diff
@@ -131,11 +134,18 @@ class DataTransferIsolateController extends ChangeNotifier {
...
+        final transferable =
```
Piinks: Nit: This could be one line

isolate_example/lib/page_three.dart
```diff
@@ -186,17 +196,26 @@ class RunningList extends StatelessWidget {
...
     (dynamic message) async {
       if (message is String && message == 'start') {
-        await generateAndSum(callerSP, createNums());
+        await generateAndSum(callerSP, createNums(), length);
 
         callerSP.send('done');
-      } else {
-        await generateAndSum(callerSP, message as List<int>);
+      } else if (message is TransferableTypedData) {
+        await generateAndSum(
```
Piinks: Nit: this could also be one line, or listed parameters on individual lines?

isolate_example/lib/page_three.dart
```diff
@@ -235,7 +255,12 @@ Widget createButtons(BuildContext context) {
           RaisedButton(
             child: const Text('Transfer Data to 2nd Isolate'),
             elevation: 8.0,
-            onPressed: controller.generateRandomNumbers,
+            onPressed: () => controller.generateRandomNumbers(false),
+          ),
+          RaisedButton(
```
Piinks: It might be nice to provide some feedback to the user when they tap a button. 
When I ran this, I wasn't sure which one I had pressed - this could be informed by highlighting that button, or displaying text that indicates what is being executed.
It also seemed that the other buttons could not be pressed while a process is running. It might be a good idea to disable them if that is the case.

isolate_example/lib/page_three.dart
bratay: I agree but the Travis Script only passes if I use auto formatting which makes this two lines.

bratay:
PTAL

isolate_example/lib/page_three.dart
Piinks: Ah I see! I heard about this offline. For the future, if I recommend something that Travis disagrees with, you can ignore me. :) This repository and the flutter/flutter repository use different style guides.

isolate_example/lib/page_three.dart
Piinks: If Travis Script has a differing opinion, ignore this nit.

isolate_example/lib/page_three.dart
```diff
@@ -11,6 +11,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+import 'dart:typed_data';
```
RedBrogdon: There are `dart` imports in two places in this file, and they should be grouped together. Ctrl-Opt-O should sort them for you in Android Studio

isolate_example/lib/page_three.dart
```diff
@@ -67,6 +69,7 @@ class DataTransferIsolateController extends ChangeNotifier {
   SendPort _newIceSP;
 
   final currentProgress = <String>[];
+  List<bool> runningTest = [false, false, false];
```
RedBrogdon: If you're only ever running one test at a time, have you considered simply storing the ID of the running test, rather than a list of bools? You could then zero out the "ID of the running test" when nothing was running and avoid keeping a separate `running` property.

isolate_example/lib/page_three.dart
```diff
@@ -103,25 +106,27 @@ class DataTransferIsolateController extends ChangeNotifier {
...
     _timer.start();
     _newIceSP.send('start');
 
     notifyListeners();
   }
 
-  Future<void> generateRandomNumbers() async {
+  Future<void> generateRandomNumbers(bool transferableTyped) async {
     if (running) return;
+    (transferableTyped) ? setRunningTest(1) : setRunningTest(0);
```
RedBrogdon: I don't think we have a lint for this, but I *much* prefer avoiding ternary expressions as if they were statements and instead using if-thens for logic like this.

RedBrogdon:
Put a PTAL on this when you're ready for another round.

bratay:
@RedBrogdon PTAL thanks!

isolate_example/lib/page_three.dart
RedBrogdon: It looks like this is still an issue in the most recent revision. The `dart:` imports should be grouped together at the top.

isolate_example/lib/page_three.dart
RedBrogdon: @bratay It's always good to either respond to a comment or mark it resolved when you make the appropriate code change. 😄

isolate_example/lib/page_three.dart
```diff
@@ -103,25 +105,32 @@ class DataTransferIsolateController extends ChangeNotifier {
 
   void generateOnSecondaryIsolate() {
     if (running) return;
-
+    _runningTest = 3;
```
RedBrogdon: I think there's been a mixup here. The array of bools is gone, which is great, but there are now three different properties tracking whether a test is running: `_runningTest`, `runningTest`, and `running`.

It looks like `runningTest` is being used to guard access to the Controller's field, which is a legit reason to use a getter. `running` doesn't need to be tracked separately, though. It could either disappear (and consumers could check whether `runningTest` was zero, or it could be a getter that did the check itself.

bratay:
@RedBrogdon PTAL thanks

PR #118, opened 2019-07-12
Add animation samples

johnpryan:
Adds a new app `animations` with just two samples to keep the code reviews small. The samples are separated into two groups; `basics` and `misc`.


animations/lib/main.dart
```diff
@@ -0,0 +1,84 @@
...
+
+class Demo {
+  final String name;
+  final String route;
+  final WidgetBuilder builder;
+
+  Demo(this.name, this.route, this.builder);
+}
+
+List<Demo> basicDemos = [
```
RedBrogdon: nit: `final` is more common in the repo than an explicit type for simple things like these lists:

`final basicDemos = [...]`

For the maps, the function that controls how their types are ultimately inferred is complex-looking enough that an explicit type can help with readability (though they could be final as well, I believe).

animations/lib/main.dart
```diff
@@ -0,0 +1,84 @@
...
+];
+
+Map<String, WidgetBuilder> basicDemoRoutes = Map.fromEntries(
+    basicDemos.map((d) => MapEntry(d.route, d.builder)));
+
+Map<String, WidgetBuilder> miscDemoRoutes = Map.fromEntries(
+    miscDemos.map((d) => MapEntry(d.route, d.builder)));
+
+Map<String, WidgetBuilder> allRoutes = <String, WidgetBuilder>{
+  ...basicDemoRoutes,
```
RedBrogdon: Never used the spread operator with a Map before. I'm slightly jealous. 😄

animations/lib/main.dart
```diff
@@ -0,0 +1,84 @@
...
+  ...miscDemoRoutes,
+};
+
+class AnimationsSamples extends StatelessWidget {
```
RedBrogdon: nit: "AnimationSamples"?

animations/lib/main.dart
```diff
@@ -0,0 +1,84 @@
...
+        primarySwatch: Colors.deepPurple,
+      ),
+      routes: allRoutes,
+      home: HomePage(),
+    );
+  }
+}
+
+class HomePage extends StatelessWidget {
+  final TextStyle _headerStyle =
```
RedBrogdon: Is it possible to use one of the existing Material text styles from Theme.of(context).textTheme to get something suitable?

animations/lib/main.dart
```diff
@@ -0,0 +1,84 @@
...
+      TextStyle(fontWeight: FontWeight.bold, fontSize: 24);
+
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text('Animation Examples'),
+      ),
+      body: ListView(
+        children: <Widget>[
```
RedBrogdon: Shouldn't need `<Widget>` here.

animations/lib/src/basics/animation_controller_demo.dart
```diff
@@ -0,0 +1,61 @@
...
+  static const String routeName = '/basics/animation_controller';
+
+  _AnimationControllerDemoState createState() =>
+      _AnimationControllerDemoState();
+}
+
+class _AnimationControllerDemoState extends State<AnimationControllerDemo>
+    with SingleTickerProviderStateMixin {
+  static const Duration _duration = Duration(seconds: 1);
+  AnimationController controller;
```
RedBrogdon: Probably better to start both properties with a leading underscore or neither.

animations/lib/src/basics/animation_controller_demo.dart
```diff
@@ -0,0 +1,61 @@
...
+    controller.dispose();
+  }
+
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(),
+      body: Center(
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: <Widget>[
```
RedBrogdon: No '<Widget>` needed. I really wish the IDE plugins didn't insert it.

animations/lib/src/basics/animation_controller_demo.dart
```diff
@@ -0,0 +1,61 @@
...
+            ConstrainedBox(
+              constraints: BoxConstraints(maxWidth: 200),
+              child: Text('${controller.value.toStringAsFixed(2)}',
+                  style: TextStyle(fontSize: 24)),
```
RedBrogdon: Unless you've already tried, see if you can use a built-in TextStyle here.

animations/lib/src/basics/animation_controller_demo.dart
```diff
@@ -0,0 +1,61 @@
...
+            ),
+            MaterialButton(
```
RedBrogdon: Would RaisedButton do what you want here? There's also a ButtonTheme in the app theme you could manipulate the color.

animations/lib/src/basics/animation_controller_demo.dart
```diff
@@ -0,0 +1,61 @@
...
+              color: Theme.of(context).primaryColor,
+              child: Text(
+                'animate',
+                style: TextStyle(color: Colors.white),
```
RedBrogdon: Better to pick one of the TextStyles in the material ThemeData and then use copyWith to modify the color. That gets you a definite style as a base.

animations/lib/src/basics/animation_controller_demo.dart
```diff
@@ -0,0 +1,61 @@
...
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(),
+      body: Center(
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: <Widget>[
+            ConstrainedBox(
+              constraints: BoxConstraints(maxWidth: 200),
+              child: Text('${controller.value.toStringAsFixed(2)}',
```
RedBrogdon: Outputting the value as the most basic example (rather than animating something) is smart. Everyone should be able to get what's going on here.

animations/lib/src/misc/expand_card.dart
```diff
@@ -0,0 +1,68 @@
+import 'package:flutter/material.dart';
+
+class ExpandCardDemo extends StatelessWidget {
+  static String routeName = '/expand_card';
```
RedBrogdon: nit: const?

animations/lib/src/misc/expand_card.dart
```diff
@@ -0,0 +1,68 @@
...
+}
+
+class _ExpandCardState extends State<ExpandCard>
+    with SingleTickerProviderStateMixin {
+  static const Duration duration = Duration(milliseconds: 300);
+  bool expanded = false;
+
+  double get size => expanded ? 200 : 100;
+
+  void initState() {
```
RedBrogdon: Cut this method?

animations/lib/src/misc/expand_card.dart
```diff
@@ -0,0 +1,68 @@
...
+  Widget build(context) {
+    return GestureDetector(
+      onTap: () => toggleExpanded(),
+      child: Card(
+        child: AnimatedCrossFade(
+          alignment: Alignment.center,
+          duration: duration,
+          crossFadeState:
+              expanded ? CrossFadeState.showSecond : CrossFadeState.showFirst,
+          firstChild: AnimatedContainer(
```
RedBrogdon: Could (and I'm really asking here, I haven't checked) the AnimatedContainer be moved up the tree to be the parent of AnimatedCrossFade (which could then have the Images as its direct descendants? If it's just there to control the size of the card, you might not need two of them this low.

animations/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: animations
+description: A new Flutter project.
+version: 1.0.0+1
+
+environment:
+  sdk: ">=2.3.0 <3.0.0"
```
RedBrogdon: nit: 2.4.0

animations/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+  flutter:
+    sdk: flutter
+  cupertino_icons: ^0.1.2
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+flutter:
+  uses-material-design: true
+  assets:
+    - assets/
```
RedBrogdon: Where did the wolf and cat images come from? Do we hold the copyright for them?

animations/README.md
```diff
@@ -0,0 +1,12 @@
+# Animation Samples
+Sample apps that showcasing Flutter's animation features
+
+## Basics
+**AnimationControllerDemo**
+Demonstrates how to use an `AnimationController`.
+
+
+## Misc
```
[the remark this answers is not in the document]
johnpryan: Hmm... yeah maybe "simple" and "advanced" makes more sense for now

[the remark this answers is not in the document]
johnpryan: The "basics" was meant to be specifically for learning the basics. "misc" was for everything else... I wouldn't call the expand-card demo advanced though.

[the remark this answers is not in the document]
johnpryan: I think we will have more than just two categories though

animations/lib/main.dart
johnpryan: 😎

animations/lib/src/basics/animation_controller_demo.dart
johnpryan: 🤓

animations/README.md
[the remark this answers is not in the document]
lucalves: I believe jumping just one line would be ideal.

animations/README.md
```diff
@@ -0,0 +1,12 @@
+# Animation Samples
+Sample apps that showcasing Flutter's animation features
+
+## Basics
+**AnimationControllerDemo**
+Demonstrates how to use an `AnimationController`.
+
```
lucalves: I believe jumping just one line here would be ideal.

animations/pubspec.yaml
johnpryan: I got them both from Pixabay:

https://pixabay.com/photos/wolf-head-profile-predator-2878633/
https://pixabay.com/photos/cat-kitty-pet-feline-animal-4269479/

johnpryan:
@RedBrogdon PTAL

animations/pubspec.yaml
RedBrogdon: While you're right that we could probably use those without issue, it'd be better to use something we know 100%, no-doubt-about-it belongs to us. I've emailed you a link to a Drive folder with images to which we own the copyright.

animations/lib/main.dart
```diff
@@ -0,0 +1,83 @@
...
+
+final allRoutes = <String, WidgetBuilder>{
+  ...basicDemoRoutes,
+  ...miscDemoRoutes,
+};
+
+class AnimationSamples extends StatelessWidget {
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      title: 'Animations Samples',
```
RedBrogdon: nit: "Animation"

animations/lib/main.dart
```diff
@@ -0,0 +1,83 @@
...
+        primarySwatch: Colors.deepPurple,
+      ),
+      routes: allRoutes,
+      home: HomePage(),
+    );
+  }
+}
+
+class HomePage extends StatelessWidget {
+
```
RedBrogdon: nit: the previous class had no whitespace in between the class declaration and the start of the build method.

RedBrogdon: Now that I'm looking at this, I can see there's no `@override` annotation on the build method. Did you remove it, or was it not added by the IDE?

animations/lib/main.dart
```diff
@@ -0,0 +1,83 @@
...
+  Widget build(BuildContext context) {
+    final headerStyle =
+    Theme.of(context).textTheme.title;
```
RedBrogdon: 52/53 should be a single line.

animations/lib/src/misc/expand_card.dart
```diff
@@ -0,0 +1,80 @@
...
+            height: size,
+            child: AnimatedCrossFade(
+              duration: duration,
+              crossFadeState: selected
+                  ? CrossFadeState.showSecond
+                  : CrossFadeState.showFirst,
+              // Use Positioned.fill() to pass the constraints to its children.
+              // This allows the Images to use BoxFit.cover to cover the correct
+              // size
+              layoutBuilder: (Widget topChild, Key topChildKey,
```
RedBrogdon: nit: don't normally put types on lambda parameters

RedBrogdon: Side note: I'm going to remember this `Positioned.fill` trick you've learned. 😄

animations/lib/main.dart
johnpryan: I don't usually use it, since build() and setState() are stable APIs.

> Use the @override annotation judiciously and only for methods where the superclass is not under the programmer's control, the superclass is in a different library or package, and it is not considered stable

I don't see any direction given in the [style guide](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo) - is this documented somewhere?

johnpryan: Added for consistency with other samples

johnpryan:
Travis is now set up - this is ready for another look

johnpryan:
@RedBrogdon PTAL

PR #120, opened 2019-07-24
add animations basics 1-6

johnpryan:


animations/lib/src/basics/04_tweens.dart
```diff
@@ -0,0 +1,65 @@
...
+
+  void initState() {
+    super.initState();
+
+    controller = AnimationController(vsync: this, duration: _duration)
+      ..addListener(() {
+        // Marks the widget tree as dirty
+        setState(() {});
+      });
+    animation = Tween(begin: 0.0, end: accountBalance).animate(controller);
```
RedBrogdon: This is a great demo from a simplicity standpoint, especially since you're writing out the value of the tween rather than just using it as a property.

I'm wondering, though, if it's missing an opportunity to show people one of Tween's most important capabilities -- mapping the controller's double to other types. Is there a way to add a ColorTween or some other interesting change of types and show that as well (two Tweens, in other words)?

animations/lib/src/basics/05_custom_tween.dart
```diff
@@ -0,0 +1,100 @@
+import 'package:flutter/material.dart';
+
+class TypewriterTween extends Tween<String> {
```
RedBrogdon: Nice idea for a custom tween.

animations/lib/src/basics/05_custom_tween.dart
```diff
@@ -0,0 +1,100 @@
...
+  TypewriterTween({String begin = '', String end})
+      : super(begin: begin, end: end);
+
+  String lerp(double t) {
+    var endLength = end.length;
```
RedBrogdon: Is this local necessary?

animations/lib/src/basics/06_animated_builder.dart
```diff
@@ -0,0 +1,57 @@
...
+
+class AnimatedBuilderDemo extends StatefulWidget {
+  static const String routeName = '/basics/animated_builder';
+
+  _AnimatedBuilderDemoState createState() => _AnimatedBuilderDemoState();
+}
+
+class _AnimatedBuilderDemoState extends State<AnimatedBuilderDemo>
+    with SingleTickerProviderStateMixin {
+  static const Color beginColor = Colors.deepPurple;
```
RedBrogdon: You could just have a ColorTween constant, rather than the individual colors. Might be a good thing to show.

RedBrogdon: Actually, no, you can't. There's no const constructor for ColorTween. 🤔

animations/lib/src/basics/01_animated_container.dart
```diff
@@ -0,0 +1,79 @@
...
+double randomMargin() {
+  return Random().nextDouble() * 64;
+}
+
+Color randomColor() {
+  return Color(0xFFFFFFFF & Random().nextInt(0xFFFFFFFF));
+}
+
+class AnimatedContainerDemo extends StatefulWidget {
+  static String routeName = '/basics/01_animated_container';
```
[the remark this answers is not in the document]
johnpryan: There's a logical order of concepts I'd like to stick to, yes. There might be better way to do it, though.

animations/lib/src/basics/02_page_route_builder.dart
```diff
@@ -0,0 +1,46 @@
...
+
+class PageRouteBuilderDemo extends StatelessWidget {
+  static const String routeName = '/basics/page_route_builder';
+
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(),
+      body: Center(
+        child: MaterialButton(
+          child: Text('Go!', style: TextStyle(color: Colors.white)),
```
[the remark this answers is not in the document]
johnpryan: I'll just switch to a `RaisedButton`

animations/lib/src/basics/04_tweens.dart
johnpryan: Let's do that in a separate sample

animations/lib/src/basics/06_animated_builder.dart
johnpryan: I'll add this to the list of questions for the team :)

johnpryan:
@RedBrogdon PTAL

animations/lib/src/basics/01_animated_container.dart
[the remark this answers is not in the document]
RedBrogdon: Your call. It's just the filenames that are weirding me out. Representing them as an ordered list in the app's navigational elements or something like that makes total sense.

animations/lib/src/basics/02_page_route_builder.dart
```diff
@@ -0,0 +1,46 @@
+import 'package:flutter/material.dart';
```
RedBrogdon: Missing its copyright header.

animations/lib/src/basics/04_tweens.dart
RedBrogdon: 👍

PR #122, opened 2019-07-26
Added new test scripts

RedBrogdon:
Scripts for the new tests repo:

https://github.com/flutter/tests.

These scripts will be used to test DevRel's samples with new versions of the Flutter SDK in order to detect regressions in backward compatibility.

### DIFF
--- a/customer_testing.bat
+++ b/customer_testing.bat
@@ -1,0 +1,43 @@
+cd "animations"
+flutter test || exit 1
+cd ..
+
+cd "flutter_maps_firestore"
+flutter test || exit 1
+cd ..
+
+cd "isolate_example"
+flutter test || exit 1
+cd ..
+
+cd "jsonexample"
+flutter test || exit 1
+cd ..
+
+cd "place_tracker"
+flutter test || exit 1
+cd ..
+
+cd "platform_design"
+flutter test || exit 1
+cd ..
+
+cd "platform_view_swift"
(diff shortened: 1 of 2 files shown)


customer_testing.bat
```diff
@@ -0,0 +1,43 @@
...
+cd "provider_shopper"
+flutter test || exit 1
+cd ..
+
+cd "shrine"
+flutter test || exit 1
+cd ..
+
+cd "veggieseasons"
+flutter test || exit 1
```
Hixie: does this work? (as in, it causes the file to fail if there's anything error?)

RedBrogdon: I don't have convenient access to a non-personal Windows machine, so I plan to test this when I leave work tonight. This is what I came up with after some online research.

Hixie: another option (easier in bash than batch probably) is to just iterate across all the directories indiscriminately and run flutter run in each one.

johnpryan: just a fly-by comment, but we might want to consider consolidating these two scripts into a Dart script

RedBrogdon: @Hixie That's how our normal travis script works. With these, I wasn't sure how to reliably iterate over a list of directories in a batch file, so I separated them out. The bash script was done the same way just so they'd be essentially the same script.

@johnpryan Certainly possible. I just did them as dead simple as I could for now.

PR #123, opened 2019-07-30
Add scroll physics demo

johnpryan:
This adds the sample produced for the [Animate a widget using a physics simulation](https://flutter.dev/docs/cookbook/animation/physics-simulation) cookbook article.

### DIFF
--- a/animations/lib/src/misc/physics_card_drag.dart
+++ b/animations/lib/src/misc/physics_card_drag.dart
@@ -35,7 +35,11 @@ class _DraggableCardState extends State<DraggableCard>
     with SingleTickerProviderStateMixin {
   AnimationController _controller;
 
-  /// The position the user has dragged the card to.
+  /// The alignment of the card as it is dragged or being animated.
+  ///
+  /// While the card is being dragged, this value is set to the values computed
+  /// in the GestureDetector onPanUpdate callback. If the animation is running,
+  /// this value is set to the value of the [_animation].
   Alignment _dragAlignment = Alignment.center;
 
   Animation<Alignment> _animation;
--- a/animations/lib/src/misc/physics_card_drag.dart
+++ b/animations/lib/src/misc/physics_card_drag.dart
@@ -40,19 +40,14 @@ class _DraggableCardState extends State<DraggableCard>
 
   Animation<Alignment> _animation;
 
-  /// Update the animation so that it runs from the dragged point back to the
-  /// center.
-  void _updateAnimation() {
+  /// Calculates and runs a [SpringSimulation]
+  void _runAnimation(Offset pixelsPerSecond, Size size) {
     _animation = _controller.drive(
       AlignmentTween(
         begin: _dragAlignment,
         end: Alignment.center,
       ),
     );
-  }
-
-  /// Calculates and runs a [SpringSimulation]
-  void _runAnimation(Offset pixelsPerSecond, Size size) {
     // Calculate the velocity relative to the unit interval, [0,1],
     // used by the animation controller.
     final unitsPerSecondX = pixelsPerSecond.dx / size.width;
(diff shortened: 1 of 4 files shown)


johnpryan:
I think the Travis failure is coming from the fonts-droid package. Fix is here: https://github.com/flutter/samples/pull/124

animations/lib/src/misc/physics_card_drag.dart
```diff
@@ -0,0 +1,113 @@
...
+      onPanUpdate: (details) {
+        setState(() {
+          _dragAlignment += Alignment(
+            details.delta.dx / (size.width / 2),
+            details.delta.dy / (size.height / 2),
+          );
+        });
+      },
+      onPanEnd: (details) {
+        _updateAnimation();
```
RedBrogdon: I had a good bit of difficulty at first parsing out what's happening here. That's more to do with the complexity of the effect you're trying to achieve than anything to do with your code, but anything that could simplify the code for readers is probably a good idea.

For example, could `_updateAnimation` be changed to accept a velocity and do the work of creating a SpringDescription/Simulation and updating the controller? In `initState` it could be called with a zero velocity, and in `onPanEnd` with a real one.

That causes the code to do a little unnecessary work during the first call in `initState`, but it also moves all the work of setting up the animation and controller to a single, easy-to-read method.

johnpryan: For now, I moved the logic from `onDragEnd` into a new method, `runAnimation`. I don't see how to combine `runAnimation` and `updateAnimation`.

johnpryan:
@RedBrogdon PTAL

animations/lib/src/misc/physics_card_drag.dart
```diff
@@ -0,0 +1,113 @@
...
+      ),
+    );
+    // Calculate the velocity relative to the unit interval, [0,1],
+    // used by the animation controller.
+    final unitsPerSecondX = pixelsPerSecond.dx / size.width;
+    final unitsPerSecondY = pixelsPerSecond.dy / size.height;
+    final unitsPerSecond = Offset(unitsPerSecondX, unitsPerSecondY);
+    final unitVelocity = unitsPerSecond.distance;
+
+    const spring = SpringDescription(
```
RedBrogdon: nit: move this out of the function and into the class scope or file scope.

animations/lib/src/misc/physics_card_drag.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  @override
+  _DraggableCardState createState() => _DraggableCardState();
+}
+
+class _DraggableCardState extends State<DraggableCard>
+    with SingleTickerProviderStateMixin {
+  AnimationController _controller;
+
+  /// The position the user has dragged the card to.
```
RedBrogdon: nit: This isn't quite true, since it could also be the current position of the card snapping back from a drag. Consider something like:

"The current alignment of the DraggableCard within its parent. While the card is being dragged, this value is set in the GestureDetector's callbacks. When the drag ends, an animation updates this value until it returns to Offset.zero."

johnpryan: That's a good clarification. I changed this to:

```
  /// The alignment of the card as it is dragged or being animated.
  ///
  /// While the card is being dragged, this value is set to the values from
  /// derived in the GestureDetector onPanUpdate callback. If the animation is
  /// running, this value is set to the value of the [_animation].
```

animations/lib/src/misc/physics_card_drag.dart
johnpryan: I think I'm going to leave this here for readability; it might be better to keep the reference in closer proximity to where it's used.

PR #124, opened 2019-07-30
use fonts-noto package

johnpryan:
fonts-droid is deprecated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ addons:
       - ubuntu-toolchain-r-test
     packages:
       - libstdc++6
-      - fonts-droid-fallback
+      - fonts-noto
 git:
   depth: 3
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ addons:
       - ubuntu-toolchain-r-test
     packages:
       - libstdc++6
-      - fonts-droid
+      - fonts-droid-fallback
 git:
   depth: 3
 env:


RedBrogdon:
It looks like `fonts-noto` is the preferred replacement:

https://lists.ubuntu.com/archives/ubuntu-devel/2016-February/039159.html

@gspencergoog, I think the travis config here is based on what Flutter used to use, and I remember you knew something about how it was set up. If so, can you provide some detail on what that font package is meant to do (do the Android emulators need it, for example)? It's possible it appeared here simply because we cloned the old Flutter CI setup, and it's not actually needed.

Either way, I'd like to be certain of whether using the fallback is a good idea or whether we're likely to run into this issue again later when it, too, is deprecated.

johnpryan:
OK, I just based this on the message I was seeing:

```
Package fonts-droid is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
fonts-droid-fallback
```

johnpryan:
FWIW it looks like the travis configuration is based on the configuration described in this article: https://medium.com/flutter/test-flutter-apps-on-travis-3fd5142ecd8c

gspencergoog:
Sorry, I don't know (or at least don't remember) why that particular font is needed, but I too would suspect that it was because the emulators needed it. I ported our CI setup to Cirrus, but I was largely just replicating the Travis setup.

RedBrogdon:
I'm going to go ahead and merge this for now, since `fonts-noto` is maintained and it gets us back to passing tests. When we actually get to the point that we're generating screenshots, we may need to revisit.

PR #125, opened 2019-08-09
Adding readme

RedBrogdon:
Quick poll: is it better to call this folder `experimental` or something like `master_channel`?

### DIFF
--- a/experimental/README.md
+++ b/experimental/README.md
@@ -25,7 +25,7 @@ flutter channel stable
 
 ## Index
 
-There aren't actually and samples in here just yet. Things are coming, though!
+There aren't actually any samples in here just yet. Things are coming, though!
 
 ## Want more info about the repo in general?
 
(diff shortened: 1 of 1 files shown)


johnpryan:
Hmm.. I like experimental better. Master channel sounds like it might mean that it only works with the flutter master channel.

domesticmouse:
Amusingly the correlation with only working on master branch is intentional, per the content of the readme. =)

RedBrogdon:
Right. My intent is that everything in `/experimental` run on the master channel, and everything else run on stable (and get tested against beta).

That's why `master_channel` might be a better name, but `experimental` gets the spirit of the directory across immediately.

domesticmouse:
I'm equally happy with all the names. Actually, I'd prefer `here_be_dragons`, but I'm old school. ;-)

RedBrogdon:
🐲

PR #127, opened 2019-08-12
add TweenSequence sample

johnpryan:
Adds sample to demonstrate 1. using TweenSequence and 2. using a `Tween<Color>` (A type other than `double`)


animations/lib/src/basics/06_tween_sequence.dart
```diff
@@ -0,0 +1,66 @@
...
+        TweenSequenceItem<Color>(
+          tween: ColorTween(begin: beginColor, end: endColor),
+          weight: weight,
+        ),
+      );
+    }
+
+    controller = AnimationController(duration: duration, vsync: this);
+    animation = TweenSequence<Color>(sequenceItems).animate(controller)
+      ..addListener(() => setState(() {}));
```
RedBrogdon: A number of the basic examples currently in the app (custom tween and animation controller, for example) use setState as the mechanism for putting the value of the animation to work, even though they could be using AnimatedBuilder.

Was using AnimatedBuilder as the "default" way to rebuild with an animation considered? Is once of the two approaches noticeably better than the other?

animations/lib/src/basics/06_tween_sequence.dart
```diff
@@ -0,0 +1,66 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(),
+      body: Center(
+        child: MaterialButton(
+          color: animation.value,
+          onPressed: () {
+            controller.forward().then((_) {
```
RedBrogdon: nit: use a param name rather than an underscore.

animations/lib/src/basics/06_tween_sequence.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:flutter/material.dart';
```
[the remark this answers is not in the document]
johnpryan: done. It looks like a few other files were missing it too.

animations/lib/src/basics/06_tween_sequence.dart
```diff
@@ -0,0 +1,66 @@
...
+  _TweenSequenceDemoState createState() => _TweenSequenceDemoState();
+}
+
+class _TweenSequenceDemoState extends State<TweenSequenceDemo>
+    with SingleTickerProviderStateMixin {
+  static const Duration duration = Duration(seconds: 1);
+  AnimationController controller;
+  Animation<Color> animation;
+
+  static final List<Color> colors = [
```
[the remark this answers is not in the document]
johnpryan: good catch, done.

animations/lib/src/basics/06_tween_sequence.dart
johnpryan: My intention was to introduce AnimatedBuilder later, but now that you mention it, it makes sense to introduce it sooner. 

I'll switch up the order to make AnimatedBuilder show up earlier.

animations/lib/src/basics/06_tween_sequence.dart
johnpryan: In this case, the value is `void` since `controller.forward()` returns a `TickerFuture` (which is a `Future<void>`). And since it's unused, I would rather leave it an underscore. It's a convention of the language (https://github.com/dart-lang/site-www/issues/1101) IIRC but I don't see why the analyzer isn't giving a warning that it's unused when it's given a name (https://github.com/dart-lang/sdk/issues/23050)

johnpryan:
PTAL

animations/lib/src/basics/06_tween_sequence.dart
RedBrogdon: SGTM.

FWIW, though, don't ever be afraid to disagree with me on questions like this. Whenever you see me write out a multi-line question in a thinking-out-loud style like this, I'm genuinely interested in other folks' opinions.

You've already saved me from my own bad ideas several times. :)

animations/lib/src/basics/06_tween_sequence.dart
RedBrogdon: That's definitely a good reason to use the underscore. FWIW, I believe you could refactor the method to use async/await and get rid of it that way:

```dart
onPressed: () async {
  await controller.forward();
  controller.reset();
}
```

johnpryan: I like that idea.  done

PR #136, opened 2019-09-04
Add helpful comments to the animation samples

theacodes:


animations/lib/src/basics/01_animated_container.dart
```diff
@@ -55,7 +60,7 @@ class _AnimatedContainerDemoState extends State<AnimatedContainerDemo> {
                     color: color,
                     borderRadius: BorderRadius.circular(borderRadius),
                   ),
-                  duration: _duration,
+                  duration: Duration(milliseconds: 400),
```
RedBrogdon: Ex post facto, unsolicited drive-by comment: 

While moving the constant from the top makes sense, this can still use the const constructor:

```
duration: const Duration(milliseconds: 400),
```

animations/lib/src/basics/03_animation_controller.dart
```diff
@@ -22,20 +27,31 @@ class _AnimationControllerDemoState extends State<AnimationControllerDemo>
     super.initState();
 
     controller = AnimationController(vsync: this, duration: _duration)
+      // The Widget's build needs to be called every time the animation's
+      // value changes. So add an listener here that will call setState()
```
RedBrogdon: nit: "a listener"

PR #140, opened 2019-09-09
Remove bitcode=NO from add-to-app flows

jmagman:
See https://github.com/flutter/flutter/issues/39359.

### DIFF
--- a/experimental/add_to_app/SimpleIOSExample/SimpleIOSExample.xcodeproj/project.pbxproj
+++ b/experimental/add_to_app/SimpleIOSExample/SimpleIOSExample.xcodeproj/project.pbxproj
@@ -580,7 +580,6 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
-				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = SimpleIOSExample/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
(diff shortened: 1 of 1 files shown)


RedBrogdon:
LGTM. You still need me to TestFlight a new version of the iOS sample app with bitcode enabled, right?

jmagman:
> LGTM. You still need me to TestFlight a new version of the iOS sample app with bitcode enabled, right?

Probably a good thing to test out, but I'm not familiar with how/when sample apps are deployed.  @dnfield do you see a problem with this?

RedBrogdon:
The iOS sample, Veggie Seasons, is the only one that gets published, and it's done on an ad hoc basis -- whenever we make an improvement or there's something new in the SDK to check out, for the most part.

jmagman:
> The iOS sample, Veggie Seasons, is the only one that gets published, and it's done on an ad hoc basis -- whenever we make an improvement or there's something new in the SDK to check out, for the most part.

Ah, I now see https://github.com/flutter/flutter/issues/39575, so it sounds like it's still needed, though this change is only part of my effort to scrub the add-to-app flows.  Looks like there's issues with devicelab not having a new enough version of Xcode, so the other examples haven't been updated. https://github.com/flutter/flutter/pull/39912

jmagman:
> Looks like there's issues with devicelab not having a new enough version of Xcode, so the other examples haven't been updated. [flutter/flutter#39912](https://github.com/flutter/flutter/pull/39912)

This means when bitcode is turned off everywhere, if Travis does any iOS sample builds, it will need to be running the same version of Xcode used to build the engine.

dnfield:
The same or newer :)  But I'm not seeing any jobs to build the iOS binaries on Travis...

PR #146, opened 2019-10-04
Adds data transfer to Add2App samples, using caching on Android

RedBrogdon:
* Updates flutter_module to:
  - Import Provider 
  - Track counter state via a ChangeNotifier model that uses a method channel as its source of truth.
* Updates Android to use a cached engine
* Updates both Android and iOS to:
  - Set up their end of a method channel
  - Track and display counter value between displays of the flutter view.


experimental/add_to_app/android_fullscreen/app/src/main/java/dev/flutter/example/androidfullscreen/MainActivity.kt
```diff
@@ -6,18 +6,71 @@ package dev.flutter.example.androidfullscreen
...
+
+    private var channel : MethodChannel? = null
+    private var count = 0
+    private var counterLabel : TextView? = null
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
 
+        FlutterMain.ensureInitializationComplete(applicationContext, null)
```
RedBrogdon: @matthew-carroll 

This line isn't something we talked about, but I found it necessary to get past this check:

https://github.com/flutter/engine/blob/master/shell/platform/android/flutter_main.cc#L50

Is this the correct approach, or have I hacked my way around a problem the app shouldn't have?

experimental/add_to_app/android_fullscreen/app/src/main/java/dev/flutter/example/androidfullscreen/MainActivity.kt
```diff
@@ -6,18 +6,71 @@ package dev.flutter.example.androidfullscreen
...
+        // Create and warm up Flutter engine.
+        val flutterEngine = FlutterEngine(this)
+
+        flutterEngine
+            .dartExecutor
+            .executeDartEntrypoint(
+                DartExecutor.DartEntrypoint.createDefault()
+            )
+
+        FlutterEngineCache.getInstance().put(ENGINE_ID, flutterEngine)
```
matthew-carroll: This caching mechanism is probably best demonstrated in an `Application` subclass. Generally, one would like to avoid having `ActivityA` know any meaningful details about `ActivityB`. So i'd be a little worried about people replicating this approach in their own app.

experimental/add_to_app/android_fullscreen/app/src/main/java/dev/flutter/example/androidfullscreen/MainActivity.kt
matthew-carroll: For now, you'll need:

```
FlutterMain.startInitialization(...);
FlutterMain.ensureInitializationComplete(...);
```

I'm working on a PR right now that @xster started, which will move those calls inside of `FlutterEngine` so that app devs no longer have to worry about it.

experimental/add_to_app/ios_fullscreen/IOSFullScreen/ViewController.swift
```diff
@@ -6,10 +6,43 @@ import UIKit
 import Flutter
 
 class ViewController: UIViewController {
```
jmagman: Does `GeneratedPluginRegistrant.register(with: self.flutterEngine);` need to be added somewhere now that there's a provider plugin?

RedBrogdon: Provider is just a package, rather than a plugin with native code of its own, so I don't believe this is necessary (correct me if I'm wrong, though!).

I *do* have a separate flutter module in the repo called `flutter_model_using_plugin` that will be used in a different iOS app, and that one will need to do whatever setup is necessary to use plugins.

experimental/add_to_app/android_fullscreen/app/src/main/java/dev/flutter/example/androidfullscreen/MainActivity.kt
RedBrogdon: SGTM. Thanks.

experimental/add_to_app/android_fullscreen/app/src/main/java/dev/flutter/example/androidfullscreen/MainActivity.kt
RedBrogdon: Got it. I moved the caching and channel setup into the application class, and I'm using a simple observable to propagate data out to my main activity. Let me know what you think.

Another option would be to subclass FlutterActivity, have it manage the channel and return a result, and update MainActivity to use startActivityForResult. I got an hour into implementing it that way, then hit a wall when I didn't know where to cut in and set the result before the FlutterActivity Destroyed itself.

RedBrogdon:
@matthew-carroll PTAL

RedBrogdon:
Updated based on offline discussion with @matthew-carroll. Mgmt of the counter's state is much simplified.

RedBrogdon:
@matthew-carroll, PTAL.

experimental/add_to_app/android_fullscreen/app/src/main/java/dev/flutter/example/androidfullscreen/MyApplication.kt
```diff
@@ -0,0 +1,55 @@
...
+
+class MyApplication : Application() {
+    var count = 0
+
+    private var channel: MethodChannel? = null
+
+    override fun onCreate() {
+        super.onCreate()
+
+        FlutterMain.startInitialization(this)
```
matthew-carroll: FYI, these 2 calls will no longer be required once I merge https://github.com/flutter/engine/pull/12806

RedBrogdon: Thanks for the reminder. Just CCed myself on your PR.

PR #148, opened 2019-10-07
touch ./animations

theacodes:
- Ran `flutter create .` to get new project structure files (including web support)
- Removed empty, unused tests.
- Use named arguments for the `Demo` class in `main.dart` (stylistic choice).


animations/.gitignore
```diff
@@ -69,4 +70,4 @@
 !**/ios/**/default.mode2v3
 !**/ios/**/default.pbxuser
 !**/ios/**/default.perspectivev3
-!/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
+!/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
```
RedBrogdon: nit: newline

animations/lib/main.dart
```diff
@@ -25,35 +25,77 @@ class Demo {
   final String route;
   final WidgetBuilder builder;
 
-  const Demo(this.name, this.route, this.builder);
+  const Demo({this.name, this.route, this.builder});
```
RedBrogdon: Nice. Much more readable this way.

animations/test/empty_test.dart
```diff
@@ -1,9 +0,0 @@
-// Copyright 2019 The Flutter team. All rights reserved.
```
RedBrogdon: Without at least one test file in the /test directory, the CI routine will error out.

This would be a great opportunity to add a real widget test, though. 😄

PR #149, opened 2019-10-07
platform_design maintenance

johnpryan:


platform_design/lib/utils.dart
```diff
@@ -26,26 +22,9 @@ const _myListOfRandomColors = [
...
-
-    if (filterWord(prefix) || filterWord(suffix)) continue;
-
-    final wordPair = WordPair(prefix, suffix);
-    yield wordPair;
-  }
-}
+// Avoid customizing the word generator, which can be slow.
+// https://github.com/filiph/english_words/issues/9
+final wordPairIterator = generateWordPairs();
```
xster: I think I did the above to be able to have non-default syllable lengths since the default lengths were too short aesthetically.

xster:
LGTM

platform_design/lib/utils.dart
johnpryan: Ah, I wasn't paying attention to that. My thinking was that it would be simpler to use the default behavior

PR #151, opened 2019-10-07
Maintenance updates for platform_view_swift

RedBrogdon:
This app was so old that I recreated the project structure from scratch, then moved code back into the individual files.


platform_view_swift/ios/Runner/AppDelegate.swift
```diff
@@ -8,36 +8,36 @@ import Flutter
...
-      }
-    });
-
-    return super.application(application, didFinishLaunchingWithOptions: launchOptions)
-  }
-
-  func didUpdateCounter(counter: Int) {
-    flutterResult?(counter)
-  }
+    var flutterResult: FlutterResult?
```
domesticmouse: I’m curious as to the source of the formatting

platform_view_swift/README.md
```diff
@@ -1,10 +1,33 @@
...
+If you have a general question about Flutter, the best places to go are:
+
+* [The FlutterDev Google Group](https://groups.google.com/forum/#!forum/flutter-dev)
+* [The Flutter Gitter channel](https://gitter.im/flutter/flutter)
+* [StackOverflow](https://stackoverflow.com/questions/tagged/flutter)
+
+If you run into an issue with the sample itself, please file an issue
+in the [main Flutter repo](https://github.com/flutter/flutter/issues).
 
-pushd ios/ ; pod install ; popd
```
domesticmouse: Is this cocoa pods install no longer required?

platform_view_swift/ios/Runner/AppDelegate.swift
RedBrogdon: This is a community contribution from way back. I think the dev in question used soft tabs with stops at 2.

platform_view_swift/README.md
RedBrogdon: No. There was a previous dependency on the Material components for iOS, which I pulled for simplicity's sake.

PR #153, opened 2019-10-15
Update add-to-app modules to latest embedding

RedBrogdon:


RedBrogdon:
CC: @blasten. 

I actually recreated both projects from scratch to make sure, but it looks like updating `.metadata` is all I need to do to get the new embedding code into `.android`.

experimental/add_to_app/flutter_module/.metadata
```diff
@@ -4,7 +4,7 @@
 # This file should be version controlled and should not be manually edited.
 
 version:
-  revision: 2d2a1ffec95cc70a3218872a2cd3f8de4933c42f
-  channel: stable
+  revision: 532a8fed41a4f6595965f02f3edf9666ba5ebf44
+  channel: master
```
matthew-carroll: Curious, will we change this to stable whenever the embedding goes stable? Or do we expect to keep this as master into the future?

RedBrogdon: Change to stable once that's possible. The `experimental` directory in the repo is (in my mind) meant for samples that are currently building against master but will eventually "graduate" once the APIs they use are released to the stable channel.

PR #154, opened 2019-10-23
Adds android_using_prebuilt_module to add-to-app samples

RedBrogdon:
New project imports flutter_module as a prebuilt module rather than building it directly.

There are some README updates required for this and the other projects. Those will come in a separate PR.


matthew-carroll:
Generally LGTM, but I'll defer to @blasten because I think the tooling aspects of this are what's important.

PR #157, opened 2019-10-29
Some kotlin cosmetic tweaks

xster:
Added some syntactic sugar for kotlin like lateinits and kotlin extension view references etc. I was just testing one thing in this project combo so I did some driveby but we should probably do this to all the kotlin code.

### DIFF
--- a/experimental/add_to_app/android_using_plugin/app/src/main/java/dev/flutter/example/androidusingplugin/MainActivity.kt
+++ b/experimental/add_to_app/android_using_plugin/app/src/main/java/dev/flutter/example/androidusingplugin/MainActivity.kt
@@ -5,23 +5,16 @@
 package dev.flutter.example.androidusingplugin
 
 import android.os.Bundle
-import android.widget.Button
-import android.widget.TextView
 import androidx.appcompat.app.AppCompatActivity
 import io.flutter.embedding.android.FlutterActivity
+import kotlinx.android.synthetic.main.activity_main.*
 
 class MainActivity : AppCompatActivity() {
-    private var counterLabel: TextView? = null
-
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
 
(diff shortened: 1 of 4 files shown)


experimental/add_to_app/android_using_plugin/app/src/main/java/dev/flutter/example/androidusingplugin/MainActivity.kt
```diff
@@ -5,23 +5,16 @@
 package dev.flutter.example.androidusingplugin
 
 import android.os.Bundle
-import android.widget.Button
-import android.widget.TextView
 import androidx.appcompat.app.AppCompatActivity
 import io.flutter.embedding.android.FlutterActivity
+import kotlinx.android.synthetic.main.activity_main.*
```
RedBrogdon: I actually avoided using the extensions on purpose, since I've read they're no longer the recommended practice. 

https://old.reddit.com/r/androiddev/comments/ala9p2/why_kotlinx_synthetic_is_no_longer_a_recommended/efdvpkg/
https://android-review.googlesource.com/c/platform/frameworks/support/+/882241

experimental/add_to_app/android_using_plugin/app/src/main/java/dev/flutter/example/androidusingplugin/MyApplication.kt
```diff
@@ -16,7 +16,7 @@ const val ENGINE_ID = "1"
 class MyApplication : Application() {
     var count = 0
 
-    private var channel: MethodChannel? = null
+    private lateinit var channel: MethodChannel
```
RedBrogdon: Should have thought to do this myself. Much better this way.

experimental/add_to_app/android_using_plugin/app/src/main/java/dev/flutter/example/androidusingplugin/MainActivity.kt
xster: Ah ok, good to know. Thanks for the info.

PR #161, opened 2019-11-15
Update flutter web samples

johnpryan:
This updates to SDK version (2a576b774c) and re-runs peanut


web/gallery/README.md
```diff
@@ -1,28 +1,32 @@
...
-San Francisco (which is not available on the web) by copy-pasting the `fonts`
-section from Google Sans:
-- Use "GoogleSans" for ".SF Pro Text".
-- Use "GoogleSansDisplay" for ".SF Pro Display"
-- Use "GoogleSans" for ".SF UI Text"
-- Use "GoogleSansDisplay" for ".SF UI Display".
-
-[Flutter gallery]: https://github.com/flutter/flutter/tree/master/examples/flutter_gallery
-[hosted url]: https://flutter.github.io/samples/gallery/#/
+A demo app for Flutter's material design and cupertino widgets, as
```
RedBrogdon: I think this came up last time you did an import like this, but it would be good to trim anything out of this README that doesn't make sense for our repo (the bit about launcher icons, for example).

johnpryan: or just take out the readme 🤷‍♂️

PR #162, opened 2019-11-18
Add CocoaPods installation to the README

Vanethos:
In `experimental/add_to_app` it references that CocoaPods needs to be updated, however there is no mention on this README of how one should proceed to do so.

As such, instructions have been added.

If needed, the following can also be added:

"To verify that you have the latest version, you can run the command:

```bash
pod --version
```

And check with the [latest releases](https://github.com/CocoaPods/CocoaPods/releases) of CocoaPods."

### DIFF
--- a/experimental/README.md
+++ b/experimental/README.md
@@ -16,13 +16,6 @@ flutter channel master
 flutter upgrade
 ```
 
-You will also need to update CocoaPods to the latest version in order to build
-for iOS. To do so, run the following command on a MacOS machine:
-
-```bash
-sudo gem install cocoapods
-```
-
 When you're done, use this command to return to the safety of the stable
 channel:
 
--- a/experimental/README.md
+++ b/experimental/README.md
@@ -16,6 +16,13 @@ flutter channel master
 flutter upgrade
 ```
 
+You will also need to update CocoaPods to the latest version in order to build
+for iOS. To do so, run the following command on a MacOS machine:
+
+```bash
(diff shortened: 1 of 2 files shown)


experimental/README.md
```diff
@@ -16,6 +16,13 @@ flutter channel master
 flutter upgrade
 ```
 
+You will also need to update CocoaPods to the latest version in order to build
```
RedBrogdon: This is the main README for all of the experimental samples, and ultimately not all of them will actually require Cocoapods. The README in add-to-app mentions Cocoapods and says to check here for details, though, so I can see why you're suggesting this change.

Rather than adding this bit here, though, I'd suggest leaving this README unchanged, and modifying add-to-app's README to correctly cover Cocoapods:

**Change disclaimer at the top to just talk about channel:**

```
***The Add-to-App sample is designed to build with Flutter's `master` channel
See the [README](../README.md) in the `experimental` directory for details.***
```

**Add your suggestion as a new section below "Goals" and above "Important bits"**

````
## Installing Cocoapods

The iOS samples in this repo require the latest version of Cocoapods. To install it,
run the following command on a MacOS machine:

```bash
sudo gem install cocoapods
```

See https://guides.cocoapods.org/using/getting-started.html for more details.
````

Vanethos: Yup, you are right, it makes sense!

With the special caveat that we will need to remember to copy/paste this part in each experimental repo that requires CocoaPods

PR #168, opened 2019-11-21
Maintenance for provider_shopper

RedBrogdon:


provider_shopper/lib/main.dart
```diff
@@ -27,8 +27,10 @@ class MyApp extends StatelessWidget {
         // of ChangeNotifierProvider. Moreover, CartModel depends
         // on CatalogModel, so a ProxyProvider is needed.
         ChangeNotifierProxyProvider<CatalogModel, CartModel>(
-            builder: (context, catalog, previousCart) =>
-                CartModel(catalog, previousCart)),
+          builder: (context, catalog, previousCart) =>
+              CartModel(catalog, previousCart),
+          initialBuilder: (context) => CartModel.empty(),
```
filiph: I would switch the order here, so it reads better. `initialBuilder` before `builder`.

RedBrogdon: Chronological order for parameters. Good call. 😄

PR #170, opened 2019-12-05
Use video_player_web

cbenhagen:


johnpryan:
Thanks, this looks good to me, I'll need to check the version of the Flutter SDK to use before we can put this up on https://flutter.github.io/samples/

cbenhagen:
I don't think there would be an upgrade of the Flutter SDK needed for the `video_player_web` plugin. But I guess I should add `video_player_platform_interface` to the pubspec manually or by running `flutter update-packages --force-upgrade` right?

johnpryan:
I don't think `video_player_platform_interface` needs to be explicitly added to the pubspec. There's just a question around which version of the Flutter SDK will work with this package.

cbenhagen:
About any recent Flutter SDK version should work with that package. It's not using anything new as far as I know. Maybe @hterkelsen can confirm?

cbenhagen:
@johnpryan has this been published to https://flutter.github.io/samples/# already?

johnpryan:
Yeah, although I wasn't sure if this was meant to fix the GIFs or the videos. The videos don't seem to be working.

cbenhagen:
It was supposed to fix the videos but I should have tested this locally. There is code in `video_demo.dart` which will not work on the web. Is it ok to fix it in this repo or do I need to fix this in https://github.com/flutter/flutter/tree/master/examples/flutter_gallery?

johnpryan:
I would recommend fixing it in the flutter repo, this version of the gallery is pulled from there.

hterkelsen:
If this is due to videos loaded from assets, I filed an issue to support that here: https://github.com/flutter/flutter/issues/46621

I will send a PR out shortly.

cbenhagen:
Thanks Harry! This is one of the issues. Others include missing connectivity plugin and usage of `dart:io` to check if it is running in an iOS simulator. I will search for a solution and open a PR in `flutter/flutter`.

ditman:
I have opened a PR in flutter/flutter that should fix video playback on web. It's linked to @cbenhagen 's ticket above:

* https://github.com/flutter/flutter/pull/46719

PR #171, opened 2019-12-07
Adds new "prebuilt framework" add-to-app sample for iOS

RedBrogdon:
Also bumps versions and enables bitcode for the three add-to-app iOS samples.


experimental/add_to_app/ios_fullscreen/IOSFullScreen.xcodeproj/project.pbxproj
```diff
@@ -494,7 +494,7 @@
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				COPY_PHASE_STRIP = NO;
 				DEBUG_INFORMATION_FORMAT = dwarf;
-				ENABLE_BITCODE = NO;
+				ENABLE_BITCODE = YES;
```
jmagman: YES is the default, so you can delete the `ENABLE_BITCODE` lines instead of setting to YES.

experimental/add_to_app/ios_using_prebuilt_module/IOSUsingPrebuiltModuleUITests/IOSUsingPrebuiltModuleUITests.swift
```diff
@@ -0,0 +1,30 @@
+// Copyright 2019 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import XCTest
```
jmagman: Should we delete the test target since it's just the boilerplate?  The test target is missing the framework search paths it would need to test Flutter bits.

experimental/add_to_app/ios_fullscreen/IOSFullScreen.xcodeproj/project.pbxproj
RedBrogdon: Done.

experimental/add_to_app/ios_using_prebuilt_module/IOSUsingPrebuiltModuleUITests/IOSUsingPrebuiltModuleUITests.swift
RedBrogdon: I've deleted the tests for now. We don't have a great test strategy for add-to-app samples in general right now, which is something I intend to fix over the holidays.

PR #179, opened 2019-12-12
Fix broken links in gallery sample readme.

ekremkenter:
Update links to adaptive.dart and splash.dart as relative links


RedBrogdon:
Thanks for the fix!

PR #184, opened 2019-12-13
Modify Makefile in flutter/samples, so that we can run the l10n script

perclasson:
- Remove deploy script as it's no longer used.
- Update the root folder in the Makefile, so that we can run the l10n script.
- Ran the formatter over the localization messages.

### DIFF
--- a/gallery/gallery/Makefile
+++ b/gallery/gallery/Makefile
@@ -1,4 +1,4 @@
-ROOT := $(shell git rev-parse --show-toplevel)
+ROOT := $(shell git rev-parse --show-toplevel)/gallery
 FLUTTER := $(shell which flutter)
 FLUTTER_BIN_DIR := $(shell dirname $(FLUTTER))
 FLUTTER_DIR := $(FLUTTER_BIN_DIR:/bin=)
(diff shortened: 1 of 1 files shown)


perclasson:
@shihaohong FYI When we are running the formatter for the localization messages it now sorts it in a different order.

gallery/gallery/Makefile
```diff
@@ -33,12 +33,6 @@ master-branch-check: fetch-master
 		$(error Your master branch is not up to date with origin/master, please pull before deploying)
   endif
 
-
-.PHONY: deploy
-deploy: master-branch-check build-web
```
rami-a: You can also remove `master-branch-check` and  `build-web`

gallery/gallery/lib/l10n/messages_all.dart
```diff
@@ -15,485 +15,485 @@ import 'package:intl/intl.dart';
...
-import 'messages_en_AU.dart' as messages_en_au;
-import 'messages_en_CA.dart' as messages_en_ca;
-import 'messages_en_GB.dart' as messages_en_gb;
-import 'messages_en_IE.dart' as messages_en_ie;
-import 'messages_en_IN.dart' as messages_en_in;
-import 'messages_en_NZ.dart' as messages_en_nz;
-import 'messages_en_SG.dart' as messages_en_sg;
-import 'messages_en_US.dart' as messages_en_us;
-import 'messages_en_ZA.dart' as messages_en_za;
+import 'messages_ko.dart' as messages_ko;
```
rami-a: @shihaohong do you know why the latest version of the l10n script moved all these around?

shihaohong: This was a regression when I introduced tests, and I made a fix immediately after. 

https://github.com/flutter/flutter/pull/46247

Could you review and it approve it if it looks good?

perclasson: Thanks. I'll revert the change here then, and then we rerun this when your change has landed in flutter/flutter.

rami-a:
Oh interesting, I was not aware of this package. But I like the idea of replacing our Makefile with it. I'll create a ticket in our backlog for that.

gallery/gallery/lib/l10n/messages_all.dart
shihaohong: The sorting fix has landed! Not sure if I need to let you all know here, but just in case :)

PR #185, opened 2019-12-13
Fix apk release issue for new Flutter Gallery

JoseAlba:
There was an issue when releasing an APK file onto the app store. The APK file would break on launch. This PR fixed that problem.

### DIFF
--- a/gallery/gallery/android/app/build.gradle
+++ b/gallery/gallery/android/app/build.gradle
@@ -38,7 +38,7 @@ android {
 
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
-        applicationId "com.example.gallery"
+        applicationId "io.flutter.demo.gallery"
         minSdkVersion 16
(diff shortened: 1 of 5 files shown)


gallery/gallery/android/app/build.gradle
```diff
@@ -38,7 +38,7 @@ android {
 
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
```
rami-a: Remove this TODO

RedBrogdon:
Drive-by comment:

If it's still possible to change the package name, we generally use dev.flutter.* these days, since it better matches our domain (flutter.dev).

rami-a:
Since this replaced the old gallery on the play store, I _think_ it needs to use the same package name. @mklim Do you know if this is the case?

mklim:
Yeah, the package name needs to stay the same. https://stackoverflow.com/a/17582393/3968079

PR #188, opened 2019-12-13
Update semantic label of study back button to better explain the action

rami-a:
The study back button takes you back to the gallery and so the semantic label reflects that now.

Closes [#480](https://github.com/material-components/material-components-flutter-gallery/issues/480)


rami-a:
Yeah, generally the way we do this is add in the english strings and then send them off to be translated asynchronously. Every week or so we dump whatever translations we have from the translators into the repo.

PR #191, opened 2019-12-17
[Gallery] Convert the Makefile to instead use the Dart workflow package Grinder.  

perclasson:
Convert the Makefile to instead use the Dart workflow package Grinder.  

Closes https://github.com/material-components/material-components-flutter-gallery/issues/506


gallery/gallery/pubspec.yaml
```diff
@@ -36,6 +36,7 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
+  grinder: ^0.8.0
```
rami-a: Does the grinder package have to be included in the gallery's pubspec.yaml? Since the gallery doesnt actually depend on it

gallery/gallery/tool/grind.dart
```diff
@@ -0,0 +1,81 @@
+import 'dart:io';
```
rami-a: I would suggest moving this file up outside the gallery directory

gallery/gallery/tool/grind.dart
```diff
@@ -0,0 +1,81 @@
...
+
+import 'package:grinder/grinder.dart';
+import 'package:path/path.dart' as p;
```
rami-a: s/p/path

gallery/gallery/tool/grind.dart
```diff
@@ -0,0 +1,81 @@
...
+@Task('Get packages')
+Future<void> pubGet({String directory}) async {
+  await _runProcess(
+    'flutter',
+    ['pub', 'get', if (directory != null) directory],
+  );
+}
+
+@Task('Run Flutter unit tests')
+Future<void> test() async => await _runProcess('flutter', ['test']);
```
rami-a: I think this isn't used, remove

gallery/gallery/pubspec.yaml
perclasson: Yes, we need to be able to download the grinder tool.

It's just put in the `dev_dependencies` here so it won't be used when building the app.

If you think we should put it down one level (as said in the other comment) then I would need to create another pubspec.yaml one level below, which would solve your concern anyways

gallery/codeviewer_cli/README.md
```diff
@@ -5,10 +5,14 @@ A command-line application to highlight dart source code.
 Code segments are highlighted before the app is compiled.
 This is done because the highlighting process can take 300ms to finish, creating a noticeable delay when the demo switches to code page.
 
-The highlighter takes all files in the `gallery/lib/demos/` folder and scans each.
-Highlighted code widgets are stored in the `gallery/lib/codeviewer/code_segments.dart` file.
-Under the root directory, run `make update-code-segments` to run the highlighter.
+The highlighter takes all files in the `gallery/lib/demos/` folder and scans each. Highlighted code widgets are stored in the `gallery/lib/codeviewer/code_segments.dart` file.
```
RedBrogdon: nit: While there are exceptions, markdown in the repo normally uses an ~80 character line break.

gallery/gallery/lib/l10n/README.md
```diff
@@ -22,18 +22,17 @@ This allows use of the English message through your localizations delegate in
...
+## How to Generate GalleryLocalizations with Grinder
 
-For more details on what `make l10n` runs, you can read below under Generate GalleryLocalizations.
+From the `samples/gallery/gallery/` directory:
+1. Make sure you have [grinder](https://pub.dev/packages/grinder) installed by running `flutter pub get`.
+2. Then run `flutter pub run grinder l10n` to generate `GalleryLocalizations`.
 
-The current supported locales list is sorted alphabetically. So after running the script, update
-`gallery_localizations.dart` to move the `en_US` locale to the top of the list.
+For more details on what `flutter pub run grinder l10n` runs, you can read below under *How to Generate GalleryLocalizations with l10n scripts*. The current supported locales list is sorted alphabetically. This means that after running the script, you have to update
```
RedBrogdon: nit: line length.

gallery/gallery/pubspec.yaml
RedBrogdon: +1. `dev_dependencies` is the recommended approach:

https://pub.dev/packages/grinder

You can also activate grinder globally and use `grind` from the command line, though that's optional.

gallery/gallery/tool/grind.dart
RedBrogdon: nit: This file could use a copyright header.

gallery/codeviewer_cli/README.md
perclasson: TIL, makes sense

Makes me think we should have something like https://github.com/markdownlint/markdownlint done in Dart, but for now I'll update it here manually.

gallery/gallery/tool/grind.dart
perclasson: Done.

perclasson: Actually after doing this I realized that this will make Android Studio confused when you open the gallery directory. 

It will then not recognize that the sub directories also have pubspec.yaml files and you have to open each directory separately. Not sure if it's worth moving this down one directory as its probably adds more confusion than help.

perclasson:
I had to update the l10n tool now and the code segments tool to take in the file paths as argument, in order for us to run the scripts in the root folder... Not sure if this is better. What do you think @rami-a?

rami-a:
oh, I think I'd rather there not be arguments to those commands, so probably move it back. Sorry to keep changing my mind

gallery/README.md
```diff
@@ -2,29 +2,65 @@
...
+- Android
+- iOS
+- web
+- macOS
+- Linux
+- Windows
 
-That being said, extra steps must be taken to [enable Desktop support](https://github.com/flutter/flutter/wiki/Desktop-shells#tooling).
+That being said, extra steps must be taken to [enable Desktop support](
+https://github.com/flutter/flutter/wiki/Desktop-shells#tooling). To for example
```
rami-a: `For example, to run the macOS app:`

PR #193, opened 2019-12-18
Shrine: Remove Large Gaps at Bottom of Screen

pennzht:
## Description
In the Shrine app, there used to be large gaps at the bottom of the screen.
This PR implements an algorithm that rearranges the order of cards so that the heights of the columns are closer, reducing the size of gaps.

## Related Issues
Closes https://github.com/material-components/material-components-flutter-gallery/issues/178

## Detailed Description
In the desktop layout of Shrine, product cards are arranged in columns. Product cards can have different heights.

A naive method to arrange product cards is to go from start to end (for example, left to right), giving one card to each column, and looping back to start.
This method creates gaps at the bottom because the cards have different heights, so although two columns have the same number of cards, their height can still be different by a large margin.

Here, the function `balancedLayout` creates a balanced layout using this process:
* Start with an arbitrary layout using the above naive method. This way the number of cards in each column is roughly the same.
* Loops through all pairs of columns.
    - For each pair of columns, `source` and `target`, we may choose a card `A` from column `source` and a card `B` from column `target` and swap them, so that the two columns' height is as close as possible.
    - `A` or `B` could be empty — In this case, we are moving a card from a column to another column.
    - In case no swap improves the situation, we don't swap at all.
* This swapping process is stopped when we cannot make any improvements (or when any improvement is small enough to be negligible). That is, when all columns' heights are close enough that we cannot make them much closer by swapping cards.

This function **does not always** generate the optimal solution. The solution is close to a **local extremum**, which gives very small gaps in most cases.

## Performance
Balanced layouts are **cached** once computed.
Using a rough testing in Chrome,
* computing a balanced layout takes ≤ 9ms;
* reading a cached layout takes ≤ 2ms.

## Screenshots
* Category **All**, before
shrine-all-before
* Category **All**, after
shrine-all-after
* Category **Clothing**, before
shrine-clothing-before
* Category **Clothing**, after
shrine-clothing-after
* Category **Accessories**, before
shrine-accessories-before
* Category **Accessories**, after
shrine-accessories-after


gallery/gallery/lib/studies/shrine/supplemental/balanced_layout.dart
```diff
@@ -0,0 +1,305 @@
...
+/// Used for iteration.
+List<_TaggedHeightData> toListAndAddEmpty(Set<_TaggedHeightData> set) {
+  List<_TaggedHeightData> result = List<_TaggedHeightData>.from(set);
+  result.add(_TaggedHeightData(index: _emptyElement, height: 0));
+  return result;
+}
+
+/// Sort a list of integers, then return the sorted list.
+/// Used in list comprehensions using "for".
+List<int> _sorted(List<int> list) {
```
rami-a: Now that extension methods are part of flutter stable, this could probably be done as an extension on List instead

pennzht: @rami-a Thank you! `flutter analyze` requires me to update SDK to 2.6.0 or later. Shall I?

```
   info • Extension methods weren't supported until version 2.6.0, but this code is required to be able to run on earlier versions •
          lib/studies/shrine/supplemental/balanced_layout.dart:45:1 • sdk_version_extension_methods
```

This requires the following version in `pubspec.yaml`:

```
environment:
  sdk: ">=2.6.0 <3.0.0"
```

gallery/gallery/lib/studies/shrine/supplemental/balanced_layout.dart
```diff
@@ -0,0 +1,305 @@
...
+      [for (final product in products) product.id.toString()].join(',');
+  return '$columnCount;$productString,$largeImageWidth,$smallImageWidth';
+}
+
+/// Given a layout, replace integers by their corresponding products.
+List<List<Product>> _generateLayout({
+  @required List<Product> products,
+  @required List<List<int>> layout,
+}) {
+  return [
```
RedBrogdon: nit: Are you by any chance a skilled python programmer? :)

This is more of a style FYI than anything else, but typically these types of list manipulations use `map` and Dart's other iterator-based methods. No need to rewrite anything here, I just wanted to mention it.

gallery/gallery/lib/studies/shrine/supplemental/balanced_layout.dart
RedBrogdon: Is this method necessary? It looks like it's just wrapping `sort`.

rami-a: I think this should be fine to do since flutter stable is now on dart 2.6. As long as the analyzer on this repo is happy then go ahead and do it!

pennzht: @rami-a @RedBrogdon 

I just realized I can use `list..sort()` to sort and return the list. Thank you for the suggestion!

gallery/gallery/lib/studies/shrine/supplemental/asymmetric_view.dart
```diff
@@ -190,68 +188,109 @@ class DesktopAsymmetricView extends StatelessWidget {
...
+
+  final int columnCount;
+  final List<Product> products;
+  final double largeImageWidth;
+  final double smallImageWidth;
+
+  @override
+  Widget build(BuildContext context) {
+    final Widget _gap = Container(width: 24);
+    final Widget _flex = Expanded(flex: 1, child: Container());
```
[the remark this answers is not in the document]
pennzht: Done.
Thank you!

gallery/gallery/lib/studies/shrine/supplemental/balanced_layout.dart
pennzht: Thank you very much for the suggestion!
I'll keep it like this for this PR, but I'll keep this in mind for future PRs.
I do use python a lot, but I'm still learning.      :D

PR #195, opened 2019-12-18
[Gallery] Implement Cupertino tab bar demo

rami-a:
Closes https://github.com/material-components/material-components-flutter-gallery/issues/95

Simulator Screen Shot - iPhone 11 Pro - 2019-12-18 at 11 58 17


gallery/gallery/lib/demos/cupertino/cupertino_tab_bar_demo.dart
```diff
@@ -0,0 +1,111 @@
...
+            for (final tabInfo in _tabInfo)
+              BottomNavigationBarItem(
+                title: Text(tabInfo.title),
+                icon: Icon(tabInfo.icon),
+              ),
+          ],
+        ),
+        tabBuilder: (context, index) {
+          switch (index) {
+            case 0:
```
johnsonmh: Looks like you might be able to simplify this switch statement with:

```dart
tabBuilder: (context, index) {
  return CupertinoTabView(
    builder: (context) => _CupertinoDemoTab(
      title: _tabInfo[index].title,
      icon: _tabInfo[index].icon,
    ),
    defaultTitle: _tabInfo[index].title,
  );
}
```

rami-a: Done

gallery/gallery/lib/demos/cupertino/cupertino_tab_bar_demo.dart
```diff
@@ -0,0 +1,88 @@
...
+  const _TabInfo(this.title, this.icon);
+
+  final String title;
+  final IconData icon;
+}
+
+class CupertinoTabBarDemo extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    final _tabInfo = <_TabInfo>[
```
RedBrogdon: The `<TabInfo>` here can be dropped. Dart should infer that type correctly.

gallery/gallery/lib/demos/cupertino/cupertino_tab_bar_demo.dart
```diff
@@ -0,0 +1,88 @@
...
+    Key key,
+    @required this.title,
+    @required this.icon,
+  }) : super(key: key);
+
+  final String title;
+  final IconData icon;
+
+  @override
+  Widget build(BuildContext context) {
```
RedBrogdon: Entirely optional (and should be done later even if you decide to do it), but consider giving these scaffolds the ability to navigate to a popup. One of the interesting differences between Cupertino's bottom nav and Android's is that each CupertinoTabView has its own navigation stack, so you can bring up a popup, go to another tab, then return to the original and see the popup still there.

rami-a: That's a good idea, I think the demo in the old gallery did this too.

gallery/gallery/lib/demos/cupertino/cupertino_tab_bar_demo.dart
rami-a: Done

PR #197, opened 2019-12-19
[Gallery] Remove MaterialApp for demo screen for simplification

perclasson:
This simplifies the demo screen by removing the MaterialApp. It looks the same and behaves the same as before.

Closes https://github.com/material-components/material-components-flutter-gallery/issues/503

Simulator Screen Shot - iPad Pro (11-inch) - 2019-12-19 at 11 23 14


gallery/gallery/lib/demos/material/list_demo.dart
```diff
@@ -22,6 +22,7 @@ class ListDemo extends StatelessWidget {
   Widget build(BuildContext context) {
```
rami-a: Does the full screen dialog demo still work?

perclasson: No it does not. Good catch. Let me fix this.

perclasson: I wrapped it in a Navigator now to fix it, will mean that Focus won't work on this page. 

See screenshot of working page:
Screenshot 2019-12-19 at 16 57 50

rami-a: That's fine for now, until we can revisit focus. Thanks!

gallery/gallery/lib/demos/material/dialog_demo.dart
```diff
@@ -124,38 +124,48 @@ class DialogDemo extends StatelessWidget {
...
-            }
-          },
-        ),
-      ),
+    return Navigator(
+      // Adding [ValueKey] to make sure that the widget gets rebuilt when
+      // changing type.
+      key: ValueKey(type),
+      onGenerateRoute: (settings) {
+        return MaterialPageRoute<Widget>(
```
RedBrogdon: Is the widget returned by this route used anywhere? I can see some pop statements on 173 and 205, but those look like they're cutting through this navigator and going to the root insatnce.

perclasson: This is to ensure that the fullscreen dialog (`_FullScreenDialogDemo`) does not take up the entire screen and instead just the space within the Navigator that is defined here.

It is true that we have `rootNavigator: true` on those lines, not sure why and it does not seem to work exactly as expected. I'll have to take a second look at this after the holidays I think.

RedBrogdon: The Navigator part I totally understand and support. I was just looking at the type parameter on line 132 and nitpickingly wondering if it should be `MaterialPageRoute<void>`. 😄

perclasson: Thanks for explaining further. That makes sense, changed it to void now because as you say whenever we use this Navigator it is never with any value. 😄 

Also I had to change the MaterialPageRoute to one without any transition animations (unrelated to this issue).

gallery/gallery/lib/demos/material/dialog_demo.dart
```diff
@@ -124,42 +124,72 @@ class DialogDemo extends StatelessWidget {
...
+            ),
+          ),
+        );
+      },
     );
   }
 }
 
+/// A MaterialPageRoute without any transition animations.
+class _NoAnimationMaterialPageRoute<T> extends MaterialPageRoute<T> {
```
rami-a: What was the need to do this @perclasson ?

perclasson: Otherwise on the macOS at least, when I change the demo type it animates in from the left. Looks a bit different compared to the other demos.

PR #200, opened 2019-12-27
[Gallery] Implement material progress indicator and cupertino activity indicator demos

rami-a:
Closes https://github.com/material-components/material-components-flutter-gallery/issues/243
Closes https://github.com/material-components/material-components-flutter-gallery/issues/241

Screen Shot 2019-12-27 at 1 55 54 PM
Screen Shot 2019-12-27 at 1 56 07 PM
Screen Shot 2019-12-27 at 1 56 13 PM


gallery/gallery/lib/demos/material/progress_indicator_demo.dart
```diff
@@ -0,0 +1,137 @@
...
+      });
+  }
+
+  @override
+  void dispose() {
+    _controller.stop();
+    super.dispose();
+  }
+
+  String _title(BuildContext context) {
```
johnsonmh: nit: since this is a `StatefulWidget`, the `context` is available as a member already, so this method doesn't need to take a `context` param, and can be a getter.

```suggestion
  String get _title {
```

gallery/gallery/lib/demos/material/progress_indicator_demo.dart
```diff
@@ -0,0 +1,137 @@
...
+    return '';
+  }
+
+  void _handleTap() {
+    setState(() {
+      // valueAnimation.isAnimating is part of our build state
+      if (_controller.isAnimating) {
+        _controller.stop();
+      } else {
+        switch (_controller.status) {
```
johnsonmh: I feel like this logic is done twice (here and in `initState`), maybe you could make a function like `startOrStopAnimation`?

gallery/gallery/lib/demos/material/progress_indicator_demo.dart
rami-a: done

gallery/gallery/lib/demos/material/progress_indicator_demo.dart
rami-a: This was slightly different in order to handle stop/starting the animation when you tap. But this logic didn't add much value to the demo so I just removed it instead

PR #205, opened 2020-01-03
Maintenance on jsonexample

RedBrogdon:
* Blows away the original project structure and replaces it with stuff created by a brand new `flutter create`
  - This updates the project to use AndroidX.
  - All dependencies have been checked and upgraded. 
  - No Dart code changes are involved.


jsonexample/.gitignore
```diff
@@ -15,57 +15,23 @@
 *.iws
 .idea/
 
-# Visual Studio Code related
-.vscode/
+# The .vscode folder contains launch configuration and tasks you configure in
+# VS Code which you may wish to be included in version control, so this line
+# is commented out by default.
+#.vscode/
```
johnpryan: nit: I think we should consider leaving out IDE directories altogether since this is a personal preference. We've discussed recommending putting them in [.gitignore_global](https://davidwalsh.name/global-gitignore) instead.

RedBrogdon: Whoa, good catch. I didn't realize they'd updated the default gitignore to leave .vscode in the repo.

I'd prefer to leave anything IDE-specific out, so we're definitely in agreement here.

johnpryan: Is this file is generated by `flutter create`?

RedBrogdon: Yep. The interesting part is that subsequent calls to `flutter create` can create new files not covered in the old .gitignore, but the SDK team doesn't want to automatically edit that file, since the developer may already have modified it themselves.

So you get this weird situation where the SDK is creating files that should be kept out of source control, but the gitignore isn't keeping them out. @jmagman has talked about creating specific directories within the project structure for ephemeral files, so we can gitignore the directories rather than specific files.

PR #208, opened 2020-01-07
Bump gallery to v2.1

rami-a:


gallery/gallery/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: gallery
 description: A resource to help developers evaluate and use Flutter.
 
-version: 2.0.0+7
+version: 2.1.0+020100
```
perclasson: What does `020100` stand for here?

rami-a: so I learned that the part after the plus is interpreted as the version code on android and is an incrementing integer.

So to make it easy to update, I aligned it with the user-facing version name "2.1.0" giving 2 digits for each part.

PR #218, opened 2020-01-09
Adding macos files, recreating android and ios to isolate_example.

domesticmouse:


isolate_example/android/app/build.gradle
```diff
@@ -22,23 +22,28 @@ if (flutterVersionName == null) {
...
+    }
+
     lintOptions {
         disable 'InvalidPackage'
     }
 
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
-        applicationId "dev.flutter.isolate_example"
+        applicationId "com.example.isolate_example"
```
RedBrogdon: Is this a necessary change? It's not hugely important, but having all our samples aligned on their domains makes my inner Marie Kondo happy.

RedBrogdon:
This caught me a bit flat-footed. I didn't realize you were planning to add desktop support to the existing apps right now (that doesn't mean it's not a good idea, just that I wasn't expecting it).

* Does this impact the app's ability to build against the stable channel for Android and iOS?
* Do they need to use a different channel for macOS?
  - If so, should the README be updated to mention Desktop support and how to build/run?
* Is there more work to be done on the desktop side for this sample (limiting window size, for example)?

isolate_example/android/app/build.gradle
domesticmouse: It's what the scripts now generate. To change it back requires some fairly tedious work, moving files around and manually editing configuration files, that will discourage keeping the samples up to date. I understand the desire, but realistically we should just go with whatever the `flutter create` script creates.

domesticmouse:
The test infrastructure is all green, so I believe that `android` and `ios` just work for stable. The major change that is introduced by this uplift is that Android now uses `androidx` and the like, which shouldn't affect these samples. For people running this code on stable, `web` and `macos` won't show up as a runnable target in the tooling. 

I'm happy to add some boilerplate to the README linking to the getting started guides for web and macOS. However, attempting to pull in window minimum sizing and the like and keep the code working on stable, which has no concept of macOS yet is going to be messy. Apart from the fact I haven't landed https://github.com/google/flutter-desktop-embedding/pull/615 yet.

isolate_example/android/app/build.gradle
RedBrogdon: Using the `--org` parameter should take any manual work out of it:

`flutter create --org=dev.flutter project_name`

If that doesn't work, though, and you'd be required to manually edit the files, then definitely ignore me here.

RedBrogdon:
SGTM.

The AndroidX change is something I've had in my todo list for all the samples for a while now, so that's great.

Re: editing the README, if the intent is to add optional desktop compilation without making a big deal out of it (read: experts can find it, but we don't want new Flutter devs paying attention to it), then leaving the READMEs alone is the smart move. My main concern was making sure someone new didn't stumble across these, compile them for desktop, and then wonder why the layout (designed in mobile-only world) looks like crap in a 1600x600 window.

isolate_example/android/app/build.gradle
domesticmouse: Ooh nice

PR #219, opened 2020-01-09
Add command to travis CI to analyze if update-code-segments has been run for Flutter Gallery

perclasson:
To make sure that code segments are always up to date, this CI check has been added. If there is any diff it will fail.

I am now sorting the code segments names to make sure that the output is the same on a Travis CI and my local Mac machine. See line 219 in `codeviewer_cli/segment_generator.dart`. This means that `gallery/gallery/lib/codeviewer/code_segments.dart` was changed even though I did not update any demo.

Closes https://github.com/material-components/material-components-flutter-gallery/issues/520


gallery/gallery/tool/grind.dart
```diff
@@ -54,12 +54,37 @@ Future<void> l10n() async {
...
-  Dart.run(path.join(codeviewerPath, 'bin', 'main.dart'));
-  final targetFilePath = path.join('lib', 'codeviewer', 'code_segments.dart');
-  await format(path: targetFilePath);
+  Dart.run(path.join(codeviewerPath, 'main.dart'));
+  final codeSegmentsPath = path.join('lib', 'codeviewer', 'code_segments.dart');
+  await format(path: codeSegmentsPath);
+}
+
+@Task('Verify and analyze dart files')
+Future<void> analyze() async {
```
RedBrogdon: nit: Do you mind calling this something other than analyze? We use that word a lot to refer specifically  to `flutter analyze`. Would something like `verifyCodeSegments` work?

travis_script.sh
```diff
@@ -57,6 +57,21 @@ do
     popd
 done
 
+declare -a ANALYZE_PROJECT_NAMES=(
```
RedBrogdon: Are there likely to be other apps that require this type of check? If not, don't be afraid to just do it directly rather than use a loop.

RedBrogdon:
Pardon my drive-by comments, since I know you're in progress on this. :)

gallery/gallery/tool/grind.dart
perclasson: Makes sense. I'll rename it to verifyCodeSegments!

travis_script.sh
perclasson: True, most likely not. I'll move it out to just do it directly.

gallery/gallery/tool/grind.dart
```diff
@@ -68,6 +99,26 @@ Future<void> _runProcess(String executable, List<String> arguments) async {
...
+// Reference: https://github.com/dart-lang/sdk/issues/31666
+Future<String> _startProcess(String executable,
+    {List<String> arguments = const [], String input}) async {
+  final output = <int>[];
+  final completer = Completer<int>();
+  final process = await Process.start(executable, arguments);
+  process.stdin.writeln(input);
+  process.stdout.listen((event) {
+    output.addAll(event);
+  }, onDone: () async => completer.complete(await process.exitCode));
```
clocksmith: nit add trailing comma and reformat

gallery/gallery/tool/grind.dart
```diff
@@ -68,6 +99,26 @@ Future<void> _runProcess(String executable, List<String> arguments) async {
...
+  await process.stdin.close();
+
+  final exitCode = await completer.future;
+  if (exitCode != 0) {
```
clocksmith: what happens if you were to call exit(0) ?

perclasson: Then we would have exited the program, even though the process was successfully run. I.e. non-intended behavior. I'll add back the line `stderr.write('Running "$executable ${args.join(' ')}" failed with $exitCode.\n');` so that it's more clear for why we have this.

gallery/gallery/tool/grind.dart
perclasson: Done!

gallery/gallery/tool/codeviewer_cli/main.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright 2019 The Flutter team. All rights reserved.
```
RedBrogdon: nit: 2020? Or is this a modified version of the previous `main.dart`?

perclasson: It is a modified version of the previous main.dart!

PR #221, opened 2020-01-09
[Gallery] Implement cupertino pull-to-refresh demo

rami-a:
Closes https://github.com/material-components/material-components-flutter-gallery/issues/240

Simulator Screen Shot - iPhone 11 Pro - 2020-01-09 at 15 01 48


gallery/gallery/lib/demos/cupertino/cupertino_refresh_demo.dart
```diff
@@ -0,0 +1,83 @@
...
+
+  @override
+  void initState() {
+    super.initState();
+    _repopulateList();
+  }
+
+  void _repopulateList() {
+    final random = Random();
+    randomList = List<int>.generate(listCount, (i) => i + 1)..shuffle(random);
```
johnsonmh: nit: the shuffling is the repopulating part, the actual list generation only needs to happen once. 

So this line would change to `randomList.shuffle(random)` and above, you'd initialize it with: 

```dart
final randomList = List<int>.generate(listCount, (i) => i + 1);
```

rami-a: Done

PR #223, opened 2020-01-09
Add web and macos, update android and ios to jsonexample

domesticmouse:


jsonexample/android/app/build.gradle
```diff
@@ -38,7 +38,7 @@ android {
 
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
-        applicationId "dev.flutter.example.jsonexample"
```
RedBrogdon: Looks like my naming conventions weren't as standard as I thought. 😄

domesticmouse: That amused me too =)

PR #224, opened 2020-01-10
[Gallery] Refactor list of demos to use common variables

rami-a:
* Factor `GalleryLocalizations.of(context)` into a variable
* Factor the documentation base url into a variable

### DIFF
--- a/gallery/gallery/lib/data/demos.dart
+++ b/gallery/gallery/lib/data/demos.dart
@@ -38,7 +38,7 @@ import 'package:gallery/demos/reference/typography_demo.dart';
 import 'package:gallery/l10n/gallery_localizations.dart';
 import 'package:gallery/themes/material_demo_theme_data.dart';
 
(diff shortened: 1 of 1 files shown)


gallery/gallery/lib/data/demos.dart
```diff
@@ -38,6 +38,8 @@ import 'package:gallery/demos/reference/typography_demo.dart';
 import 'package:gallery/l10n/gallery_localizations.dart';
 import 'package:gallery/themes/material_demo_theme_data.dart';
 
+const docsBaseUrl = 'https://api.flutter.dev/flutter';
```
johnsonmh: nit: this should probably be private to this class.

```suggestion
const _docsBaseUrl = 'https://api.flutter.dev/flutter';
```

rami-a: oops, fixed

PR #231, opened 2020-01-13
jsonexample: update web dir

domesticmouse:


jsonexample/web/manifest.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "jsonexample",
+    "short_name": "jsonexample",
+    "start_url": ".",
+    "display": "minimal-ui",
+    "background_color": "#0175C2",
+    "theme_color": "#0175C2",
+    "description": "A new Flutter project.",
```
RedBrogdon: 🤔 

The description in the pubspec is "A demonstration of JSON parsing." Was this kicked out by `flutter create`? If so, we should probably file an issue.

domesticmouse: Yup. Manually patched, for now.

PR #233, opened 2020-01-13
platform_design: update android and ios, add web

domesticmouse:


platform_design/web/manifest.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "platform_design",
+    "short_name": "platform_design",
+    "start_url": ".",
+    "display": "minimal-ui",
+    "background_color": "#0175C2",
+    "theme_color": "#0175C2",
+    "description": "A new Flutter project.",
```
RedBrogdon: There it is again. This project also has a real description in the pubspec.

domesticmouse: Who is the PM for Flutter for web these days? This file only just started being generated in the last couple of days.

domesticmouse: https://github.com/flutter/flutter/issues/48742

PR #237, opened 2020-01-13
update android and ios, add macos and web

domesticmouse:


veggieseasons/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -386,9 +386,10 @@
 					"$(inherited)",
 					"$(PROJECT_DIR)/Flutter",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = dev.flutter.VeggieSeasons;
```
RedBrogdon: This is a published app, so we can't change the Bundle ID.

veggieseasons/lib/main.dart
```diff
@@ -11,6 +11,7 @@ import 'package:veggieseasons/screens/home.dart';
 import 'package:veggieseasons/styles.dart';
 
 void main() {
+  WidgetsFlutterBinding.ensureInitialized();
```
RedBrogdon: Remind me -- which platform/call requires this to be up front? Shared prefs plugin?

veggieseasons/macos/Runner/Configs/AppInfo.xcconfig
```diff
@@ -0,0 +1,14 @@
...
+//
+// This may be replaced with something auto-generated from metadata (e.g., pubspec.yaml) in the
+// future. If not, the values below would default to using the project name when this becomes a
+// 'flutter create' template.
+
+// The application's name. By default this is also the title of the Flutter window.
+PRODUCT_NAME = veggieseasons
+
+// The application's bundle identifier
+PRODUCT_BUNDLE_IDENTIFIER = io.flutter.veggieseasons.veggieseasons
```
RedBrogdon: Should this match the Bundle ID used for iOS? If so, it should be set to dev.flutter.veggieseasons as well.

veggieseasons/lib/main.dart
domesticmouse: The following line spits out a warning at run time because it is being run prior to `runApp()`.

veggieseasons/ios/Runner.xcodeproj/project.pbxproj
domesticmouse: Reverted Bundle ID change

veggieseasons/macos/Runner/Configs/AppInfo.xcconfig
domesticmouse: Done

PR #240, opened 2020-01-13
Deleting web/gallery in favour of the newer gallery/gallery

domesticmouse:


RedBrogdon:
LGTM in principle. @johnpryan, what do you think?

PR #241, opened 2020-01-14
flutter pub upgrade

domesticmouse:


RedBrogdon:
A thought occurs. Now that I've broken the seal on having a `/tool` directory, and you're `flutter create`-ing and checking over each app, we could consider adding a new script that would

PR #242, opened 2020-01-14
Scripting flutter pub upgrade

domesticmouse:


tool/flutter_pub_upgrade.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+
+scriptDirectory="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+for file in `find "${scriptDirectory}/.." -name pubspec.yaml`
+do
+    (
+        cd `dirname $file`
+        echo "Updating `pwd`"
+        "${localSdkPath}/bin/flutter" packages get
```
RedBrogdon: Should this be `packages upgrade`?

tool/flutter_pub_upgrade.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+set -e
+
+# Backs up one directory at a time, looking for one called "flutter". Once it
+# finds that directory, an absolute path to it is returned.
+function getFlutterPath() {
```
RedBrogdon: When I was thinking about this script before, I imagined it as something we'd run locally to update all the apps at once, rather than something that Travis would run.

If that's your idea as well, all this localSdkPath code can be taken out, since `flutter` should just be in $PATH.

tool/flutter_pub_upgrade.sh
domesticmouse: Yeah, it really should =)

tool/flutter_pub_upgrade.sh
domesticmouse: Done

RedBrogdon:
CC @johnpryan and @theacodes so you'll know this exists for future use. 😄

PR #243, opened 2020-01-14
index: Remove stale link

ybakos:
Matt Carroll's flutter_ui_challenge stuff doesn't exist any more, so the search linked here ends up being blank.

Remove the reference and link.


RedBrogdon:
Thanks for submitting this. We're working on a replacement for that index file, which makes it easy to forget to maintain it.

PR #251, opened 2020-01-22
[Gallery] Implement cupertino pickers demo

rami-a:
Closes https://github.com/material-components/material-components-flutter-gallery/issues/242

Simulator Screen Shot - iPhone 11 Pro - 2020-01-22 at 17 02 12
Simulator Screen Shot - iPhone 11 Pro - 2020-01-22 at 17 02 14
Simulator Screen Shot - iPhone 11 Pro - 2020-01-22 at 17 02 18


gallery/gallery/lib/l10n/gallery_localizations.dart
```diff
@@ -1866,6 +1866,56 @@ class GalleryLocalizations {
...
+        locale: _localeName,
+        name: 'demoCupertinoPickerTimer',
+        desc: r'Label to open a countdown timer picker.');
+  }
+
+  String get demoCupertinoPickerTitle {
+    return Intl.message('Pickers',
+        locale: _localeName,
+        name: 'demoCupertinoPickerTitle',
+        desc: r'Title for the cupertino pickers component demo.');
```
guidezpl: Would mention time picker here in the description

rami-a:
> LGTM.
> 
> Can you add a screenshot from another locale, say German? Just to see how it affects the date formatting, assuming we care about that in the demos.

@perclasson so looks like the pickers themselves are localized but the text label isn't. I think this is fine for now but I'm asking Hans about it

Simulator Screen Shot - iPhone 11 Pro - 2020-01-23 at 11 00 52

gallery/gallery/lib/l10n/gallery_localizations.dart
rami-a: Done

PR #256, opened 2020-01-24
[Gallery] Use new package flutter_localized_locales

guidezpl:
I've published a new package flutter_localized_locales, that we'll use instead of relying on my fork of flutter_localized_countries.

Closes https://github.com/material-components/material-components-flutter-gallery/issues/526
https://github.com/material-components/material-components-flutter-gallery/issues/527


guidezpl:
I've requested review from the owner for the PR. But I was just thinking it might be better to have it as its own package published separately

PR #260, opened 2020-01-24
[Gallery] Add version number to About dialog

guidezpl:
Closes https://github.com/material-components/material-components-flutter-gallery/issues/529

Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-24 at 18 01 51

### DIFF
--- a/gallery/gallery/lib/pages/about.dart
+++ b/gallery/gallery/lib/pages/about.dart
@@ -54,12 +54,12 @@ class _AboutDialog extends StatelessWidget {
           mainAxisSize: MainAxisSize.min,
           children: [
             FutureBuilder(
-                future: getVersionNumber(),
-                builder: (context, snapshot) => Text(
-                      snapshot.hasData ? '$name ${snapshot.data}' : '$name',
-                      style: textTheme.display1
-                          .apply(color: colorScheme.onPrimary),
-                    )),
+              future: getVersionNumber(),
+              builder: (context, snapshot) => Text(
+                snapshot.hasData ? '$name ${snapshot.data}' : '$name',
+                style: textTheme.display1.apply(color: colorScheme.onPrimary),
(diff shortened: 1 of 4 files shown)


gallery/gallery/lib/pages/about.dart
```diff
@@ -47,12 +53,13 @@ class _AboutDialog extends StatelessWidget {
...
-                style: textTheme.display1.apply(color: colorScheme.onPrimary),
-              ),
-            ),
+            FutureBuilder(
+                future: getVersionNumber(),
+                builder: (context, snapshot) => Text(
+                      snapshot.hasData ? '$name ${snapshot.data}' : '$name',
+                      style: textTheme.display1
+                          .apply(color: colorScheme.onPrimary),
+                    )),
```
rami-a: This is indented somewhat weirdly, I would try adding a trailing comma here. And maybe not using the => syntax.

gallery/gallery/lib/pages/about.dart
```diff
@@ -20,6 +21,11 @@ void showAboutDialog({
   );
 }
 
+Future<String> getVersionNumber() async {
+  PackageInfo packageInfo = await PackageInfo.fromPlatform();
```
rami-a: Can you check if this works on web and macOS? If it doesn't thats ok but maybe add a comment about that

gallery/gallery/lib/pages/about.dart
guidezpl: Fixed with a comma

gallery/gallery/lib/pages/about.dart
guidezpl: The package doesn't support web and macOS yet. I'll see if I can add it for macOS, otherwise will create an issue

PR #287, opened 2020-01-28
[Gallery] Preserve demo state, simplify demo width calculations on desktop

guidezpl:
TLDR: Fixes a bugs on both desktop and mobile which caused demo state to be reset. 
This was caused by rendering different trees; on mobile: based on whether the options/info panels were open; on desktop, based on whether the demo was in fullscreen. Additionally on desktop, it seemed that conditionally rendering the sized box (or container) preceding the demoContent caused the demoContent widget to be fully rebuilt. Instead, I always specify its width.

This PR also removes width calculations, instead specifying a max width for info and options panes. This incidentally causes the demoContent to remain the same width regardless of the chosen section on desktop. The demoContent grows with the screen size, as before. The section on the left grows until the specified max width.

Screen Shot 2020-01-28 at 11 57 23 AM
Screen Shot 2020-01-28 at 11 57 27 AM
Screen Shot 2020-01-28 at 11 57 29 AM
Screen Shot 2020-01-28 at 11 57 39 AM

Closes https://github.com/material-components/material-components-flutter-gallery/issues/312

### DIFF
--- a/gallery/gallery/lib/pages/demo.dart
+++ b/gallery/gallery/lib/pages/demo.dart
@@ -345,7 +345,6 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
       demoContent = Semantics(
         label: MaterialLocalizations.of(context).modalBarrierDismissLabel,
         child: GestureDetector(
-          behavior: HitTestBehavior.opaque,
           onTap: () {
             if (_state != _DemoState.normal) {
               setStateAndUpdate(() {
--- a/gallery/gallery/lib/pages/demo.dart
+++ b/gallery/gallery/lib/pages/demo.dart
@@ -256,12 +256,14 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
         appBar.preferredSize.height;
     final maxSectionHeight = isDesktop ? contentHeight : contentHeight - 64;
     final horizontalPadding = isDesktop ? mediaQuery.size.width * 0.12 : 0.0;
+    final maxSectionWidth = 420.0;
 
     Widget section;
     switch (_state) {
       case _DemoState.options:
         section = _DemoSectionOptions(
           maxHeight: maxSectionHeight,
+          maxWidth: maxSectionWidth,
           configurations: widget.demo.configurations,
           configIndex: _configIndex,
           onConfigChanged: (index) {
@@ -277,6 +279,7 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
       case _DemoState.info:
         section = _DemoSectionInfo(
           maxHeight: maxSectionHeight,
+          maxWidth: maxSectionWidth,
           title: _currentConfig.title,
           description: _currentConfig.description,
         );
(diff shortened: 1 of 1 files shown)


gallery/gallery/lib/pages/demo.dart
```diff
@@ -314,46 +316,15 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
...
-                  child: demoContent,
-                ),
-              ],
-            );
+      final isFullScreen = _state == _DemoState.fullscreen;
+      final Widget sectionAndDemo = Row(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          if (!isFullScreen) Expanded(child: section),
+          SizedBox(width: !isFullScreen ? 48.0 : 0),
```
rami-a: I think this might be easier to read as:

```
if (!isFullScreen) ...[
  Expanded(child: section),
  SizedBox(),
]
Expanded(child: demoContent),
```

gallery/gallery/lib/pages/demo.dart
```diff
@@ -371,20 +342,23 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
...
-          label: MaterialLocalizations.of(context).modalBarrierDismissLabel,
-          child: GestureDetector(
-            behavior: HitTestBehavior.opaque,
-            onTap: () {
+      demoContent = Semantics(
+        label: MaterialLocalizations.of(context).modalBarrierDismissLabel,
+        child: GestureDetector(
+          behavior: HitTestBehavior.opaque,
+          onTap: () {
+            if (_state != _DemoState.normal) {
```
rami-a: Since the gestureDetector is added always now, does this interfere with interacting with the demo?

gallery/gallery/lib/pages/demo.dart
```diff
@@ -506,49 +482,52 @@ class _DemoSectionOptions extends StatelessWidget {
...
-              top: 12,
-              end: 24,
-            ),
-            child: Text(
-              GalleryLocalizations.of(context).demoOptionsTooltip,
-              style: textTheme.display1.apply(
-                color: colorScheme.onSurface,
-                fontSizeDelta:
-                    isDisplayDesktop(context) ? desktopDisplay1FontDelta : 0,
+    return Align(
```
rami-a: What is the `Align` for?

gallery/gallery/lib/pages/demo.dart
guidezpl: This is what I was saying causes the demoContent to be rebuilt from scratch. I could write it as 

```dart
if (!isFullScreen) ...[
  Expanded(child: section),
  SizedBox(),
]
Expanded(child: demoContent),
```
and it works

rami-a: Very strange...ok I think I prefer what you originally had instead of the 2nd sizedbox

gallery/gallery/lib/pages/demo.dart
guidezpl: I've tested on chrome, iOS and macOS, and saw no adverse effects. I'll remove behavior: HitTestBehavior.opaque because it's not doing anything here and might be confusing

gallery/gallery/lib/pages/demo.dart
guidezpl: A Container within an Expanded will fill the space, even if you set width/height constraints, unless you use an Align.

PR #288, opened 2020-01-28
[Gallery] Close Shrine cart on back button press for android

perclasson:
Move up `WillPopScope` above `MaterialApp` to make sure that when the expanding bottom sheet (cart) is open, it will close the sheet instead of going back when pressing the back button on Android.

Closes https://github.com/material-components/material-components-flutter-gallery/issues/519

### DIFF
--- a/gallery/gallery/lib/studies/shrine/app.dart
+++ b/gallery/gallery/lib/studies/shrine/app.dart
@@ -78,7 +78,7 @@ class _ShrineAppState extends State<ShrineApp> with TickerProviderStateMixin {
     );
   }
 
-  // Closes the bottom sheet if it is open (this should only be relevant for Android).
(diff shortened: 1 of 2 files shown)


gallery/gallery/lib/studies/shrine/app.dart
```diff
@@ -78,6 +78,18 @@ class _ShrineAppState extends State<ShrineApp> with TickerProviderStateMixin {
     );
   }
 
+  // Closes the bottom sheet if it is open (this should only be relevant for Android).
```
guidezpl: nit, suggestion: would remove the () about Android

perclasson: Done!

PR #289, opened 2020-01-28
[Gallery] Add padding to demo appBar

guidezpl:
This adds a bit of padding after the actions for the last ripple to be visible

Closes https://github.com/material-components/material-components-flutter-gallery/issues/208

Screen Shot 2020-01-28 at 3 28 12 PM
Screen Shot 2020-01-28 at 3 45 24 PM

### DIFF
--- a/gallery/gallery/lib/pages/demo.dart
+++ b/gallery/gallery/lib/pages/demo.dart
@@ -179,6 +179,7 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
     final colorScheme = Theme.of(context).colorScheme;
     final iconColor = colorScheme.onSurface;
     final selectedIconColor = colorScheme.primary;
+    final appBarPadding = 20.0;
 
     final appBar = AppBar(
       backgroundColor: Colors.transparent,
@@ -190,6 +191,7 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
         },
       ),
       actions: [
+        SizedBox(width: appBarPadding),
         if (_hasOptions)
           IconButton(
             icon: FeatureDiscovery(
@@ -245,6 +247,7 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
                 _state == _DemoState.fullscreen ? selectedIconColor : iconColor,
             onPressed: () => _handleTap(_DemoState.fullscreen),
           ),
+        SizedBox(width: appBarPadding),
       ],
     );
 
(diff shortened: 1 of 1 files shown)


perclasson:
I don't think this is the correct screenshot - seems to be the Banner demo instead of the AppBar?

Also can you add a screenshot for how it looks like on mobile?

gallery/gallery/lib/pages/demo.dart
```diff
@@ -245,6 +247,7 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
                 _state == _DemoState.fullscreen ? selectedIconColor : iconColor,
             onPressed: () => _handleTap(_DemoState.fullscreen),
           ),
+        SizedBox(width: appBarPadding),
```
rami-a: This would only solve the problem for the buttons on the right, but the back button on the left would still have the problem.

gallery/gallery/lib/pages/demo.dart
```diff
@@ -190,6 +191,7 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
         },
       ),
       actions: [
+        SizedBox(width: appBarPadding),
```
rami-a: I don't think this SizedBox is necessary.

guidezpl: It's just to keep things balanced/centered

gallery/gallery/lib/pages/demo.dart
guidezpl: Hence the SizedBox on the left :)

guidezpl:
> I don't think this is the correct screenshot - seems to be the Banner demo instead of the AppBar?
> 
> Also can you add a screenshot for how it looks like on mobile?

This is the appBar for all demos, and sure!

gallery/gallery/lib/pages/demo.dart
guidezpl: Oh I see what you mean

guidezpl: It looks like the back button is already afforded enough space for the ripple to show

rami-a:
@guidezpl the issue is more evident in light theme but look at the ink ripple in the following screenshots

Screen Shot 2020-01-28 at 9 51 02 AM
Screen Shot 2020-01-28 at 9 51 09 AM

gallery/gallery/lib/pages/demo.dart
```diff
@@ -179,15 +179,19 @@ class _DemoPageState extends State<DemoPage> with TickerProviderStateMixin {
     final colorScheme = Theme.of(context).colorScheme;
     final iconColor = colorScheme.onSurface;
     final selectedIconColor = colorScheme.primary;
+    final appBarPadding = 20.0;
```
rami-a: Only do this if it is desktop, otherwise it'll look strange on mobile

guidezpl: Good call

Issue #245, opened 2020-01-21
[ Provider Doc ] ChangeNotifier inside update


tperraut:
https://github.com/flutter/samples/blob/b7ec38af9cb57fa39375bf793b25e1d4b43e8975/provider_shopper/lib/main.dart#L32

I was implementing a little test project with official Flutter doc about state management and looking at your awesome example. Thank you a lot btw.

While reading the ChangeNotifierProxyProvider documentation, I found this : 
```
/// - **DON'T** create the [ChangeNotifier] inside `update` directly.
///
///   This will cause your state to be lost when one of the values used updates.
///   It will also cause uncesserary overhead because it will dispose the
///   previous notifier, then subscribes to the new one.
///
///  Instead reuse the previous instance, and update some properties or call
///  some methods.
```
So, if creating a new ChangeNotifier inside update callback is a bad practice as the doc says, you could change a bit the sample behavior here or add a little comment.

RedBrogdon:
Huh.

It looks like @rrouselgit's chief concern is that you don't lose the state contained in your ChangeNotifier by tossing it out and creating a new one. Though it does create a new instance, the app doesn't do that, since it uses a copy constructor of sorts to maintain the selected item state:

https://github.com/flutter/samples/blob/master/provider_shopper/lib/models/cart.dart#L20

@filiph What do you think? I appreciate both that the code is correct and that @tperraut is making a valid point. Maybe a one-liner comment on L31 would help explain what's going on in the update method?

https://github.com/flutter/samples/blob/master/provider_shopper/lib/main.dart#L31

rrousselGit:
There's more than losing the state to that statement
Some background: https://github.com/rrousselGit/provider/issues/260

By cloning the instance instead of reusing it, it'll force an update on dependents for no reason.

It may also be very dangerous with click handlers.
The community likes doing:

```dart
class MyButton {
  ...

  Widget build(context) {
    final myNotifier = Provider.of<MyNotifier>(context, listen: false);

   return RaisedButton(
     onPressed: myNotifier.increment,
   );
  }
}
```

But if since they pass `listen: false`, the click handler _may_ point to a previous instance of the notifier.

It's effectively their fault, but not cloning simply removed the possibility of making a mistake.

rrousselGit:
A more concrete dangerous use-case is asynchronous operations:

```dart
class MyNotifier with ChangeNotifier {
  String property;  

  Future<void> doSomething() async {
    property = await something();
  }
}
```


[...]

PR #298, opened 2020-01-31
Follow best practice of ProxyProvider updating

filiph:
The addresses feedback from https://github.com/flutter/samples/issues/245. Instead of constructing a new `CartModel`, we merely update the `catalog` field.

We no longer need a fancy constructor, and `CartModel._catalog` cannot be final any more.

This change also upgrades to `pkg:provider` 4.0.2, just to stay on top of things (it's was a no-op in the case of this sample).

### DIFF
--- a/provider_shopper/lib/main.dart
+++ b/provider_shopper/lib/main.dart
@@ -27,9 +27,11 @@ class MyApp extends StatelessWidget {
         // of ChangeNotifierProvider. Moreover, CartModel depends
         // on CatalogModel, so a ProxyProvider is needed.
         ChangeNotifierProxyProvider<CatalogModel, CartModel>(
-          create: (context) => CartModel.empty(),
-          update: (context, catalog, previousCart) =>
-              CartModel(catalog, previousCart),
+          create: (context) => CartModel(),
+          update: (context, catalog, cart) {
+            cart.catalog = catalog;
+            return cart;
+          },
         ),
       ],
       child: MaterialApp(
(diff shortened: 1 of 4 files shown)


provider_shopper/lib/main.dart
```diff
@@ -27,9 +27,11 @@ class MyApp extends StatelessWidget {
         // of ChangeNotifierProvider. Moreover, CartModel depends
         // on CatalogModel, so a ProxyProvider is needed.
         ChangeNotifierProxyProvider<CatalogModel, CartModel>(
-          create: (context) => CartModel.empty(),
-          update: (context, catalog, previousCart) =>
-              CartModel(catalog, previousCart),
+          create: (context) => CartModel(),
+          update: (context, catalog, cart) {
+            cart.catalog = catalog;
```
rrousselGit: Nit but I'd use the `..` operator instead of making a block

RedBrogdon: To offer a counterpoint, in a sample that's aimed at beginners (like this one), we sometimes stretch out the syntax to maximize readability for new folks.

Entirely your call which way to go, though, Filip.

filiph: Yes, I had the cascade operator there at first but went with the more verbose syntax for the reason @RedBrogdon mentions. We're not teaching Dart niceties here, so I go with as "vanilla" syntax as possible.

nitinmehtaa: @filiph @RedBrogdon Thanks a ton, As a beginner user it's very hard to understand these kind of syntax `..` 
vanilla syntax rocks:)

PR #299, opened 2020-01-31
[Gallery] Use google_fonts

guidezpl:
In all but two places, we can use the google_fonts package. 

Merging is blocked on https://github.com/material-foundation/google-fonts-flutter/pull/58 (for web support)


gallery/gallery/pubspec.yaml
```diff
@@ -35,6 +36,7 @@ dev_dependencies:
 
 flutter:
   assets:
+    - fonts/google_fonts/
```
rami-a: Why is this needed?

rami-a: Also I see all the fonts are still included in the directory, is that how it's supposed to work?

guidezpl: The google_fonts package will first check assets for font files before fetching them and caching them. See https://github.com/material-foundation/google-fonts-flutter/blob/1255da8321fb84be2ffe56fcc5f792eb1dadf83d/README.md#bundling-font-files-in-your-applications-assets for more details

guidezpl: Yes, see above

guidezpl: We could chose not to bundle them, but then you'd get a flicker as fonts load the first time, and the fonts wouldn't load if you opened the app offline for the first time.

rami-a: Gotcha, the readme mentions that the directory should be `- google_fonts/`, but I guess ours works with the extra level in there? Or should we modify that too

guidezpl: It can use any directory. I just figured it made more sense to put them under fonts/

PR #301, opened 2020-01-31
[Gallery] Fix slow motion setting announcement

guidezpl:
Fixes a bug with the slow motion setting a11y announcement

### DIFF
--- a/gallery/gallery/lib/pages/home.dart
+++ b/gallery/gallery/lib/pages/home.dart
@@ -700,6 +700,7 @@ class _CarouselState extends State<_Carousel>
         controller: _controller,
         itemCount: widget.children.length,
         itemBuilder: (context, index) => builder(index),
+        allowImplicitScrolling: true,
       ),
       controller: widget.animationController,
     );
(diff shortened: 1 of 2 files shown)


gallery/gallery/lib/pages/home.dart
```diff
@@ -700,6 +700,7 @@ class _CarouselState extends State<_Carousel>
         controller: _controller,
         itemCount: widget.children.length,
         itemBuilder: (context, index) => builder(index),
+        allowImplicitScrolling: true,
```
rami-a: If i remember correctly, this flag was new and we couldn't use it yet because it wasn't in stable. Is this no longer the case?

guidezpl: I wasn't sure, so I figured I'd let CI tell me

guidezpl: Looks like it's in beta but not stable yet

rami-a: Ok, let's hold off on this until the next stable release then

guidezpl: cool, I'll repurpose this PR

PR #306, opened 2020-02-01
add web_dashboard sample

johnpryan:
This adds a responsive master-detail app with browser URL navigation. The app combines a new `AdaptiveScaffold` widget that switches between a `Drawer`, `NavigationRail`, and `BottomAppBar` for navigation. The body of the `Scaffold` is a `Navigator` that handles three routes using `onGenerateRoute`. The URL stays in sync with the routing in the app.

Right now, there are two Navigators, the root Navigator and the child Navigator. The root Navigator forwards the initial route to the child using `onGenerateInitialRoutes` and passing the `RouteSettings`. The `AppShell` widget's keeps the AdaptiveScaffold Navigator in sync.


experimental/web_dashboard/lib/src/adaptive_scaffold.dart
```diff
@@ -0,0 +1,152 @@
...
+  final IconData icon;
+
+  const AdaptiveScaffoldDestination({
+    @required this.title,
+    @required this.icon,
+  });
+}
+
+/// A widget that adapts to the current display size, displaying a [Drawer],
+/// [NavigationRail], or [BottomAppBar]. Navigation destinations are defined in
```
RedBrogdon: nit: BottomNavigationBar rather than BottomAppBar?

experimental/web_dashboard/lib/src/adaptive_scaffold.dart
```diff
@@ -0,0 +1,152 @@
...
+        ],
+      );
+    }
+
+    // Show a navigation rail
+    if (_isMediumScreen(context)) {
+      return Scaffold(
+        appBar: AppBar(title: widget.title,),
+        body: Row(
+          children: <Widget>[
```
RedBrogdon: nitL: No type specifier necessary here.

experimental/web_dashboard/lib/main.dart
```diff
@@ -0,0 +1,204 @@
...
+      onNavigationIndexChange: (idx) {
+        setState(() {
+          _currentIndex = idx;
+        });
+
+        // Use the inner Navigator to change the route.
+        var navState = _navigatorKey.currentState as NavigatorState;
+        navState.pushNamed(_routeForIndex(idx));
+      },
+      floatingActionButton: FloatingActionButton(
```
RedBrogdon: I can imagine wanting this button to have different effects based on which page is loaded into the navigator inside the AdaptiveScaffold. Is that something that would be supported?

johnpryan: I don't think the AdaptiveScaffold would cover that case. The user would be responsible for applying any effects (or hiding the FAB altogether) based on the `currentIndex`.

johnpryan:
closing in favor of https://github.com/flutter/samples/pull/333, which moves this into an experiments/ directory

PR #318, opened 2020-02-06
[Gallery] Implement Crane middle front layer

guidezpl:
This PR implements the expected design for Crane's middle front layer being higher than the others. I opted not to create recreate custom version of existing Material widgets, instead using an OverflowBox. The only downside is there is a small portion of the middle tab that cannot be interacted with (see below in red). Since everything else remains functional, I think this is acceptable 

Simulator Screen Shot - iPhone 11 Pro Max - 2020-02-06 at 10 22 53
Simulator Screen Shot - iPhone 11 Pro Max - 2020-02-06 at 10 22 58
Simulator Screen Shot - iPhone 11 Pro Max - 2020-02-06 at 10 22 51

Closes https://github.com/material-components/material-components-flutter-gallery/issues/133


guidezpl:
> LGTM. I think the widgets are getting pretty big so if they can be broken down into smaller more readable parts that would probably be good but not a blocker.

I tried something like that and it felt less readable so I'm going to leave it as is for now

PR #319, opened 2020-02-06
[Gallery] Add screenshots, edit README

guidezpl:
- Updating README, adding a screenshot & fixing links
- Adding a file with a list of screenshots

Open to suggestions on updating the copy.


gallery/screenshots.md
```diff
@@ -0,0 +1,18 @@
+# Screenshots
```
rami-a: Should we link to this markdown file from the main README?

guidezpl: I don't think it's necessary. It's more of a place to point people to when they ask us for screenshots

PR #320, opened 2020-02-06
3 commits: web sample tweaks

kevmoo:


RedBrogdon:
@kevmoo, now that you're back in action, we should talk about where we're going with the web samples and whether some of these should be removed from the repo to make room for new ones.

The DPEs will be in SEA next week, if that's a good time for you.

kevmoo:
> @kevmoo, now that you're back in action, we should talk about where we're going with the web samples and whether some of these should be removed from the repo to make room for new ones.
> 
> The DPEs will be in SEA next week, if that's a good time for you.

Great to hear! Can't wait to talk!

PR #326, opened 2020-02-13
Searchbar style fix

jurgenizer:
A styling fix to remove the white area behind the CupertinoTextField. 
The fix makes search bar look like default iOS search bar.

before:
image1

after:
image0


veggieseasons/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -374,6 +374,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = 8;
+				DEVELOPMENT_TEAM = "";
```
RedBrogdon: These lines can just be taken out.

veggieseasons/lib/widgets/search_bar.dart
```diff
@@ -39,6 +39,12 @@ class SearchBar extends StatelessWidget {
               child: CupertinoTextField(
                 controller: controller,
                 focusNode: focusNode,
+                 decoration: BoxDecoration( // to remove white area behind text field
```
RedBrogdon: @LongCatIsLooong/@xster, what's your thought here?

Is this a flaw in how the original code is using CupertinoTextField?

RedBrogdon:
Ah, it looks like our automated tests have flagged the formatting in `search_bar.dart`. It looks like one of the new lines isn't indented correctly. Just run `flutter format` on that file and it'll be corrected.

veggieseasons/ios/Runner.xcodeproj/project.pbxproj
jurgenizer: lines removed

veggieseasons/lib/widgets/search_bar.dart
```diff
@@ -39,6 +39,11 @@ class SearchBar extends StatelessWidget {
               child: CupertinoTextField(
                 controller: controller,
                 focusNode: focusNode,
+                decoration: BoxDecoration(
```
LongCatIsLooong: nit: [`CupertinoTextField`'s documentation](https://master-api.flutter.dev/flutter/cupertino/CupertinoTextField-class.html) states that "If you set the decoration property to null, the decoration will be removed entirely".  Maybe we should just set it to `null`?

jurgenizer: @LongCatIsLooong I agree. I have set the decoration property to null. I was not sure how to do it so I just did `decoration: BoxDecoration(),`

veggieseasons/lib/widgets/search_bar.dart
```diff
@@ -39,6 +39,7 @@ class SearchBar extends StatelessWidget {
               child: CupertinoTextField(
                 controller: controller,
                 focusNode: focusNode,
+                decoration: BoxDecoration(),
```
LongCatIsLooong: According to the documentation `decoration` is nullable. So:
```suggestion
                decoration: null,
```

jurgenizer: Thanks! I updated it now

RedBrogdon:
Awesome. Sounds like we've got the right change. @jurgenizer, I think there are still some changes in the project file that need to be taken out. Other than that, though, this looks ready to land!

PR #332, opened 2020-02-19
Updates the framework build in travis script to use an absolute path.

RedBrogdon:
See this issue for detail:

https://github.com/flutter/flutter/issues/46993

This should produce no noticeable changes other than working tests. 😄


RedBrogdon:
> LGTM. Did you mean to include the Podfile.lock changes?
> Thanks for your help debugging this!

Thanks for the help finding the issue!

Re: the lock files, they were updated when I was testing everything, and I figured I'd leave the changes in. These samples are still so new they might as well be as up to date as possible.

Issue #322, opened 2020-02-07
Veggie Seasons UI doesn't handle dark mode well


RedBrogdon:
Title says it all. We need to update the app to deal with iOS's dark mode properly rather than displaying black-on-black text in several places within the app.

abd99:
I have started working on the fix for this issue.
I am planning on changing the declared styles to a method and passing it the BuildContext as a parameter.

Eg:
```
static const headlineName = TextStyle(
    color: Color.fromRGBO(0, 0, 0, 0.9),
    fontFamily: 'NotoSans',
    fontSize: 24,
    fontStyle: FontStyle.normal,
    fontWeight: FontWeight.bold,
  );
```

Will turn into:
```
static TextStyle headlineName(BuildContext context) => TextStyle(
        color: CupertinoTheme.of(context).textTheme.textStyle.color,
        fontFamily: 'NotoSans',
        fontSize: 24,
        fontStyle: FontStyle.normal,
        fontWeight: FontWeight.bold,
      );
```
Do you think this is a good solution? Or is there a better solution?
Please let me know @RedBrogdon.

RedBrogdon:
To be quite honest, the way I originally set up the fonts and colors in Veggieseasons was not a great architecture. A lot of the support for theming in Cupertino didn't exist 12-16 months ago when the app was written.

Now, though, it should be possible to use CupertinoTextThemeData:

https://github.com/flutter/flutter/blob/210f4d83136a2eae773bda471db90ac27676f662/packages/flutter/lib/src/cupertino/text_theme.dart#L110

And CupertinoThemeData:

https://github.com/flutter/flutter/blob/210f4d83136a2eae773bda471db90ac27676f662/packages/flutter/lib/src/cupertino/theme.dart#L146

to do the job right.

If you look at an app like the Material Gallery, rather than having a file that sets up a bunch of font/color constants, it has a file that constructs its ThemeData objects for light and dark theme:

https://github.com/flutter/gallery/blob/master/lib/themes/gallery_theme_data.dart

So, while the change you're suggesting *would* work, there's an opportunity here to improve the app in a more fundamental way. Since these apps are meant to serve as examples of the best ways to do things with Flutter, that's where we should be aiming. 

If that's more work than you were looking to contribute, I completely understand, but I do feel it's the right way to go here.

RedBrogdon:
Some further context: 

[...]

PR #334, opened 2020-02-21
Fix issue #322

kmrakash:
Change some of required text color to "color: Color(0xFF303030)" which is dark enabled

### DIFF
--- a/veggieseasons/lib/styles.dart
+++ b/veggieseasons/lib/styles.dart
@@ -49,7 +49,7 @@ abstract class Styles {
   );
 
   static const cardCategoryText = TextStyle(
-    color: Color.fromRGBO(255, 255, 255, 0.9),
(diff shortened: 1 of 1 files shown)


kmrakash:
@googlebot  I signed it!

RedBrogdon:
@kmrakash, thanks for submitting a pull request!

I'm afraid issue #322 is much larger than just a change of text color changes, though. For a start, we'd need font colors that are _different_ depending on whether dark mode is activated. We also need a number of larger changes to the settings screen and some other widgets.

For example:

image

Those widgets need their backgrounds changed, text colors changed, and a number of other small details updated.

If you're interested in continuing here, that's great! But we can't land this pull request as-is.

RedBrogdon:
I'm going to go ahead and close this for now. Feel free to reopen if you're still interested!

Issue #280, opened 2019-04-04
[Veggieseasons] Let overscroll happen in the header section


xster:
Instead of between the header and the list under it. It should also zoom out the image a bit. See the iOS podcast app. One a podcast and drag down.

RedBrogdon:
Additional notes:

On the detail popup, rather than having the list below separate from the image header, instead the header should grow taller to absorb the scroll and the image should blur slightly.

PR #338, opened 2020-02-23
addded overscroll in header section

AjayThakur12:
Fixed: #280 

Changes: Added _buildHelper in ListView in details.dart file. Also increased image size

Screenshots
Screenshot 2020-02-23 at 8 54 15 PM

Screenshot 2020-02-23 at 8 54 24 PM

### DIFF
--- a/veggieseasons/lib/screens/details.dart
+++ b/veggieseasons/lib/screens/details.dart
@@ -284,7 +284,7 @@ class _DetailsScreenState extends State<DetailsScreen> {
             child: ListView(
               children: [
                 _buildHeader(context, appState),
-                SizedBox(height:20),
+                SizedBox(height: 20),
                 CupertinoSegmentedControl<int>(
                   children: {
                     0: Text('Facts & Info'),
--- a/veggieseasons/lib/screens/details.dart
+++ b/veggieseasons/lib/screens/details.dart
@@ -245,7 +245,7 @@ class _DetailsScreenState extends State<DetailsScreen> {
     final veggie = model.getVeggie(widget.id);
 
     return SizedBox(
-      height: 400,
+      height: 150,
       child: Stack(
         children: [
           Positioned(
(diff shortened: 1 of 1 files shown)


AjayThakur12:
> @googlebot I signed it!

veggieseasons/lib/screens/details.dart
```diff
@@ -245,7 +245,7 @@ class _DetailsScreenState extends State<DetailsScreen> {
     final veggie = model.getVeggie(widget.id);
 
     return SizedBox(
-      height: 150,
+      height: 400,
```
RedBrogdon: I'm not sure why this change was made. In the original design, just a portion of the image was shown, since it's enough to give some visual appeal to the popup but doesn't push the detailed information for the veggie too far down the page. With this change, the image can take up > 1/2 the screen, pushing the details quite far down.

Unless you have a strong reason for keeping this change, I'd recommend reverting it.

AjayThakur12: Screenshot 2020-03-03 at 12 40 28 AM

thank you @RedBrogdon  for review my PR. I changed the size of image because it is asked in issue.
If your are recommending then i revert changes

RedBrogdon: Ah, got it. I think we have a miscommunication. 😄 

I wrote that note after talking with the colleague who wrote the original bug report. What he meant was that, ideally, we'd implement an animated header that grew when the user pulled the screen down, similar to this one:

https://stackoverflow.com/questions/54580218/stretchy-header-effect

For right now, I recommend just taking the image size change out, and we can merge the rest of the changes you've included. If you have time later to work on another change (and you're interested in learning how Flutter's slivers work), you could try to implement the stretchy header in another PR.

AjayThakur12: reverted image size @RedBrogdon  thank you for your feedback . I will try to implement the stretchy header.

RedBrogdon: It's a difficult bit of UI to get right, so it may take some time. 😄

PR #345, opened 2020-03-01
feat: add more animations examples

GOVINDDIXIT:
Added animation example 
### GIF
- AnimatedList

GIF-200305_012304


RedBrogdon:
@GOVINDDIXIT, thanks for submitting a PR.

The code in the PR doesn't appear to match the rest of the examples in the animations app. They use a different navigation paradigm for getting into and out of the screens, for example.

I'm also not quite sure what topics they're meant to be demonstrating. The "parenting" animation, for example, doesn't seem to relate to any particular widget or animation technique ("parent animation" isn't a term I've seen used before).

The bottom line is that extensive changes would be needed before these additions could be merged into the codebase. If that seems onerous, another option is simply to make your own animation sample project in another repo. The samples you see here are no more "official" than any others, and there are a number of projects on [Awesome Flutter](https://github.com/Solido/awesome-flutter) that get as much attention as this repo does!

GOVINDDIXIT:
@RedBrogdon Thanks for the review.
I will remove these animations from here and move them to another repo of my own. For this PR, I would add something related to a particular widget.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,137 @@
...
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+
+class AnimatedListDemo extends StatefulWidget {
+  static String routeName = '/basics/08_animated_list';
+
+  _AnimatedListDemoState createState() => _AnimatedListDemoState();
+}
+
+class _AnimatedListDemoState extends State<AnimatedListDemo> {
```
RedBrogdon: This state object is holding the GlobalKey, but no other actual state. It needs to contain the "master" copy of the list of people. You can fix this by:

* Adding a `listData` field that holds a list of user models
* Initializing it with the data from the current listData array below (try renaming that one to `initialListData` or something similar).
* When the code adds or removes an item, update the list held by this stateful widget and *then* call insert or remove on the AnimatedListState. These can be done in a single `setState` call.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,137 @@
...
+  final GlobalKey<AnimatedListState> _listKey = GlobalKey();
+
+  void addUser() {
+    int index = listData.length;
+    listData.add(listData[Random().nextInt(5)]);
+    _listKey.currentState
+        .insertItem(index, duration: Duration(milliseconds: 500));
```
RedBrogdon: While you've done a good job customizing the animations, I'd recommend dropping this duration change. The default is 300ms, and any complexity that can be pulled out of a sample helps the people trying to learn from it.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,137 @@
...
+            axisAlignment: 0.0,
+            child: _buildItem(user),
+          ),
+        );
+      },
+      duration: Duration(milliseconds: 600),
+    );
+  }
+
+  Widget _buildItem(UserModel user, [int index]) {
```
RedBrogdon: Is it possible for index to be null? If not, I'd make it a required parameter and remove the ternary expression on line 54.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,137 @@
...
+      ),
+      onLongPress: index != null ? () => deleteUser(index) : null,
+    );
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text("Animated List Demo"),
```
RedBrogdon: The other demos (as far as I'm aware) don't use AppBar titles, so this should probably come out.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,137 @@
...
+        centerTitle: true,
+        leading: IconButton(
```
RedBrogdon: This icon button is appear in the place normally used for the AppBar's back button. Rather than putting it into the `leading` property, add it to the AppBar's `actions` list. That way it'll appear on the right side, and the back button will still be present as in the other demos.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,137 @@
...
+}
+
+class UserModel {
+  UserModel({this.firstName, this.lastName, this.profileImageUrl});
+
+  String firstName;
+  String lastName;
+  String profileImageUrl;
+
+  @override
```
RedBrogdon: While it's a good coding practice, the equality and hashCode overrides here aren't necessary for this example. You can pull them out.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,137 @@
...
+  int get hashCode =>
+      firstName.hashCode ^ lastName.hashCode ^ profileImageUrl.hashCode;
+}
+
+List<UserModel> listData = [
+  UserModel(
+    firstName: "Govind",
+    lastName: "Dixit",
+    profileImageUrl:
+        "https://ca.slack-edge.com/TADUGCD9D-UC5F6HJ6T-gfbe5883d03f-1024",
```
RedBrogdon: This image URL adds a dependency for the sample. If that URL were to ever stop working, it would break the app.

Rather than using an image for the avatar, try using a material Icon instead:

```
CircleAvatar(
   child: Icon(Icons.person),
),
```

The icons are guaranteed to be available, and can be loaded without any network calls. This would also allow you to simplify UserModel from three fields to two.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,137 @@
...
+            );
+          },
+        ),
+      ),
+    );
+  }
+}
+
+class UserModel {
+  UserModel({this.firstName, this.lastName, this.profileImageUrl});
```
RedBrogdon: Could this be a `const` constructor, and the fields made `final`? I don't believe these objects are mutated anywhere in the code.

animations/README.md
```diff
@@ -23,6 +23,7 @@ Building blocks and patterns
 6. **CustomTweenDemo** Demonstrates how to extend `Tween`.
 7. **TweenSequenceDemo** Demonstrates how to use `TweenSequence` to build a
 button that changes between different colors.
+8. **AnimatedList** Demonstrate how to use `AnimatedList`
```
RedBrogdon: nit: This needs an "s" on "demonstrates" and a period at the end:

    Demonstrates how to use `AnimatedList`.

animations/lib/src/basics/08_animated_list.dart
RedBrogdon: This call is using the length of the initial list as the index for adding a new one. That means that if you delete a person and *then* try to add one, it throws an exception. This can be fixed by keeping the "master" copy of the list as a field in AnimatedListDemoState.

animations/lib/src/basics/08_animated_list.dart
GOVINDDIXIT: @RedBrogdon Thanks for the review. 😃
I have made all other requested changes but I am not fully sure that I have made this particular change as requested. Pls review if something more need to be changed.

GOVINDDIXIT:
> This state object is holding the GlobalKey, but no other actual state. It needs to contain the "master" copy of the list of people. You can fix this by:
> 
> * Adding a `listData` field that holds a list of user models
> * Initializing it with the data from the current listData array below (try renaming that one to `initialListData` or something similar).
> * When the code adds or removes an item, update the list held by this stateful widget and _then_ call insert or remove on the AnimatedListState. These can be done in a single `setState` call.

@RedBrogdon Thanks for the review. 😃
I have made all other requested changes but I am not fully sure that I have made this particular change as requested. Pls review if something more need to be changed.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,118 @@
...
+
+class AnimatedListDemo extends StatefulWidget {
+  static String routeName = '/basics/08_animated_list';
+
+  _AnimatedListDemoState createState() => _AnimatedListDemoState();
+}
+
+class _AnimatedListDemoState extends State<AnimatedListDemo> {
+  final GlobalKey<AnimatedListState> _listKey = GlobalKey();
+  List<UserModel> initialListData = listData;
```
RedBrogdon: These two things should have their names swapped:

`final listData = initialListData`

And line 97 should be:

`List<UserModel> initialListData = [`

The list at the bottom of the file is the initial (starting) data, so it should have that name.

animations/lib/src/basics/08_animated_list.dart
```diff
@@ -0,0 +1,118 @@
...
+
+  void addUser() {
+    setState(() {
+      int index = initialListData.length;
+      _listKey.currentState
+          .insertItem(index, duration: Duration(milliseconds: 300));
+      initialListData.add(
```
RedBrogdon: I would move this statement to be above the call to `insertItem`. Better to update the list before you tell the AnimatedListView to begin animating.

GOVINDDIXIT:
@RedBrogdon Thanks for the review. 😃
I have made the requested changes

PR #346, opened 2020-03-02
Image size reduced

ayushs136:
Image size reduced by 8%


RedBrogdon:
While reducing image size is certainly a good aim, a pull request like this is too large to be reviewed effectively. We do not have image-based goldens for the samples repository, so testing would need to be done manually (e.g. open each app on several devices to make sure the images at each resolution are being decoded and displayed properly.

If you'd like to keep going with this effort, that's great, but it would need to be done one app at a time.

PR #351, opened 2020-03-04
[Gallery] Remove initial splash

clocksmith:


gallery/lib/pages/splash.dart
```diff
@@ -85,21 +83,10 @@ class _SplashPageState extends State<SplashPage>
         duration: Duration(
           milliseconds: splashPageAnimationDurationInMilliseconds,
         ),
-        value: 1,
```
perclasson: Should we not remove this duration here? As it otherwise would wait before showing the rest of the page.

There's also other changes related to `splashPageAnimationDurationInMilliseconds` that can should be removed if we remove the splash animation.

gallery/lib/pages/splash.dart
```diff
@@ -85,21 +83,10 @@ class _SplashPageState extends State<SplashPage>
...
         vsync: this)
       ..addListener(() {
         this.setState(() {});
       });
-    if (widget.isAnimated) {
-      _launchTimer = Timer(
```
perclasson: We can remove the `_launchTimer` from `dispose` above since it's no longer used.

clocksmith:
Yes the idea is to remove the initial splash page for all platforms (web, mobile, etc), but still allow the user to see the animations by clicking the top portion of the screen

PR #333, opened 2020-02-20
add web_dashboard API

johnpryan:
This is the latest version of the web_dashboard sample. Replaces https://github.com/flutter/samples/pull/306

PR #352, opened 2020-03-05
[Gallery] Revert provisioning changes

guidezpl:
Reverting accidental change in https://github.com/flutter/samples/pull/333/ that broke Gallery builds


gallery/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -375,10 +375,10 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
-				CODE_SIGN_IDENTITY = "Apple Development";
-				CODE_SIGN_STYLE = Automatic;
+				CODE_SIGN_IDENTITY = "iPhone Developer";
```
pennzht: This line (and 2 other `CODE_SIGN_IDENTITY = "iPhone Developer";`) were not in the original file. Why are they added?

guidezpl: Not sure, but works without. Removed so its identical as before

PR #359, opened 2020-03-06
Add samples index

johnpryan:
This adds the new index and sets up `package:peanut` tooling to host the web demos alongside the samples.

[demo deployment using peanut](http://jpryan.me/samples/#)


clocksmith:
Nice!

web/samples_index/lib/src/templates.dart
```diff
@@ -0,0 +1,219 @@
+import 'dart:convert';
+
+import 'data.dart';
+import 'util.dart' as util;
+
+String _e(String s) =>
+    HtmlEscape(HtmlEscapeMode.attribute).convert(s);
+String _ea(String s) => HtmlEscape(HtmlEscapeMode.element).convert(s);
```
RedBrogdon: What does the "a" in `_ea` signify?

web/samples_index/web/add_to_app.html
```diff
@@ -0,0 +1,72 @@
...
+      </button>
+        </div>
+        <div class="tags-container">
+          <div class="tags-label">
+            <i class="material-icons">local_offer</i>
+            <span>Tags</span>
+          </div>
+          <div class="tags">
+            <a href="./#?search=tag%3Aadvanced">advanced</a>
+        <a href="./#?search=tag%3Asample">sample</a>
```
RedBrogdon: Indenting looks a little weird in this file.

web/samples_index/web/add_a_drawer_to_a_screen.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
```
RedBrogdon: Are these HTML files in /web the generated ones built by the samples index grinder tasks? If so, are they intended to be regenerated each time someone goes to publish the index? And if *that's* so, could they be left out of source control?

web/samples_index/lib/src/templates.dart
johnpryan: I think these need to be renamed, originally they were _escape and _escapeAttribute I think

web/samples_index/web/add_a_drawer_to_a_screen.html
johnpryan: Yep, it looks like things are still working without them.

web/samples_index/web/add_to_app.html
johnpryan: there's some indentation code in `templates.dart` that might need some adjustments

web/samples_index/lib/src/templates.dart
johnpryan: fixed

web/samples_index/web/add_to_app.html
johnpryan: these generated files aren't committed to source control anymore, so I think this is a lower priority now.

johnpryan:
@RedBrogdon This is ready for another look.

web/_tool/common.dart
```diff
@@ -0,0 +1,35 @@
+
```
RedBrogdon: nit: copyright header

web/samples_index/.gitignore
```diff
@@ -0,0 +1,17 @@
+# firebase public directory
```
RedBrogdon: nit: newline

web/samples_index/lib/browser.dart
```diff
@@ -0,0 +1 @@
+export 'src/search.dart';
```
RedBrogdon: nit: copyright header for this and any other code file.

mono0926:
Is http://jpryan.me/samples final version URL?

I found these urls at https://github.com/flutter/samples/blob/master/web/readme.md, 

- https://flutter.github.io/samples/
- flutter.github.io/samples/#?platform=web

but both redirect to this Flutter Gallery 🤔 

image


https://flutter.github.io/samples/#/

PR #365, opened 2020-03-10
chore: remove deprecated code

GOVINDDIXIT:
- Remove deprecated code in the animations app, following [this](https://github.com/flutter/flutter/blob/60d0fef17ae60b88cb52e6158bda2d91b600299c/packages/flutter/lib/src/material/text_theme.dart#L70)

`title` -> `headline6`
`display1` -> `headline4`
`display2` ->  `headline3`

### DIFF
--- a/animations/lib/main.dart
+++ b/animations/lib/main.dart
@@ -117,7 +117,7 @@ class AnimationSamples extends StatelessWidget {
 
 class HomePage extends StatelessWidget {
   Widget build(BuildContext context) {
-    final headerStyle = Theme.of(context).textTheme.title;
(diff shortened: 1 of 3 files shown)


RedBrogdon:
@GOVINDDIXIT, thanks for the PR, but @AyushBherwani1998 is correct. The code in the samples in intended to run with the stable channel:

https://github.com/flutter/samples/blob/master/CONTRIBUTING.md#a-few-ground-rules

We'll need to update a few things when the next stable release is made, but that won't be for a while.

GOVINDDIXIT:
@RedBrogdon @AyushBherwani1998  Thanks for the info. I will change my flutter SDK to the latest stable version.

PR #367, opened 2020-03-11
[platform_design] fix: Polished LogOut dialog

### LINKED ISSUE #363: bug: dialog message and action buttons unsuitable for logout
## Adaptive Music App screenshot: (in platform_design)

One solution could be to have a landing page (login-like page) for new users so that we can use that page on logging out as well. Currently, we directly land on the homepage having the list of songs.
We also need to update the logout dialog message and buttons - "Go back" and "cancel" pretty much makes the same meaning to me.
 OR
We could remove the Log out button itself from the profile screen as it is pretty much unfunctional otherwise.

jddeep:
Fixes #363 

Polished LogOut dialog in the Adaptive Music App.

### DIFF
--- a/platform_design/lib/profile_tab.dart
+++ b/platform_design/lib/profile_tab.dart
@@ -160,8 +160,8 @@ class PreferenceCard extends StatelessWidget {
 }
 
 class LogOutButton extends StatelessWidget {
-  static const _logoutMessage =
-      Text("You can't actually log out! This is just a demo of how alerts work.");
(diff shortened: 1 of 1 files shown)


jddeep:
@RedBrogdon @kwalrath @johnpryan Reviews, please

johnpryan:
Can you make sure you've run `dartfmt` on this file? Thanks!

jddeep:
@johnpryan done

PR #364, opened 2020-03-10
[animations] new animation sample: Hero Animation Demo

jddeep:
new animation sample: Hero Animation Demo

Additional Context:
[Hero Class](https://api.flutter.dev/flutter/widgets/Hero-class.html)

PR #369, opened 2020-03-12
hero animation added in animation section

AjayThakur12:
Demonstrates how we can use  hero animation during page transition.


jddeep:
Hi, there is already a PR regarding this here-https://github.com/flutter/samples/pull/364

RedBrogdon:
@AjayThakur12, while I appreciate the effort here (and on your previous PR), I'm afraid this is a duplication of something already in progress.

This is a new problem, so I'm updating the contributor's guide with a new policy that anyone considering work on a bug or new feature file an issue *before* beginning work on it. That should help keep a situation like this from arising again.

PR #371, opened 2020-03-13
update samples_index UX

johnpryan:
- Use Google Sans
- add brand colors
- adjust spacing on elements
- align MDC Chips to spec
- update type labels to spec


web/samples_index/web/styles.scss
```diff
@@ -1,24 +1,47 @@
+// Copyright 2020 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file
+
+// Flutter brand colors
```
RedBrogdon: Nice idea.

PR #374, opened 2020-03-13
Adding issue request notice to contributor's guide

RedBrogdon:


domesticmouse:
Looks like we need moar quota:

```
Http error while creating test matrix: ResponseError 429: Insufficient test quota. See http://firebase.google.com/pricing/ for details.
```

PR #375, opened 2020-03-15
[animations] Added FadeTransition Demo

### LINKED ISSUE #370: [animations] Add AnimatedWidget sample - FadeTransition
Missing animation samples of various Animation widgets.

@RedBrogdon @johnpryan Do we plan to add samples of widget animations like [FadeTransition](https://api.flutter.dev/flutter/widgets/FadeTransition-class.html), [DecoratedBoxTransition](https://api.flutter.dev/flutter/widgets/DecoratedBoxTransition-class.html), [RotationTransition](https://api.flutter.dev/flutter/widgets/RotationTransition-class.html), etc.

If yes please do let me know. Would love to add samples for them here.

jddeep:
Added FadeTransition Demo

Fixes #370


jddeep:
@RedBrogdon @johnpryan Reviews, please. Thanks,

RedBrogdon:
@johnpryan, this seems like a good one for you to review. Let me know if you're too busy to get to it, though.

animations/lib/src/basics/08_fade_transition.dart
```diff
@@ -0,0 +1,83 @@
...
+      end: 0.0,
+    ).animate(_curve);
+
+    _animation.addStatusListener((status) {
+      if (status == AnimationStatus.completed) {
+        _controller.reverse();
+      } else if (status == AnimationStatus.dismissed) _controller.forward();
+    });
+
+    super.initState();
```
johnpryan: It's considered a best-practice to call `super.initState()` first so that any code in `initState` can assume things are initialized. The `dispose()` method is the opposite.

animations/lib/src/basics/08_fade_transition.dart
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @override
+  void dispose() {
+    _controller.dispose();
+    super.dispose();
```
johnpryan: 👍

animations/lib/src/basics/08_fade_transition.dart
```diff
@@ -0,0 +1,83 @@
+// Copyright 2019 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
+class FadeTransitionDemo extends StatefulWidget {
```
johnpryan: Consider adding a comment linking to the [AnimatedWidget docs](https://api.flutter.dev/flutter/widgets/AnimatedWidget-class.html) for examples of other common animated widgets.

animations/lib/src/basics/08_fade_transition.dart
```diff
@@ -0,0 +1,83 @@
...
+          opacity: _animation,
+          child: Icon(
+            Icons.star,
+            color: Colors.amber,
+            size: 300,
+          ),
+        ),
+      ),
+      floatingActionButtonLocation: FloatingActionButtonLocation.centerFloat,
+      floatingActionButton: FloatingActionButton.extended(
```
johnpryan: Consider changing this to a RaisedButton like `basics/03_animation_controller.dart` for consistency

animations/lib/src/basics/08_fade_transition.dart
```diff
@@ -0,0 +1,83 @@
...
+
+    _animation = Tween(
+      begin: 1.0,
+      end: 0.0,
+    ).animate(_curve);
+
+    _animation.addStatusListener((status) {
+      if (status == AnimationStatus.completed) {
+        _controller.reverse();
+      } else if (status == AnimationStatus.dismissed) _controller.forward();
```
johnpryan: It's a little cleaner to use the AnimationController API. In `onPressed`, you can do:

```dart
          _controller
              .animateTo(1.0)
              .then((value) => _controller.animateBack(0.0));
```

jddeep:
@johnpryan updated.

animations/lib/src/basics/08_fade_transition.dart
```diff
@@ -65,18 +62,13 @@ class _FadeTransitionDemoState extends State<FadeTransitionDemo>
         ),
       ),
       floatingActionButtonLocation: FloatingActionButtonLocation.centerFloat,
-      floatingActionButton: FloatingActionButton.extended(
+      floatingActionButton: RaisedButton(
```
johnpryan: This should probably be a sibling of the star widget in a `Column` instead of the `floatingActionButton` widget.

animations/lib/src/basics/08_fade_transition.dart
```diff
@@ -4,6 +4,10 @@
 
 import 'package:flutter/material.dart';
 
+/* 
+Refer to the AnimatedWidget docs here - https://api.flutter.dev/flutter/widgets/AnimatedWidget-class.html
+for examples of other common animated widgets. 
+*/
```
johnpryan: Consider changing this to a `//` comment, it's considered better style to avoid block comments for documentation: https://dart.dev/guides/language/effective-dart/documentation#dont-use-block-comments-for-documentation

jddeep:
@johnpryan Thanks :) Updated.

johnpryan:
@jddeep can you fix the merge conflicts?

Issue #376, opened 2020-03-16
[platform_design] Some of the UI elements turn invisible in dark mode


abd99:
I found the following issues in the platform_design app:
- The Settings show black on black text in dark mode.
- There are a couple of more widgets that don't go well with the dark mode throughout the app.

Screenshots:
![Screenshot_20200317-000736](https://user-images.githubusercontent.com/30566863/76789764-7d8f4580-67e3-11ea-855c-f7f930ac07ac.png)
![Screenshot_20200317-000747](https://user-images.githubusercontent.com/30566863/76789789-854eea00-67e3-11ea-8cb1-41a3af400f01.png)

I'm currently working on the solution. 

I've read the contribution guide but this is my first PR in a big repo. So, please bear with me.
I'm thinking of just changing color to grab from the theme wherever it is hardcoded.

Let me know what you think of it @RedBrogdon @johnpryan @domesticmouse

jddeep:
This has been an issue for the iOS version of VeggieSeasons app as well. There is already an issue open about that #322. @RedBrogdon should we toggle the app font colors with respect to the app theme brightness(light or dark).

RedBrogdon:
There are a few things that should happen here:

1) Audit all the apps in the repo to make sure there aren't any others with dark mode problems, and file issues for them.
2) Hard-code a bright theme for any that do, this app included. See #340 for how this was done for Veggieseasons.
3) Update each of those apps to properly respect dark mode.

That third task is the hardest, and for each app it should be done in several PRs rather than just one big one. It's too large a change to review otherwise.

abd99:
As per your suggestion @RedBrogdon, I have checked all the apps in the repo except for web and experimental apps for dark mode problems. As of now, none of the apps have that problem except for VeggieSeasons(which you solved temporarily). Most of them don't even support dark mode. Let me know if you want me to add dark mode support to any of them. I have started working on the fix for VeggieSeasons and have shared the solution I'm opting for in #322. Please let me know what you think of it.

RedBrogdon:
@abd99, is there any remaining short-term work for VeggieSeasons now that #322 is merged? I believe we can go ahead and close this if not.

abd99:
I think we are good for now. I'll close this issue

PR #377, opened 2020-03-16
[platform_design]Fix problems in dark mode

abd99:
I've fixed all the problems related to dark mode in #376.
I've simply changed the color property of the problem causing widgets to grab from the current theme.

Please have a look @RedBrogdon @johnpryan.

iOS:

Android:

### DIFF
--- a/platform_design/lib/main.dart
+++ b/platform_design/lib/main.dart
@@ -20,6 +20,7 @@ class MyAdaptingApp extends StatelessWidget {
         // Use the green theme for Material widgets.
         primarySwatch: Colors.green,
       ),
+      darkTheme: ThemeData.dark(),
       builder: (context, child) {
         return CupertinoTheme(
           // Instead of letting Cupertino widgets auto-adapt to the Material
--- a/platform_design/lib/widgets.dart
+++ b/platform_design/lib/widgets.dart
@@ -222,7 +222,7 @@ class SongPlaceholderTile extends StatelessWidget {
         child: Row(
           children: [
             Container(
-              color: Colors.grey[400],
+              color: Theme.of(context).disabledColor,
               width: 130,
             ),
             Padding(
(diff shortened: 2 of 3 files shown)


abd99:
@jddeep as per your request, I have added the screencasts to both iOS and Android. Please take a look.

platform_design/lib/widgets.dart
```diff
@@ -235,27 +235,27 @@ class SongPlaceholderTile extends StatelessWidget {
                   Container(
                     height: 9,
                     margin: EdgeInsets.only(right: 60),
-                    color: Colors.grey[300],
+                    color: Theme.of(context).disabledColor,
```
RedBrogdon: nit: Since these rows are stand-ins for text, something like `Theme.of(context).textTheme.body1.color` might be more appropriate asa color choice.

abd99: Actually, by using the text theme's color the views looked too bright and didn't go well with the dark theme. Nonetheless, I understand that they are stand-ins for text and using the text theme's color would be better for people looking at this repo. I've made the changes according to your comment. Please take a look.

tusharojha:
We can modify the default dark theme for android by providing background color of body to black. Therefore, it will save battery, better for amoled displays and match with iOS counter part.

What do you think?

abd99:
> We can modify the default dark theme for android by providing background color of body to black. Therefore, it will save battery, better for amoled displays and match with iOS counter part.
> 
> What do you think?

Good idea! I'll do it.

platform_design/lib/main.dart
```diff
@@ -20,6 +20,9 @@ class MyAdaptingApp extends StatelessWidget {
         // Use the green theme for Material widgets.
         primarySwatch: Colors.green,
       ),
+      darkTheme: ThemeData.dark().copyWith(
+        scaffoldBackgroundColor: Colors.black,
```
RedBrogdon: I'm not sure that changing the default scaffold background to black is a good idea.

* It can cause the background of the scaffold to merge with the black background of the back/home buttons at the bottom of the screen.
* One of the general rules of thumb for design is to avoid using pure black.
  - See https://ianstormtaylor.com/design-tip-never-use-black/
* In general, we try to have practice good code size economy in every sample, since the longer a sample is, the more effort a reader has to spend learning from it. For that reason, if an addition of new code isn't absolutely necessary, it's generally best to take it out.

tusharojha:
I am sorry! I was just giving a suggestion.
My goal is just to improve the app and user experience.

jddeep:
@tusharojha Hey, no issues at all. But before making any breaking change, it's always good to get a nod of the maintainers as well. We are always open to suggestions in this community 😊.

tusharojha:
I will always take of it!
Thanks.

abd99:
I'm sorry about that @RedBrogdon @jddeep. I thought about the simplicity factor that @RedBrogdon  mentioned, but again the change was really small, so I did it. I wasn't aware of the other reasons that it wasn't a good idea. Thanks for sharing it @RedBrogdon.  I'll make sure that I do not make a tweaking/breaking change without project maintainers approving it. Thanks for the support.

RedBrogdon:
@tusharojha & @abd99, there's no need to apologize. Flutter isn't the kind of community where we come down hard on each other. 😄 

I just wanted you to understand my thought process, so you could know why I was asking you to update the code. Even when I do need to put on my tech lead hat and say, "This code needs to be changed," it's intended to be constructive rather than critical.

abd99:
I'm so happy my first PR got merged. Thanks! @RedBrogdon @jddeep @tusharojha

RedBrogdon:
Thanks for your contribution. It's a great help.

Just to be clear, though, Flutter is an open source, community project. If you write a sample in your own repo, it's just as "real" or "official" as the code you see here. I wouldn't want anyone thinking that the contributors to this repository are more special than anyone else who's helping the Flutter community. 😄

Issue #372, opened 2020-03-13
[veggieseasons] Android subproject should be removed.


aadarsh-patel:
On the details screen, when users tap the back button of the device, the app gets closed instead of navigating back to the home screen.
Back button should behave like the close button given in the header section of the details screen.

![veggie_bug](https://user-images.githubusercontent.com/53324291/76656032-c1d8d680-6594-11ea-8cd5-7a07eaadcd62.gif)

aadarsh-patel:
If this is a valid issue then i would like to work on it.

RedBrogdon:
Thanks for filing this issue. The problem here is that Veggieseasons isn't designed or intended to be run on Android. Because Flutter is multiplatform, it *will* run, but it doesn't make sense for us to invest time in the Android experience. 

What should be done instead is:

* Remove the `android`  directory within the project
* Add `android` to the .gitignore in the project directory, so it won't return.

Doing both in one PR requires a little git trickery, but if you're up for it, please feel free to file one.

aadarsh-patel:
Sure. I'll try.
Time for some github tutorials

aadarsh-patel:
Hey @RedBrogdon ,
Just saw the description of VeggieSeasons app, There it is clearly mentioned that 
>  This application is designed specifically for iOS to showcase Cupertino package.
> It's not intended to be run on Android, web, or desktop.

This issue opened by me is nothing but a silly mistake.
VeggieSeason is a cool app that also showcases how to use `models`, `shared_preference,` etc. which is useful for android developers as well.

Due to the above reasons I don't think removing android support will be a good idea for the app.
And hence i am closing this issue.

aadarsh-patel:
Sorry for closing this issue without waiting for your response.

jddeep:
Removing the directory from the sample isn't ideal. Is it? :thinking: At least devs can get a feel of the sample if they have an android device with them solely. What do you think?

RedBrogdon:
Thanks for reopening, @aadarsh-patel. 

Removing the Android subproject doesn't make it *impossible* to build and run the app for Android. All you'd need to do is run `flutter create .` from within the project directory, the directory would be recreated in your local copy, and then you'd be ready to go.

[...]

PR #378, opened 2020-03-16
Android subproject removed

aadarsh-patel:
For issue https://github.com/flutter/samples/issues/372.

Here's what I did,

1.  Added this line `android` to `.gitignore` file.
2. Ran this command `git rm -r --cached android`
3. Committed and pushed to repository. 

Now the android directory was in my local repository but it got removed from remote repository.

One interesting thing that I noticed is that after deleting the android directory, My IDE was not giving option to launch the android emulator and also on running command `flutter run` I get the following output `No supported devices connected.`

However after running command `flutter create .` everything was back to normal and I still was able to run the app on android.


veggieseasons/.gitignore
```diff
@@ -1,3 +1,6 @@
+# App is designed specifically for iOS to showcase Cupertino package.
```
RedBrogdon: nit: A little extra detail here might help:

"This app is designed to show how to use Flutter's cupertino package to build UI for iOS. It's not intended to run on Android, desktop, or web."

aadarsh-patel:
Hey @RedBrogdon ,
Thanks for reviewing my PR.
I have done the requested changes.

PR #379, opened 2020-03-17
[platform_design] Added dark mode for Android Side/Platform

tusharojha:
As Android side of sample project platform_design is lagging with dark mode.  I have committed changes, and successfully added the dark mode for android platform as well.

I have verified that the app works perfectly and the dark mode for iOS platform is not affected.

Here is a demo of dark mode on Android: 
screen-record

@RedBrogdon Please have a look.

### DIFF
--- a/platform_design/lib/main.dart
+++ b/platform_design/lib/main.dart
@@ -19,6 +19,12 @@ class MyAdaptingApp extends StatelessWidget {
       theme: ThemeData(
         // Use the green theme for Material widgets.
         primarySwatch: Colors.green,
+        backgroundColor: Colors.white,
+      ),
+      darkTheme: ThemeData(
+        brightness: Brightness.dark,
+        primarySwatch: Colors.green,
+        backgroundColor: Colors.black,
       ),
       builder: (context, child) {
         return CupertinoTheme(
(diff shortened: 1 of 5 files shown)


tusharojha:
@googlebot I signed it!

jddeep:
@tusharojha Hi, Thanks for sending a PR :) But isn't this similar to #377?

tusharojha:
@jddeep This is not similar to that as the PR mentioned by you have fixed bugs in the dark mode of iOS side/platform but I have added dark mode to Android side/platform of the app.

As the android side/platform is not having dark mode functionality already.

abd99:
The PR #377 already fixes dark mode for iOS as well as Android. So much work wasn't required for this. You can check it out. This was the reason I didn't mention any platform in the PR. It works just fine for both of the platforms. All you need to do is switch on dark mode from the System Settings(Android 10 and above)

tusharojha:
@abd99 I have seen your code #377  but the dark mode for android added is the default dark mode from ThemeData.dark() which does not provide amoled dark theme but has grey background. 

In this PR  I have not touched anything which effects dark mode of iOS but the dark mode of Android is amoled compatible.

jddeep:
@tusharojha please check it. @abd99 is right. If you see changes have been made in the dart code. So how can you say that it will only affect iOS and not Android. If #377 at all has any issues for Android, we will direct it over there in that PR itself. Thanks for your contributions though. Appreciate it. You can close the PR here. Helps us tackle duplicacy :)

tusharojha:
@jddeep Okay Sir, I will take care of that in further contributions :)

PR #381, opened 2020-03-17
Replaced deprecated terms in animations with newer ones

aksayushx:
Updated deprecated term 'title' in animations/lib/main.dart
Updated deprecated term 'display2' in animations/lib/src/basics/03_animation_controller.dart
Updated deprecated term 'display1' in animations/lib/src/basics/02_page_route_builder.dart


aksayushx:
@AyushBherwani1998 Thanks for the review :)

Issue #382, opened 2020-03-17
[animations] Add Animation Sample


aksayushx:
Hey @RedBrogdon @johnpryan , 
I would like to add a simple Slide Transition widget demo to the basics/ of animations which is missing from it. 
Should I go ahead and begin working on it?

jddeep:
Hi @aksayushx Thanks for filing an issue. 
Quoting @johnpryan regarding this:
> We don't need all the transition widgets, just one example of an AnimatedWidget (e.g. FadeTransition). Providing a link to AnimatedWidget will let users discover more *Transition widgets.

So the work for the AnimatedWidget sample - FadeTransition in this case has been already going on in #375.

aksayushx:
No problem. 
I have an animation idea using CurvedAnimation which will be suitable for the misc section of animations. So I ll be working on that. Thanks ;)

RedBrogdon:
@aksayushx, can you tell us a little about what you have in mind, and how it would complement the demos that are already in the app?

aksayushx:
Hey @RedBrogdon I am almost done with it and it looks here is a gif of the animation .
![ezgif com-crop](https://user-images.githubusercontent.com/55887638/76909096-4f3b6400-68d0-11ea-8755-916d6ab53177.gif)
Should I add it to misc section and make a new PR?

RedBrogdon:
Maybe I can help by talking over some of the things I think about when creating a sample. This is how the process tends to work for me:

1) What is the goal for the sample?

This repo is maintained by Flutter's developer relations team, and a phrase we frequently use on the DevRel team is "[learning outcome](https://teaching.utoronto.ca/teaching-support/course-design/developing-learning-outcomes/what-are-learning-outcomes/)." A learning outcome is the goal for an educational exercise -- what a student should gain by going through it. In this case, what should a developer who reads through your CurvedAnimation sample come away with? Should they:

* Just know that CurvedAnimation exists? 
* Know how to call its constructor and create one?
* Know how to combine it with other classes that it would normally be used with?
* Understand what an Animation object is?
* Understand what "curve" means in this context?

Sample code is different than demo code or app code. It exists first and foremost to educate, so it's important to know exactly what educational goals you have.

2) What do I need to show in order to achieve the learning outcome?

Once you know what you're aiming at, you can start thinking about how to get it. How can you guide someone in a simple and easily understood way to the outcome you want them to achieve? Say you chose these two learning outcomes:

[...]

PR #385, opened 2020-03-19
Added CurvedAnimation Sample

aksayushx:
#382 
The sample includes a rotating flutter logo which generates random curves for forward and reverse rotations every time the animate button is pressed to demonstrate the use of different curves with CurvedAnimation. 
CurvedAnimation

### DIFF
--- a/animations/README.md
+++ b/animations/README.md
@@ -23,7 +23,9 @@ Building blocks and patterns
 6. **CustomTweenDemo** Demonstrates how to extend `Tween`.
 7. **TweenSequenceDemo** Demonstrates how to use `TweenSequence` to build a
 button that changes between different colors.
-8. **FadeTransitionDemo** Demonstrates how to use `FadeTransition`. 
+8. **FadeTransitionDemo** Demonstrates how to use `FadeTransition`.
+9. **AnimatedPositionedDemo** Demonstrates how to use `AnimatedPositioned`.
+10. **AnimatedSwitcherDemo** Demonstrates how to use `AnimatedSwitcher`. 
 
 ### Misc
 
--- a/animations/README.md
+++ b/animations/README.md
@@ -41,6 +41,7 @@ spring simulation.
 detection to drive an animation.
 - **AnimatedList** Demonstrates how to use `AnimatedList`
 - **HeroAnimationDemo** Demonstrates how to use `Hero` animation.
+- **CurvedAnimationDemo** Demonstrates how to use different curves in `CurvedAnimation`
 
 ## Other Resources
 
--- a/animations/README.md
+++ b/animations/README.md
@@ -12,7 +12,7 @@ Sample apps that showcasing Flutter's animation features
 
 Building blocks and patterns
 
-1. **AnimatedContainerDemo** Demonstrates how to use `AnimatedContainer`
+1. **AnimatedContainerDemo** Demonstrates how to use `AnimatedContainer`.
 2. **PageRouteBuilderDemo** Demonstrates how to use `Tween` and `Animation` to
 *build a custom page route transition.
 3. **AnimationControllerDemo** Demonstrates how to use an `AnimationController`.
@@ -41,9 +41,9 @@ it using a `PageRouteBuilder`.
 spring simulation.
 - **CardSwipeDemo** A swipeable card that demonstrates how to use gesture
 detection to drive an animation.
-- **AnimatedList** Demonstrates how to use `AnimatedList`
+- **AnimatedList** Demonstrates how to use `AnimatedList`.
 - **HeroAnimationDemo** Demonstrates how to use `Hero` animation.
-- **CurvedAnimationDemo** Demonstrates how to use different curves in `CurvedAnimation`
+- **CurvedAnimationDemo** Demonstrates how to use different curves in `CurvedAnimation`.
 
 ## Other Resources
 
--- a/animations/lib/main.dart
+++ b/animations/lib/main.dart
@@ -21,6 +21,7 @@ import 'src/misc/expand_card.dart';
 import 'src/misc/focus_image.dart';
 import 'src/misc/physics_card_drag.dart';
 import 'src/misc/repeating_animation.dart';
+import 'src/misc/curved_animation.dart';
 
 void main() => runApp(AnimationSamples());
 
@@ -108,6 +109,10 @@ final miscDemos = [
       name: 'Hero Animation',
       route: HeroAnimationDemo.routeName,
       builder: (context) => HeroAnimationDemo()),
+  Demo(
+      name: 'Curved Animation',
+      route: CurvedAnimationDemo.routeName,
+      builder: (context) => CurvedAnimationDemo()),
 ];
 
 final basicDemoRoutes =
--- a/animations/lib/main.dart
+++ b/animations/lib/main.dart
@@ -17,11 +17,11 @@ import 'src/misc/hero_animation.dart';
 import 'src/misc/animated_list.dart';
 import 'src/misc/card_swipe.dart';
 import 'src/misc/carousel.dart';
+import 'src/misc/curved_animation.dart';
 import 'src/misc/expand_card.dart';
 import 'src/misc/focus_image.dart';
 import 'src/misc/physics_card_drag.dart';
 import 'src/misc/repeating_animation.dart';
-import 'src/misc/curved_animation.dart';
 
 void main() => runApp(AnimationSamples());
 
--- a/animations/lib/src/misc/curved_animation.dart
+++ b/animations/lib/src/misc/curved_animation.dart
@@ -86,15 +86,14 @@ class _CurvedAnimationDemoState extends State<CurvedAnimationDemo>
           ),
           SizedBox(height: 25.0),
           RaisedButton(
-            onPressed: () {
-              curvedAnimation.curve =
-                  forwardCurves[random.nextInt(forwardCurves.length)];
-              curvedAnimation.reverseCurve =
-                  reverseCurves[random.nextInt(reverseCurves.length)];
-              animController.forward();
-            },
-            child: Text('Animate'),
-          ),
+              onPressed: () {
+                curvedAnimation.curve =
+                    forwardCurves[random.nextInt(forwardCurves.length)];
+                curvedAnimation.reverseCurve =
+                    reverseCurves[random.nextInt(reverseCurves.length)];
+                animController.forward();
+              },
+              child: Text('Animate'))
         ],
       ),
     );
--- a/animations/lib/src/misc/curved_animation.dart
+++ b/animations/lib/src/misc/curved_animation.dart
@@ -93,7 +93,10 @@ class _CurvedAnimationDemoState extends State<CurvedAnimationDemo>
                     reverseCurves[random.nextInt(reverseCurves.length)];
                 animController.forward();
               },
-              child: Text('Animate'))
+              child: Text(
+                'Animate'
+            ),
+          ),
         ],
       ),
     );
--- a/animations/lib/src/misc/curved_animation.dart
+++ b/animations/lib/src/misc/curved_animation.dart
@@ -26,11 +26,17 @@ class _CurvedAnimationDemoState extends State<CurvedAnimationDemo>
   static const _duration = Duration(seconds: 4);
   List<CurveChoice> curves = [
     CurveChoice(curve: Curves.bounceIn, name: 'Bounce In'),
+    CurveChoice(curve: Curves.bounceOut, name: 'Bounce Out'),
     CurveChoice(curve: Curves.easeInCubic, name: 'Ease In Cubic'),
+    CurveChoice(curve: Curves.easeOutCubic, name: 'Ease Out Cubic'),
     CurveChoice(curve: Curves.easeInExpo, name: 'Ease In Expo'),
+    CurveChoice(curve: Curves.easeOutExpo, name: 'Ease Out Expo'),
     CurveChoice(curve: Curves.elasticIn, name: 'Elastic In'),
+    CurveChoice(curve: Curves.elasticOut, name: 'Elastic Out'),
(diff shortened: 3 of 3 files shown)


animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,107 @@
...
+          SizedBox(height: 25.0),
+          RaisedButton(
+              onPressed: () {
+                curvedAnimation.curve =
+                    forwardCurves[random.nextInt(forwardCurves.length)];
+                curvedAnimation.reverseCurve =
+                    reverseCurves[random.nextInt(reverseCurves.length)];
+                animController.forward();
+              },
+              child: Text('Animate'))
```
jddeep: Put a comma after the `Text` widget and run `flutter format` on the file. Formats and aligns the code better vertically that way.

aksayushx: Sure

domesticmouse: I'm not sure that change was required, this child `Text` element is quite small, and having it inline reduces the amount of vertical space consumed by the build method.

domesticmouse:
Hey @aksayushx, while I appreciate that you are putting together a sample for our repository, there is an ongoing support cost for each and every sample. I'm curious what you are attempting to communicate with this sample, and how you intend to support it going forward as we change Flutter's API?

animations/lib/src/misc/curved_animation.dart
jddeep: @domesticmouse Agreed! Actually, in the past, I had seen cases where we needed to put commas after every single element and follow `dartfmt` over that. Hence, I suggested the change.
But I also agree with on the fact of reducing the vertical spacing of the build method by keeping single widget elements inline.
@aksayushx You can revert your last commit in that case. Thanks.

aksayushx:
Hey @domesticmouse ,
Through my sample I intend any developer going through it to :
1. Understand what 'curve' means in this context
2. Know how to combine it with other classes it is used with.

I also want to assure you that I will remain associated with flutter and try and upgrade the samples with every major change in Flutter's API. 
Thanks

animations/lib/src/misc/curved_animation.dart
aksayushx: Updated acc to latest reviews

RedBrogdon:
This is a great first pass, @aksayushx. Thanks for sending it in.

 Let me pose a couple questions about your learning goals and the design and we can talk about them.

**Understand what 'curve' means in this context**

In general, more people read through the code via GitHub.com than actually clone and run it, so many people will gain knowledge that way. But let's stop for a moment and consider someone who sees the app running. How much information are they getting about the curves that are used?

They would see the animation move, sure, but would they know which curve has been selected while the animation is being run? If not, how might that information be conveyed in the demo's UI?

Also, consider the choice of a rotating Flutter logo. Is it possible that a different animation might better convey how curves work? Or even more than one animation, presented in a row or column? The Flutter documentation for curves shows four(!) in its demos in the API docs:

https://api.flutter.dev/flutter/animation/Curves-class.html

**Know how to combine it with other classes it is used with.**

For this learning goal, I think your sample is doing a good job. There is nothing extraneous being used in the animation, and showing how Curves, AnimationController, CurveAnimation, and Tween all work together is definitely a good thing. You might consider using an AnimatedBuilder around the widgets being animated, though, since it would build on techniques shown in the "basics" section.

It's common for us to revise things a few times before we settle on a design for a sample, so take a moment to consider these questions and then let me know what you're thinking.

Again, this is a good start!

aksayushx:
Thanks @RedBrogdon for the review.
I completely agree with you on the point that a user who just sees the animation in action might not know about the curves being used. So to serve my first learning outcome, I am willing to add two Dropdown lists at the top which would enable the users to choose a curve for the forward rotation and one for reverse. 
I think that the rotating animation along with the dropdown lists would serve my purpose and adding more widgets might bring redundancy to the sample. Please let me know your views on this so that I can make amendments.

aksayushx:
Please review @RedBrogdon @domesticmouse 
I have added dropdown buttons for users to select curves and see them in action.

animations/README.md
```diff
@@ -41,6 +41,7 @@ spring simulation.
 detection to drive an animation.
 - **AnimatedList** Demonstrates how to use `AnimatedList`
 - **HeroAnimationDemo** Demonstrates how to use `Hero` animation.
```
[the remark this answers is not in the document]
aksayushx: Sure

aksayushx:
@RedBrogdon Can you please review the changes I have made.

RedBrogdon:
Thanks for the additional changes, @aksayushx! Letting the user select the curves will definitely help them understand what's going on.

Did you give any additional thought to this question from the previous comment?

> Also, consider the choice of a rotating Flutter logo. Is it possible that a different animation might better convey how curves work? Or even more than one animation, presented in a row or column? The Flutter documentation for curves shows four(!) in its demos in the API docs:

> https://api.flutter.dev/flutter/animation/Curves-class.html

aksayushx:
Thanks for the review @RedBrogdon 
I think maybe adding a translating widget below the rotating widget would better demonstrate the use of curves. Should I also add a widget that translates back and forth based on the curves chosen?

aksayushx:
Hey @RedBrogdon 
I have come up with this and will add it in my upcoming commit after getting your views on it.
CurvedAnimationDemo

animations/README.md
```diff
@@ -41,6 +43,7 @@ spring simulation.
...
+- **CurvedAnimationDemo** Demonstrates how to use different curves in `CurvedAnimation`
```
RedBrogdon, line 46: nit: Need a period at the end of the sentence.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
+// Copyright 2019 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+import 'dart:math' as math;
+
+class CurvedAnimationDemo extends StatefulWidget {
+  static const String routeName = '/misc/curved_animation';
+  @override
```
RedBrogdon, line 10: nit: Add a line of whitespace between the field and the method.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+  _CurvedAnimationDemoState createState() => _CurvedAnimationDemoState();
+}
+
+class MyCurve {
```
RedBrogdon, line 14: Rather than `MyCurve`, consider `CurveChoice` or `CurveSetting` as a name, since `MyCurve` is a fairly generic name.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+  Curve curve;
+  String curveName;
```
RedBrogdon, line 16: `name` would likely be enough here.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+  MyCurve({this.curve, this.curveName});
```
RedBrogdon, line 17: Could this be a `const` contructor?

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+}
+
+class _CurvedAnimationDemoState extends State<CurvedAnimationDemo>
+    with SingleTickerProviderStateMixin {
+  static const String routeName = '/misc/curved_animation';
```
RedBrogdon, line 22: Does the route name need to be here as well? It's included above in a separate class.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+  AnimationController animController;
```
RedBrogdon, line 23: Wherever possible in the samples repo, we try to avoid abbreviations as a way to make code more readable. Consider renaming this just to `controller` (or `animationController` if you prefer it that way).

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+  Animation<double> animation;
+  static const _duration = Duration(seconds: 4);
+  TextStyle style = TextStyle(fontSize: 18.0);
+  List<MyCurve> forwardCurves = [
```
RedBrogdon, line 27: Out of curiosity, why are there two different lists of curves here? All of them would be valid for use as either the forward or reverse curve as far as the CurvedAnimation widget is concerned.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+    MyCurve(curve: Curves.bounceOut, curveName: 'Bounce Out'),
+    MyCurve(curve: Curves.easeOutCubic, curveName: 'Ease Out Cubic'),
+    MyCurve(curve: Curves.easeOutExpo, curveName: 'Ease Out Expo'),
+    MyCurve(curve: Curves.elasticOut, curveName: 'Elastic Out'),
+    MyCurve(curve: Curves.easeOutQuart, curveName: 'Ease Out Quart'),
+    MyCurve(curve: Curves.easeOutCirc, curveName: 'Ease Out Circ')
+  ];
+  MyCurve selectedForwardCurve, selectedReverseCurve;
+  CurvedAnimation curvedAnimation;
+  @override
```
RedBrogdon, line 45: nit: Add a line of whitespace here to separate the curvedAnimation property from the initState method. As you can see in the other code in the repo, it's typical to see a line of whitespace between fields of any kind.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+  MyCurve({this.curve, this.curveName});
+}
+
+class _CurvedAnimationDemoState extends State<CurvedAnimationDemo>
+    with SingleTickerProviderStateMixin {
+  static const String routeName = '/misc/curved_animation';
+  AnimationController animController;
+  Animation<double> animation;
+  static const _duration = Duration(seconds: 4);
+  TextStyle style = TextStyle(fontSize: 18.0);
```
RedBrogdon, line 26: Rather than defining a new text style here, just use one of the available text styles from `Theme.of(context).textTheme` when creating your Text widgets.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,146 @@
...
+            },
+            value: selectedForwardCurve,
+          ),
+          SizedBox(height: 15.0),
+          Text(
+            'Select Curve for reverse rotation',
+            style: style,
+          ),
+          DropdownButton<MyCurve>(
+            items: reverseCurves.map((dropDownItem) {
```
RedBrogdon, line 103: nit: The variable name `dropDownItem` implies that the value it contains is already a drop down item. Consider using `curve` instead.

aksayushx:
@RedBrogdon I have made all changes you have suggested above and also added the translating widget.

animations/lib/main.dart
```diff
@@ -21,6 +21,7 @@ import 'src/misc/expand_card.dart';
 import 'src/misc/focus_image.dart';
 import 'src/misc/physics_card_drag.dart';
 import 'src/misc/repeating_animation.dart';
+import 'src/misc/curved_animation.dart';
```
jddeep, line 24: @aksayushx Can you order the import alphabetically? This is why your Travis build is failing.

aksayushx: Cool. I will do that

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,156 @@
...
+  const CurveChoice({this.curve, this.name});
+}
+
+class _CurvedAnimationDemoState extends State<CurvedAnimationDemo>
+    with SingleTickerProviderStateMixin {
+  AnimationController controller;
+  Animation<double> animationRotation;
+  Animation<Offset> animationTranslation;
+  static const _duration = Duration(seconds: 4);
+  List<CurveChoice> curves = [
```
RedBrogdon, line 27: Ah! No, I meant you could combine all the choices in both lists to make one bigger one! The `SomethingSomethingOut` curves can also be included here. 😄

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,156 @@
...
+        if (status == AnimationStatus.completed) {
+          controller.reverse();
+        }
+      });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(title: Text('Curved Animation')),
```
RedBrogdon, line 81: nit: Add a comma in between the two parentheses at the end of this line:

```
appBar: AppBar(title: Text('Curved Animation'),),
```

While it's not completely necessary, it'll space the parent/child out the same way it's done in most of our other samples.

animations/lib/src/misc/curved_animation.dart
```diff
@@ -0,0 +1,156 @@
...
+                size: 100,
+              ),
+            ),
+          ),
+          SizedBox(height: 25.0),
+          RaisedButton(
+              onPressed: () {
+                controller.forward();
+              },
+              child: Text('Animate')),
```
RedBrogdon, line 145: Add a comma in between the two parentheses on the end of this line as well.

aksayushx: Although I am inserting a comma between the parenthesis ,I would like to quote @domesticmouse's reply from an earlier conversation  
> I'm not sure that change was required, this child `Text` element is quite small, and having it inline reduces the amount of vertical space consumed by the build method.

aksayushx:
@RedBrogdon Thanks for your review and wonderful guidance all this way which helped me in making something that's close to be merged. 
I have made the suggested changes. Thanks ;)

PR #395, opened 2020-03-24
[VeggieSeasons] Fix dark mode problems

abd99:
Fixes problems in #322
I've changed the declared const Styles to a method which were causing the issues.

Please take a look @RedBrogdon @johnpryan.

Screencast:

I've made sure that nothing breaks in the light mode as well

### DIFF
--- a/veggieseasons/lib/screens/details.dart
+++ b/veggieseasons/lib/screens/details.dart
@@ -28,11 +28,12 @@ class ServingInfoChart extends StatelessWidget {
       builder: (context, snapshot) {
         final target = snapshot?.data ?? 2000;
         final percent = standardPercentage * 2000 ~/ target;
+        final CupertinoThemeData themeData = CupertinoTheme.of(context);
 
         return Text(
           '$percent% DV',
           textAlign: TextAlign.end,
-          style: Styles.detailsServingValueText(context),
+          style: Styles.detailsServingValueText(themeData),
         );
       },
     );
@@ -40,6 +41,7 @@ class ServingInfoChart extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final CupertinoThemeData themeData = CupertinoTheme.of(context);
     return Column(
       children: [
         SizedBox(height: 16),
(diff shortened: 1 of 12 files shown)


veggieseasons/lib/widgets/settings_item.dart
```diff
@@ -132,20 +132,40 @@ class SettingsItemState extends State<SettingsItem> {
                           crossAxisAlignment: CrossAxisAlignment.start,
                           children: <Widget>[
                             SizedBox(height: 8.5),
-                            Text(widget.label),
+                            Text(
+                              widget.label,
```
RedBrogdon: Consider making text styles in `styles.dart` for these Text widgets as well. They were originally copied into the project from another source, which is why they're set up slightly differently. There's no reason they can't be the same as the others, though.

veggieseasons/lib/screens/details.dart
```diff
@@ -70,14 +70,14 @@ class ServingInfoChart extends StatelessWidget {
                       TableCell(
                         child: Text(
                           'Serving size:',
-                          style: Styles.detailsServingLabelText,
+                          style: Styles.detailsServingLabelText(context),
```
RedBrogdon: There is a small cost to repeatedly using BuildContexts to look up an inherited widget, which is what the methods in styles.dart are doing with the context being passed here. Consider having those methods take a CupertinoThemeData directly, so you can fetch that data once in this class, and use it repeatedly in calls to the styles.

This would also help in a small way with separation of concern, since the styling really just needs the theme data itself, rather than full access to the build context in most cases.

abd99:
I have made the changes as per your suggestion. One problem I faced was that whenever I was trying to check the brightness of the CupertinoThemeData, it was always returning light no matter what the mode of the app was. 
Eg.
```
CupertinoThemeData themeData = CupertinoTheme.of(context);
//themeData.brightness was returning Brightness.light
```
So, I had to directly create a Brightness object from the context and then pass it to methods in some cases where the output was dependent on the brightness and predefined theme wasn't there for that specific case. Any specific reason it was happening that I might be unaware of?

RedBrogdon:
> So, I had to directly create a Brightness object from the context and then pass it to methods in some cases where the output was dependent on the brightness and predefined theme wasn't there for that specific case. Any specific reason it was happening that I might be unaware of?

`brightnessOf` would be the correct method to use, since it will account for CupertinoThemeData objects that have a null brightness (though I don't believe it would matter here). I'm not sure why you'd be getting a light brightness value every time. If I have a chance this week, I'll see if I can replicate it with a smaller project.

abd99:
> If I have a chance this week, I'll see if I can replicate it with a smaller project.
Sure, thanks for the support on this PR.

PR #400, opened 2020-03-27
[provider_shopper] Added widget tests for login, catalog and cart screens

### LINKED ISSUE #387: [provider_shopper] Add Widget Tests
In the `provider_shopper` sample app, we have a widget test already there for demonstration here - https://github.com/flutter/samples/blob/master/provider_shopper/test/widget_test.dart.
I suggest including a driver test for the app as well for demonstration of integration tests. The driver test itself should be pretty straight forward according to the app functioning.

jddeep:
Fixes #387 

Added widget tests for login, catalog and cart screens.

### DIFF
--- a/provider_shopper/lib/models/catalog.dart
+++ b/provider_shopper/lib/models/catalog.dart
@@ -30,12 +30,6 @@ class CatalogModel {
     'Currying',
   ];
 
-  CatalogModel();
-
-  CatalogModel.modifyModel(int lrange) {
-    itemNames.removeRange(lrange, itemNames.length);
-  }
-
   /// Get item by [id].
   ///
   /// In this sample, the catalog is infinite, looping over [itemNames].
--- a/provider_shopper/lib/models/catalog.dart
+++ b/provider_shopper/lib/models/catalog.dart
@@ -32,7 +32,7 @@ class CatalogModel {
 
   CatalogModel();
 
-  CatalogModel.modifyModel(int lrange){
+  CatalogModel.modifyModel(int lrange) {
     itemNames.removeRange(lrange, itemNames.length);
   }
 
--- a/provider_shopper/lib/models/catalog.dart
+++ b/provider_shopper/lib/models/catalog.dart
@@ -12,7 +12,7 @@ import 'package:flutter/material.dart';
 /// For simplicity, the catalog is expected to be immutable (no products are
 /// expected to be added, removed or changed during the execution of the app).
 class CatalogModel {
-  static const _itemNames = [
+  static const itemNames = [
     'Code Smell',
     'Control Flow',
     'Interpreter',
@@ -32,8 +32,8 @@ class CatalogModel {
 
   /// Get item by [id].
   ///
-  /// In this sample, the catalog is infinite, looping over [_itemNames].
-  Item getById(int id) => Item(id, _itemNames[id % _itemNames.length]);
+  /// In this sample, the catalog is infinite, looping over [itemNames].
+  Item getById(int id) => Item(id, itemNames[id % itemNames.length]);
 
   /// Get item by its position in the catalog.
   Item getByPosition(int position) {
--- a/provider_shopper/lib/models/catalog.dart
+++ b/provider_shopper/lib/models/catalog.dart
@@ -12,7 +12,7 @@ import 'package:flutter/material.dart';
 /// For simplicity, the catalog is expected to be immutable (no products are
 /// expected to be added, removed or changed during the execution of the app).
 class CatalogModel {
-  static const itemNames = [
+  static List<String> itemNames = [
     'Code Smell',
     'Control Flow',
     'Interpreter',
@@ -29,6 +29,12 @@ class CatalogModel {
     'Bit Shift',
     'Currying',
   ];
+
+  CatalogModel();
+
+  CatalogModel.modifyModel(int lrange){
+    itemNames.removeRange(lrange, itemNames.length);
+  }
 
   /// Get item by [id].
   ///
--- a/provider_shopper/test/cart_widget_test.dart
+++ b/provider_shopper/test/cart_widget_test.dart
@@ -1,0 +1,72 @@
+// Copyright 2019 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:provider/provider.dart';
+import 'package:provider_shopper/models/cart.dart';
+import 'package:provider_shopper/models/catalog.dart';
+import 'package:provider_shopper/screens/cart.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+  List<SingleChildCloneableWidget> _providers = [
+    Provider(create: (context) => CatalogModel()),
+    ChangeNotifierProxyProvider<CatalogModel, CartModel>(
+      create: (context) => CartModel(),
+      update: (context, catalog, cart) {
+        cart.catalog = catalog;
+        return cart;
+      },
+    ),
+  ];
+  group('Cart Screen basic widget testing', () {
+    testWidgets('Testing the layout of the widgets', (tester) async {
+      await tester.pumpWidget(MultiProvider(
+        providers: _providers,
+        child: MaterialApp(
+          home: MyCart(),
+        ),
+      ));
+
+      // Testing the layout of the screen widgets
+      expect(find.text('Cart'), findsOneWidget);
+      expect(find.byType(Divider), findsOneWidget);
+      expect(
+          find.ancestor(
+              of: find.byType(Divider), matching: find.byType(Column)),
+          findsOneWidget);
+      WidgetPredicate cartTotalPredicate =
+          (widget) => widget is SizedBox && widget.height == 200;
+      expect(find.byWidgetPredicate(cartTotalPredicate), findsOneWidget);
+      expect(find.text('\$0'), findsOneWidget);
+      expect(find.text('BUY'), findsOneWidget);
+    });
+  });
+
+  group('Widget Testing for the BUY button click and Snackbar', () {
+    testWidgets('Testing the button tap and snackbar', (tester) async {
+      final snackbarMessage = 'Buying not supported yet.';
+
+      await tester.pumpWidget(MultiProvider(
+        providers: _providers,
+        child: MaterialApp(
+          home: MyCart(),
+        ),
+      ));
+
+      expect(
+          find.ancestor(
+              of: find.text('BUY'), matching: find.byType(FlatButton)),
+          findsOneWidget);
+      // verify no snacknar initially
+      expect(find.text(snackbarMessage), findsNothing);
+      await tester.tap(find.text('BUY'));
+      // schedule animation
+      await tester.pump();
+      // verifying the snackbar upon clicking the button
+      expect(find.text(snackbarMessage), findsOneWidget);
+    });
+  });
+}
--- a/provider_shopper/test/cart_widget_test.dart
+++ b/provider_shopper/test/cart_widget_test.dart
@@ -53,7 +53,7 @@ void main() {
         await tester.pumpAndSettle();
         expect(find.text(item.name), findsOneWidget);
       }
-      
+
       // Testing total price of the five items.
       expect(find.text('\$${42 * 5}'), findsOneWidget);
       expect(find.byIcon(Icons.done), findsNWidgets(5));
--- a/provider_shopper/test/cart_widget_test.dart
+++ b/provider_shopper/test/cart_widget_test.dart
@@ -11,24 +11,24 @@ import 'package:provider_shopper/screens/cart.dart';
 
 void main() {
   TestWidgetsFlutterBinding.ensureInitialized();
-  List<SingleChildCloneableWidget> _providers = [
-    Provider(create: (context) => CatalogModel()),
-    ChangeNotifierProxyProvider<CatalogModel, CartModel>(
-      create: (context) => CartModel(),
-      update: (context, catalog, cart) {
-        cart.catalog = catalog;
-        return cart;
-      },
-    ),
-  ];
-  group('Cart Screen basic widget testing', () {
-    testWidgets('Testing the layout of the widgets', (tester) async {
-      await tester.pumpWidget(MultiProvider(
-        providers: _providers,
+  Widget cartScreen() => MultiProvider(
+        providers: [
+          Provider(create: (context) => CatalogModel()),
+          ChangeNotifierProxyProvider<CatalogModel, CartModel>(
+            create: (context) => CartModel(),
+            update: (context, catalog, cart) {
+              cart.catalog = catalog;
+              return cart;
+            },
+          ),
+        ],
         child: MaterialApp(
           home: MyCart(),
         ),
-      ));
+      );
+  group('Cart Screen basic widget testing', () {
+    testWidgets('Testing the layout of the widgets', (tester) async {
+      await tester.pumpWidget(cartScreen());
 
       // Testing the layout of the screen widgets
       expect(find.text('Cart'), findsOneWidget);
@@ -49,12 +49,7 @@ void main() {
     testWidgets('Testing the button tap and snackbar', (tester) async {
       final snackbarMessage = 'Buying not supported yet.';
 
-      await tester.pumpWidget(MultiProvider(
-        providers: _providers,
-        child: MaterialApp(
-          home: MyCart(),
-        ),
-      ));
+      await tester.pumpWidget(cartScreen());
 
       expect(
           find.ancestor(
--- a/provider_shopper/test/cart_widget_test.dart
+++ b/provider_shopper/test/cart_widget_test.dart
@@ -31,36 +31,30 @@ Widget createCartScreen() => MultiProvider(
 
 void main() {
   group('CartScreen widget tests', () {
-    testWidgets('Testing when the cart is empty', (tester) async {
+    testWidgets('Tapping BUY button displays snackbar.', (tester) async {
       await tester.pumpWidget(createCartScreen());
 
-      // Testing the layout of the screen widgets
-      expect(find.text('Cart'), findsOneWidget);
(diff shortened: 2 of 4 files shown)


jddeep:
@RedBrogdon Reviews here please. Thanks.

provider_shopper/test/catalog_widget_test.dart
```diff
@@ -0,0 +1,72 @@
...
+    await tester.drag(find.byType(CustomScrollView), const Offset(0.0, 0.0));
+    await tester.pumpAndSettle();
+
+    // Should find 11 buttons at the current/base viewport height.
+    expect(find.text('ADD'), findsNWidgets(11));
+
+    await tester.tap(find.widgetWithText(FlatButton, 'ADD').first);
+    await tester.pumpAndSettle();
+
+    // expect(find.byIcon(Icons.check), findsOneWidget);
```
jddeep: @RedBrogdon Also wanted to confirm something with you. This statement should successfully pass right given the previous actions. But surprisingly this was failing :thinking: Could you give me any insight?

provider_shopper/test/catalog_widget_test.dart
```diff
@@ -0,0 +1,72 @@
+// Copyright 2019 The Flutter team. All rights reserved.
```
AyushBherwani1998: Hey @jddeep, the year should be 2020.

AyushBherwani1998: Hey @jddeep , the year should be 2020.

provider_shopper/test/catalog_widget_test.dart
AyushBherwani1998: Hey @jddeep, I guess the issue is with the scrolling which you are doing. I have a few doubts when performing drag you are giving the `Offest(0, -380)`. How did you calculate that?

If I give the `Offset(0, -350)`, and instead of finding exactly 11 widgets, if I look for N widgets, the test is passing. The funny thing is if I give `Offset(0, -360)` the test will fail. And if I tap on the last instead of first `FlatButton` with ADD the test is passing.

Here's a proposed solution to avoid that, if you want to check if the scrolling is working properly you can do like this.

``` dart

// Perform Scroll Up
    await tester.drag(find.byType(CustomScrollView), const Offset(0.0, -360.0));
    await tester.pumpAndSettle();
    
    // Once scroll up is performed, we should not be able to find Catalog
    expect(find.text('Catalog'),findsNothing);
    
    // Perform Scroll down
    await tester.drag(find.byType(CustomScrollView), const Offset(0.0, 360.0));
    await tester.pumpAndSettle();
    
    // Once scroll down is performed, we should be able to find Catalog
    expect(find.text('Catalog'),findsOneWidget);
    
    // Should find multiple widget
    expect(find.text('ADD'), findsWidgets);
    await tester.tap(find.text('ADD').first);
    await tester.pumpAndSettle();

    expect(find.byIcon(Icons.check), findsOneWidget);
```

jddeep: @AyushBherwani1998 actually, I got that offset value according to the viewport height and also by dry running the test and evaluating my desired/expected results. Everything else is working as expected cause I also wanted to test the list items(the item names) and how they are laid out. Only testing the scrolling was not my actual and sole aim.
After tapping the `first` - `ADD` button, there should be `one` - `check` icon by means of simple logic. But that particular expectation was not being satisfied. That's what got me thinking :thinking:

AyushBherwani1998: Exactly, the logic should work. I did try the same locally but not able to exactly identify the issue.

jddeep: Yeah cc: @RedBrogdon

provider_shopper/lib/models/catalog.dart
```diff
@@ -41,6 +41,8 @@ class CatalogModel {
     // is also its id.
     return getById(position);
   }
+
+  static List<String> getCatalogItems() => _itemNames;
```
RedBrogdon: Rather than add a property here, you could simply "unhide" _itemNames by removing the underscore. It's a constant, so it can't be modified anyway.

jddeep:
@RedBrogdon Have made some updates here. Have a look now. Agreed on all your points! I did not commit the cart test specifically of the fact that I was pumping `MyApp` every time and wanted to clear that out since we needed to pass the provider and cartmodel. Also, I had kept the login screen test logic simple and direct in one test function without any group.
Do let me know your further thoughts and if something else needs updates. Thanks :)

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,72 @@
...
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:provider/provider.dart';
+import 'package:provider_shopper/models/cart.dart';
+import 'package:provider_shopper/models/catalog.dart';
+import 'package:provider_shopper/screens/cart.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+  List<SingleChildCloneableWidget> _providers = [
```
AyushBherwani1998, line 14: Instead of `SingleChildCloneableWidget` use `SingleChildWidget`. The `SingleChildCloneableWidget` is removed in the recent version of `Provider`. You can refere the readme.

jddeep: Cool. Understandable changes. Will update.

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,67 @@
+// Copyright 2019 The Flutter team. All rights reserved.
```
RedBrogdon, line 1: Nit: 2020, since this file is entirely new.

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,67 @@
...
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:provider/provider.dart';
+import 'package:provider_shopper/models/cart.dart';
+import 'package:provider_shopper/models/catalog.dart';
+import 'package:provider_shopper/screens/cart.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+  Widget cartScreen() => MultiProvider(
```
RedBrogdon, line 14: Typically we don't created nested functions like this, so it'd be better to move this into global scope in the file. I'd recommend renaming it to "createCartScreen", as well.

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,67 @@
...
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:provider/provider.dart';
+import 'package:provider_shopper/models/cart.dart';
+import 'package:provider_shopper/models/catalog.dart';
+import 'package:provider_shopper/screens/cart.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
```
RedBrogdon, line 13: This call shouldn't be necessary. If you're seeing a failure without it, that's something we'd want to know about.

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,67 @@
...
+              cart.catalog = catalog;
+              return cart;
+            },
+          ),
+        ],
+        child: MaterialApp(
+          home: MyCart(),
+        ),
+      );
+  group('Cart Screen basic widget testing', () {
```
RedBrogdon, line 29: It looks like this file has a group for each test, which isn't typically how test groups are used. I would suggest having a single group for all the tests in the file, called "CartScreen widget tests" or something similar.

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,67 @@
...
+          (widget) => widget is SizedBox && widget.height == 200;
+      expect(find.byWidgetPredicate(cartTotalPredicate), findsOneWidget);
+      expect(find.text('\$0'), findsOneWidget);
+      expect(find.text('BUY'), findsOneWidget);
+    });
+  });
+
+  group('Widget Testing for the BUY button click and Snackbar', () {
+    testWidgets('Testing the button tap and snackbar', (tester) async {
+      final snackbarMessage = 'Buying not supported yet.';
```
RedBrogdon, line 50: I would probably avoid using the actual text of the snackbar (which strongly couples the test to that detail) and instead check for the widget by type (wince what you're most interested in testing is whether hitting the button produces a snackbar at all).

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,67 @@
...
+              return cart;
+            },
+          ),
+        ],
+        child: MaterialApp(
+          home: MyCart(),
+        ),
+      );
+  group('Cart Screen basic widget testing', () {
+    testWidgets('Testing the layout of the widgets', (tester) async {
```
RedBrogdon, line 30: Testing the entire basic layout of the screen is likely unnecessary. Let's be honest -- it would be *super* weird if that behavior were to suddenly change. 😄 

What you could test, though, is the functionality that makes CartScreen *interesting*. You're doing that below when you check to see if the BUY button causes the SnackBar to appear. Most screens don't have a button with that label that cause a snackbar to appear, but CartScreen does, so you test *that*. 

Similarly, what you could do here is create two tests, one with an empty cart and one with a full one, and just make sure the text for the total cost and number of rows are correct. *That's* work that CartScreen is doing that's specific to it.

jddeep:
@RedBrogdon, Have made updates :)

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,68 @@
...
+      ),
+    );
+
+void main() {
+  group('CartScreen widget tests', () {
+    testWidgets('Testing when the cart is empty', (tester) async {
+      await tester.pumpWidget(createCartScreen());
+
+      // Testing the layout of the screen widgets
+      expect(find.text('Cart'), findsOneWidget);
```
RedBrogdon, line 38: The strength of this test is in lines 45-50, which are a great check that the behavior of the BUY button works as designed.

In order to keep the focus of this test as narrow as possible, I would remove these three expect statements and change the name of the test to "Tapping BUY button displays snackbar."

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,68 @@
...
+      expect(find.text('\$0'), findsOneWidget);
+      expect(
+          find.ancestor(
+              of: find.text('BUY'), matching: find.byType(FlatButton)),
+          findsOneWidget);
+      // verify no snacknar initially
```
RedBrogdon, line 44: nit: "Verify no snackbar initially exists."

Within the repo, comments start with a capital letter and end with a period, so it'd be good to make sure all comments added in this PR do the same. It's a grammar annoyance, I admit, but those little details do matter. :)

provider_shopper/test/cart_widget_test.dart
```diff
@@ -0,0 +1,68 @@
...
+      await tester.pumpWidget(createCartScreen());
+
+      // Adding five items in the cart and testing
+      for (int i = 0; i < 5; i++) {
+        var item = catalogModel.getByPosition(i);
+        cartModel.add(item);
+        await tester.pumpAndSettle();
+        expect(find.text(item.name), findsOneWidget);
+      }
+      // Testing total price of the five items
```
RedBrogdon, line 63: nit: Add a line of whitespace before this line.

provider_shopper/test/catalog_widget_test.dart
```diff
@@ -0,0 +1,101 @@
+// Copyright 2020 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/widgets.dart';
```
RedBrogdon, line 5: These imports need to be sorted.

provider_shopper/test/catalog_widget_test.dart
```diff
@@ -0,0 +1,101 @@
...
+
+void main() {
+  final catalogListItems = CatalogModel.itemNames;
+
+  group('CatalogScreen Widget Tests', () {
+    testWidgets('Testing the layout and presentation of widgets',
+        (tester) async {
+      await tester.pumpWidget(createCatalogScreen());
+
+      // Verifying the layout
```
RedBrogdon, line 37: These two expect statements aren't all that interesting (it would almost be a failure of the Flutter SDK rather than the app if they failed). I would recommend cutting them and focusing the test on making sure the catalog screen displays the right number of rows, which you're doing below.

provider_shopper/test/catalog_widget_test.dart
```diff
@@ -0,0 +1,101 @@
...
+      expect(find.byIcon(Icons.shopping_cart), findsOneWidget);
+
+      // Testing for the first N(9) items which appear on the screen at first
+      // according to the base viewport
+      for (String item in catalogListItems.sublist(0, 9)) {
+        expect(find.text(item), findsOneWidget);
+      }
+    });
+
+    testWidgets('Testing the scroll behaviour', (tester) async {
```
RedBrogdon, line 48: While I admit that it's cool to see scrolling done in one of the widget tests, the behavior that's actually being tested here (and in the next test) is that of the SliverList and SliverAppBar rather that the code that's written specifically for provider_shopper. It just doesn't make sense for the app, and both of these tests should be removed.

provider_shopper/test/catalog_widget_test.dart
```diff
@@ -0,0 +1,101 @@
...
+            findsOneWidget);
+      }
+    });
+
+    testWidgets('Testing the ADD buttons and check after clicking',
+        (tester) async {
+      await tester.pumpWidget(createCatalogScreen());
+
+      // Should find 9 buttons(since 9 items) at the current/base viewport height.
+      expect(find.text('ADD'), findsNWidgets(9));
```
RedBrogdon, line 91: I am concerned that the use of 9 in this and some of the other tests is tightly coupling the test to the particular size of viewport used during the test. If we were to use these tests with Firebase Test Lab, for example, on more than one device, it's likely that they'd start to fail. Consider doing something like:

Changing the CatalogModel class so that you can create it with the full catalog or a smaller amount of items (three, maybe), that you could guarantee would always be visible for any devices screen size. You could then use the smaller version of the CatalogModel in your tests, and be sure that viewport size wouldn't matter.

provider_shopper/test/catalog_widget_test.dart
```diff
@@ -0,0 +1,101 @@
...
+
+      // Performing the click on the ADD button of the first item in the list
```
RedBrogdon, line 93: Lines 93-98 here are really great. You're testing one specific thing (clicking add causes the checkmark to appear) and leaving everything else out.

provider_shopper/test/login_widget_test.dart
```diff
@@ -0,0 +1,33 @@
...
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:provider_shopper/screens/login.dart';
+
+void main() {
+  testWidgets('Login page Widget test', (tester) async {
+    await tester.pumpWidget(MaterialApp(home: MyLogin()));
+
+    // Testing the layout and positioning of the screen widgets
```
RedBrogdon, line 13: This test is mostly checking that the build method is creating the basic layout correctly, which is not all that interesting of a thing to test. I can't imagine this test ever failing, for instance.

What would be a more effective test would be a check that the ENTER button does what it's supposed to do (navigate to a new route). That behavior is specific to the LoginScreen, and should definitely be tested. You could convert this test to do that by cutting everything above line 30, adding a tap on the ENTER button, doing a pumpAndSettle, and then looking for the CatalogScreen.

RedBrogdon:
> @RedBrogdon, Have made updates :) Let me know if something else needs a update.

Thanks for the additional changes! I've left some more comments. As you've probably seen in some of the other PRs, it can take a lot of review, especially for a relatively new contributor, to land a change this big. It's just part of the normal process, though.

jddeep:
> Thanks for the additional changes! I've left some more comments. As you've probably seen in some of the other PRs, it can take a lot of reviews, especially for a relatively new contributor, to land a change this big. It's just part of the normal process, though.

@RedBrogdon Thanks a lot for your support on the PRs!! Being a fair consistent contributor now in this community has helped me to understand the system and workflow of our different projects. 
Moreover, this exact process is something which I enjoy more rather than the fact of just getting something merged.
Let me know your further thoughts here. Again, Thanks for your useful reviews on the PRs :smile:

provider_shopper/test/catalog_widget_test.dart
```diff
@@ -0,0 +1,58 @@
...
+      child: MaterialApp(
+        home: MyCatalog(),
+      ),
+    );
+
+void main() {
+  final catalogListItems = CatalogModel.itemNames;
+
+  group('CatalogScreen Widget Tests', () {
+    testWidgets('Testing the layout and presentation of widgets',
```
RedBrogdon, line 33: Now that you've done a good job of narrowing the scope of this test, you could make this name more specific. Something like "Testing item row counts and text".

provider_shopper/test/login_widget_test.dart
```diff
@@ -0,0 +1,44 @@
...
+        initialRoute: '/',
+        routes: {
+          '/': (context) => MyLogin(),
+          '/catalog': (context) => MyCatalog(),
+        },
+      ),
+    ));
+
+    // Verifying the behaviour of ENTER button.
+    final enterBtnFinder = find.widgetWithText(RaisedButton, 'ENTER');
```
RedBrogdon, line 36: These two lines can be removed. If there's no widget with 'ENTER' as its text, the finder on line 39 will fail, and there's no way there will ever be two or more. 😄

RedBrogdon:
Thanks!

jddeep:
Thanks @RedBrogdon. Now that i have understood more clearly the way we write tests for a whole app in our repo here, i feel confident going about writing tests further for different apps in our repo :) Moreover, i had also submitted a proposal for the Testing sample app in our repo during the GSoC period. Could you please have a look at it and give me your thoughts on that? Wanted to get our thoughts aligned and finalised before coding starts. Thanks 😊

PR #408, opened 2020-04-04
a little change to make the code more readable

samrat19:
In the previous version 

it was
void main(){
  runApp(MyApp());
}

but 
using fat arrow the code looks good and more readable

main() => runApp(MyApp())

### DIFF
--- a/provider_shopper/lib/main.dart
+++ b/provider_shopper/lib/main.dart
@@ -11,9 +11,7 @@ import 'package:provider_shopper/screens/cart.dart';
 import 'package:provider_shopper/screens/catalog.dart';
 import 'package:provider_shopper/screens/login.dart';
 
-void main() {
-  runApp(MyApp());
-}
+main() => runApp(MyApp());
 
 class MyApp extends StatelessWidget {
   @override


provider_shopper/lib/main.dart
```diff
@@ -11,9 +11,7 @@ import 'package:provider_shopper/screens/cart.dart';
 import 'package:provider_shopper/screens/catalog.dart';
 import 'package:provider_shopper/screens/login.dart';
 
-void main() {
-  runApp(MyApp());
-}
+main() => runApp(MyApp());
```
aadarsh-patel, line 14: Add return type `void`.

Travis CI build failed due to the same reason.
```
 error • Missing return type for 'main' • lib/main.dart:14:1 •
         implicit_dynamic_return
```

RedBrogdon:
@samrat19, while I appreciate the time it took to send in a PR, this is not a change I would want to land. Dart's arrow syntax is great for a lot of things, but we typically reserve it for specific uses, such as lambdas and methods that quickly return a one-line value. Because the `main` method is so important, I would not want to recommend its use there. 

Again, thanks for the PR!

PR #422, opened 2020-04-27
Update README.md

johnpryan:


experimental/web_dashboard/README.md
```diff
@@ -1,6 +1,17 @@
 # web_dashboard
 
-(In progress) A dashboard app for the web, built with Flutter.
+**In progress** 
+
+A dashboard app that displays daily entries.
+
+1. How to use an AdaptiveScaffold adaptive layout for large, medium, and small
```
RedBrogdon, line 7: Is this the goals section, similar to the ones we have for other apps?

johnpryan: yeah, we should probably make this a proper section in the future.

PR #425, opened 2020-04-30
Moved two animation vignettes into misc.

RedBrogdon:
Moves the vignettes featuring AnimatedPositioned and AnimatedSwitcher from `basics` to `misc`.

The basics section is designed to follow a particular progression of topics, easing the reader into more and more advanced bits of the animations API. For that reason, these two examples are better suited to the miscellaneous section.


AyushBherwani1998:
@RedBrogdon we can also update the README.md and move the examples from Basics to Misc section.

johnpryan:
That's a good point, we need to update the README too.

jddeep:
> @RedBrogdon we can also update the README.md and move the examples from Basics to Misc section.

:+1:

RedBrogdon:
Yep, should have updated the README. I've pushed another commit.

@johnpryan, I saw you approved, but do you mind taking another quick look? I reformatted a couple things and added a codelab link. Just want to make sure it's all good by you.

PR #428, opened 2020-05-05
[Veggie Seasons] Update dependencies on xcode project

goderbauer:
Dependency update removes a deprecation warning during launch and during first run flutter was upgrading the xcode project to the latest version.


domesticmouse:
@RedBrogdon should we do a new TestFlight of VeggieSeasons on 1.17 to try out Metal rendering?

RedBrogdon:
Oooooh, that's a really good idea. I'll try to make time this week.

PR #433, opened 2020-05-07
State Restoration support for Veggie Seasons app

goderbauer:
This PR enables state restoration for the Veggie Seasons app using Flutter's state restoration framework (Design Doc: http://flutter.dev/go/state-restoration-design).

The changes in this PR require a Flutter version after commit https://github.com/flutter/flutter/commit/053ebf2c080c7b8efbf4020683a5ba27d9daa3b8.

State Restoration for Flutter is currently only enabled for Android (adding support for iOS is tracked in https://github.com/flutter/flutter/issues/62915). In order to test the state restoration aspect of this app follow these steps:

1. Change into the veggieseasons directory.
2. Run `flutter create .`. This generates the boilerplate required to run the veggie seasons app on Android.
3. Enable "Don't keep activities" in the developer settings of your Android device.
4. Run the app on the Android device via `flutter run`.
5. Navigate around in the app, change some state.
6. Switch to the home screen or another app via the app switcher. This kills the Veggie Seasons app.
7. Switch back the Veggie Seasons app via the app switcher. This will restart the Veggie Seasons app.
8. Observe that no state is lost.

Part of https://github.com/flutter/flutter/issues/62916.

## Background

When an app is backgrounded on e.g. Android, the OS may kill the app at any time to reclaim resources. When the user switches back to the app, the app is relaunched and expected to restore the same state it had when it was killed. For that, apps can serialize out state information before they are killed and that information is handed back to them when they are relaunched.

## Summary

In this PR, widget `State` objects that want some of their properties restored when the app is relaunched mixin the `RestorationMixin`. They then use (subclasses of) `RestorableProperty` to store information that they need restored. When the app is killed, the current value of the `RestorableProperty` is serialized out and when the app relaunches, the property magically has its old value again. App developer need to serialize out enough information to rebuild the same widget tree that was active when the app was killed.

Those `RestorableProperty`s in a `State` object need to be bound to a restoration id by calling `RestorationMixin.registerForRestoration` in the `restoreState` method. The serialized data of a property is stored under that id and the id is used again to pull the correct data for a given property out of the app's serialized state again when the app is relaunched.

The `restoreState` method added to `State` objects by the `RestorationMixin` is called after `initState`. In this method, `RestorableProperty` should be registered to initialize them with the data from the serialized state. Other initialization code, that depends on the value of a `RestorableProperty`, should also run here. The `restoreState` method will be called again if the app is asked to restore from different serialized state while it is running (this is currently uncommon, but may be relevant for some use cases on the web).

The framework comes with built-in `RestorableProperty` subclasses to store common values like an int, double, String, TextEditingController, etc. Apps can create their own subclasses to serialize and deserialize more complex custom objects.

Widgets, that ship with the framework, will have the functionality to restore state built-in. In order to enable it, app developers need to provide a restoration id to those widgets. The widgets will store their data under this id in the surrounding restoration scope.

Restoration ids need to be unique in their surrounding restoration scope. A new restoration scope can be established by inserting a `RestorationScope` widget into the tree.


veggieseasons/lib/main.dart
```diff
@@ -16,16 +16,74 @@ void main() {
...
   ]);
 
-  runApp(
-    ScopedModel<AppState>(
-      model: AppState(),
+  // The root scope is usually injected automatically by the
+  // CupertinoApp/MaterialApp. However, in this particular app there are widgets
+  // above the CupertinoApp that need their state restored, so we are injecting
+  // the root scope manually above everything.
+  runApp(RestorationScope.root(
```
xster, line 23: this code isn't on master yet is it? Are there non-root constructors?

veggieseasons/lib/main.dart
```diff
@@ -16,16 +16,74 @@ void main() {
...
+  State<VeggieApp> createState() => _VeggieAppState();
+}
+
+class _VeggieAppState extends State<VeggieApp> with RestorationMixin {
+  RestorableAppState appState = RestorableAppState();
+
+  @override
+  void initState() {
+    super.initState();
+    register(appState, const RestorationId('state'));
```
xster, line 44: could this name be more tightly scoped? `register` sounds really broad.

veggieseasons/lib/screens/search.dart
```diff
@@ -12,19 +12,25 @@ import 'package:veggieseasons/widgets/search_bar.dart';
...
+  final RestorationId restorationId;
+
   @override
   _SearchScreenState createState() => _SearchScreenState();
 }
 
-class _SearchScreenState extends State<SearchScreen> {
-  final controller = TextEditingController();
+class _SearchScreenState extends State<SearchScreen> with RestorationMixin {
+  RestorableTextEditingController controller = RestorableTextEditingController(TextEditingController());
```
xster, line 24: oh wow, neat

veggieseasons/lib/screens/settings.dart
```diff
@@ -149,12 +155,7 @@ class SettingsScreen extends StatelessWidget {
         },
       ),
       onPress: () {
-        Navigator.of(context).push<void>(
-          CupertinoPageRoute(
-            builder: (context) => CalorieSettingsScreen(),
-            title: 'Calorie Target',
-          ),
-        );
+        Navigator.of(context).restorablePush<void>(_calorieSettingsScreen);
```
xster, line 158: could the .of happen to be implemented by a RestorableNavigator? Or it's intended for the user to choose whether each navigation change is restorable or not?

veggieseasons/lib/main.dart
goderbauer: The code currently lives in a personal branch of mine. I am in the process of cleaning that up and then will make it available.

There are other constructors. The default one just starts a new child-scope within its parent scope. The root one actually inserts the data received from the engine into the tree.

veggieseasons/lib/main.dart
goderbauer: I am open to suggestions :)

However, it was meant to have a somewhat broader scope because, after this call the RestorationMixin manages the registered proeprty.

veggieseasons/lib/screens/settings.dart
goderbauer: There's no RestorableNavigator class. Just like any other built-in widget, the Navigator will attempt to restore its state if it has been configured with a restoration id.

veggieseasons/lib/main.dart
RedBrogdon: +1 to @xster's note. Something like `registerForRestoration` would help readers know immediately what the call is doing.

veggieseasons/lib/screens/search.dart
```diff
@@ -12,19 +12,25 @@ import 'package:veggieseasons/widgets/search_bar.dart';
...
   final focusNode = FocusNode();
-  String terms = '';
+  String terms;
 
   @override
   void initState() {
     super.initState();
+    register(controller, const RestorationId('text'));
     controller.addListener(_onTextChanged);
+    terms = controller.value.text;
```
RedBrogdon, line 33: Would it be possible to just route the properties directly inside the `RestorableTextEditingController` class:

```
String get text => _value.text;
```

in order to make `terms = controller.value.text` back into `terms = controller.text`?

RedBrogdon:
Took the liberty of DO NOT SUBMIT => WIP, since that's the check our repo runs.

veggieseasons/lib/screens/search.dart
goderbauer: Interesting idea. I think that may be possible. Let me experiment with that...

goderbauer:
@RedBrogdon Here are answers to your questions:

> What is your plan for this PR? Are you actually wanting to land it, or is this mostly an experiment as you work through your ideas?

Eventually, I would like to submit this to have one example app with state restoration fully enabled. However, before that can happen, I need to finish the framework side API for state restoration and submit that to the flutter repository. In that process, this PR will probably change a little bit. If having this PR open for a long time is a problem, I am happy to close this. Opening this PR seems like the easiest way to solicit early feedback on API usage.

> Does this code need to be in the SDK itself, or could it be distributed as a plugin?

My plan is to add the state restoration functionality directly to the Flutter SDK. In theory, the functionality itself could be implemented in a plugin, but we also want to enable state restoration for the widgets that ship with flutter. To do that, the SDK would have to depend on this plugin, which is awkward and something we try to avoid. Therefore, I am just going to implement this directly in the framework.

>Is there a design doc for the API?

There is a half-backed one. I intent to finish that one up and make it available to the public to collect more feedback soon. I'll let you know.

> Have you worked through how the usage would be for someone building an app with a state management library like redux or BLoC?

I have not. If the state held by those libraries is serializable, it should _just work™_. Do we have an example redux/BLoC app somewhere to which I could add state restoration to prove that theory?

veggieseasons/lib/main.dart
goderbauer: Another idea I just had: `restoreProperty` because this is actually the call that retrieves the serialized value from the restoration state, deserializes it, and puts it into the property.

domesticmouse:
@goderbauer answers inline

> @RedBrogdon Here are answers to your questions:
> 
> > What is your plan for this PR? Are you actually wanting to land it, or is this mostly an experiment as you work through your ideas?
> 
> Eventually, I would like to submit this to have one example app with state restoration fully enabled. However, before that can happen, I need to finish the framework side API for state restoration and submit that to the flutter repository. In that process, this PR will probably change a little bit. If having this PR open for a long time is a problem, I am happy to close this. Opening this PR seems like the easiest way to solicit early feedback on API usage.

The only issue with a long open PR is the usual maintenance concerns with updates. If you don't mind us pressing the sync button on the PR as we land maintenance changes around this PR, then I have no concerns. This style of early preview on new APIs is gold from my POV.

> > Does this code need to be in the SDK itself, or could it be distributed as a plugin?
> 
> My plan is to add the state restoration functionality directly to the Flutter SDK. In theory, the functionality itself could be implemented in a plugin, but we also want to enable state restoration for the widgets that ship with flutter. To do that, the SDK would have to depend on this plugin, which is awkward and something we try to avoid. Therefore, I am just going to implement this directly in the framework.

SGTM

> > Is there a design doc for the API?
> 
> There is a half-backed one. I intent to finish that one up and make it available to the public to collect more feedback soon. I'll let you know.
> 
> > Have you worked through how the usage would be for someone building an app with a state management library like redux or BLoC?
> 
> I have not. If the state held by those libraries is serializable, it should _just work™_. Do we have an example redux/BLoC app somewhere to which I could add state restoration to prove that theory?

Here's the world's simplest Provider app, which should hopefully be a good stepping stone to things like redux and BLoC: https://github.com/flutter/samples/tree/master/provider_counter

goderbauer:
I've made some updates to incorporate the feedback I've gotten so far and to allow restoring to a different serialized state while the app is running (instead of just initializing the app from serialized state when it is launched).

veggieseasons/lib/main.dart
```diff
@@ -16,16 +16,76 @@ void main() {
...
+    );
+  }
+
+  @override
+  RestorationId get restorationId => const RestorationId('app');
+}
+
+// A custom RestorableProperty that saves and restores the data in an [AppState]
+// object.
+class RestorableAppState extends RestorableListenable<AppState> {
```
Hixie, line 67: is RestorableListenable a RestorableProperty? If so it would be nice if it had Property in the name for clarity.
Ditto RestorableAppState.

veggieseasons/lib/widgets/trivia.dart
```diff
@@ -23,24 +24,31 @@ enum PlayerStatus {
...
+class _TriviaViewState extends State<TriviaView> with RestorationMixin {
   /// Current app state. This is used to fetch veggie data.
   AppState appState;
 
   /// The veggie trivia about which to show.
   Veggie veggie;
 
   /// Index of the current trivia question.
-  int triviaIndex = 0;
+  RestorableNum<int> triviaIndex = RestorableNum<int>(0);
```
Hixie, line 35: I recommend just having RestorableInt and RestorableDouble, it's cleaner

Hixie: (they can just be defined subclasses of RestorableNum)

veggieseasons/lib/widgets/trivia.dart
```diff
@@ -203,12 +211,34 @@ class _TriviaViewState extends State<TriviaView> {
...
+
+  @override
+  Object serialize() {
+    return value.index;
+  }
+
+  @override
+  void didUpdateValue(PlayerStatus oldValue) {
+    notifyListeners();
+  }
```
Hixie, line 243: Can you make a RestorableEnum<T> ?

Hixie:
This looks great.

What's the testability story?

goderbauer:
> Is there a design doc for the API?

The design document for the restoration API is available here: http://flutter.dev/go/state-restoration-design

veggieseasons/lib/widgets/trivia.dart
goderbauer: Originally I looked into that, but since enums don't have a common base class other than Object I don't see a good way to do that.

goderbauer:
I updated this PR to the latest implementation of the state restoration framework in Flutter. It requires a version of flutter that's newer than this commit: https://github.com/flutter/flutter/commit/053ebf2c080c7b8efbf4020683a5ba27d9daa3b8. 

@RedBrogdon Is this something we can land in this repository as an example of how to make an app restorable? If so, I will add some tests to ensure that this doesn't break in the future.

RedBrogdon:
I'm in a quandary.

On the one hand, Having state restoration represented in the samples repo would be a clear win, and VeggieSeasons is a great way to do it. On the other, samples in the repo (other than the ones in `/experimental`) are intended to run on the stable channel, which wouldn't be possible if this change were landed.

We could:

* Update the version in place, and add a big note to its README that it no longer works on stable.
* Leave the new version in your fork and simply direct people there via social outreach.
* Move VeggieSeasons into the experimental directory for the next few months, leaving a redirect README in the old location.
* Create a separate copy of VeggieSeasons in the experimental directory for the time being, then delete it when the new APIs get to stable and the primary one could be updated.

WDYT? I would lean toward the second or third one.

goderbauer:
I understand your quandary. :) 

Given these options, I would prefer number 3:

 > Move VeggieSeasons into the experimental directory for the next few months, leaving a redirect README in the old location.

That way, there continues to be just one canonical copy of the app and we don't have to deal with multiple forks/copies that may become out of date and cause merge conflicts once state restoration fully lands on flutter stable.

If you're cool with that, I would adapt the PR to implement this option.

goderbauer:
@RedBrogdon This one is ready for review now. For a more sane review experience, I'd recommend reviewing the commits separately:

1. Making the app restorable: https://github.com/flutter/samples/pull/433/commits/be9222ccec7c93b4c954dd53a3bf592dfdba5f83
2. Moving it to experimental: https://github.com/flutter/samples/pull/433/commits/3cec26c432417457621a1e6cd37d23f1083dc76b

veggieseasons/lib/main.dart
```diff
@@ -17,19 +17,73 @@ void main() {
...
+        home: HomeScreen(restorationId: 'home'),
+        restorationScopeId: 'app',
       ),
-    ),
-  );
+    );
+  }
+}
+
+class _RestorableAppState extends RestorableListenable<AppState> {
```
RedBrogdon: Is it possible to merge this class with AppState itself, or is the better pattern to make a restorable wrapper as done here? I ask because between VeggieAppState, _RestorableAppState, and AppState, there's a whole lot of state now. 😄

veggieseasons/lib/screens/search.dart
RedBrogdon: I'm guessing this didn't end up being feasible/advisable? 😄

veggieseasons/lib/screens/details.dart
```diff
@@ -238,15 +238,42 @@ class InfoView extends StatelessWidget {
...
+  final String restorationId;
 
-  DetailsScreen(this.id);
+  DetailsScreen({this.id, this.restorationId});
+
+  static String show(NavigatorState navigator, int veggieId) {
+    return navigator.restorablePush<void>(_routeBuilder, arguments: veggieId);
+  }
+
+  static Route<void> _routeBuilder(BuildContext context, Object arguments) {
```
RedBrogdon, line 249: This bit is interesting to me, because it brings the type of route used to display the screen into the class itself, whereas that used to be something the launching screen was in control of (e.g. previously VeggieCard was responsible for deciding to show DetailsScreen as a dialog, and DetailsScreen itself was unconcerned with anything to do with how it was added to the navigation stack).

Is the pattern used here something we'd expect to see in most apps using State Restoration? It looks like the main reason for the change is to allow DetailsScreen to make sure its restoration ID to 'details', but I could be reading this wrong.

veggieseasons/README.md
```diff
@@ -1,42 +1,10 @@
...
-If you run into an issue with the sample itself, please file an issue
-in the [main Flutter repo](https://github.com/flutter/flutter/issues).
+The Veggie Seasons app has temporarily been moved to the
+[experimental](https://github.com/flutter/samples/tree/master/experimental/veggieseasons)
+directory of this repository due to the integration of state restoration into
+the app. Not all features of state restoration are available on the stable
+channel of flutter yet. Once commit
+[flutter/flutter@053ebf2](https://github.com/flutter/flutter/commit/053ebf2c080c7b8efbf4020683a5ba27d9daa3b8)
+has reached the stable channel, the app can move back to its old home in
+this directory.
```
RedBrogdon, line 10: Consider throwing something like

    For more info on how to run the samples in the `experimental` directory,
    see the [README](experimental/README.md) in that folder.

onto the end, just in case anyone's confuse as to why it went to that folder specifically.

RedBrogdon (commented):
I have a few questions, but that's it. :)

veggieseasons/lib/screens/search.dart
goderbauer: As mentioned in the comment about AppState, merging the two also makes it harder to throw good/proper error messages when you forget to register the class.

veggieseasons/lib/main.dart
goderbauer: Also, when https://github.com/flutter/samples/issues/573 is fixed, the _RestorableAppState can be removed because after the bug is fixed, the state of AppState would already be persisted in shared_preferences.

RedBrogdon: > You have one class that represents the actual value and another wrapper class that is in charge of serializing/deserializing that thing.

Got it. That makes sense. We (meaning DevRel) will want to remember that as a recommended practice when talking over state restoration.

veggieseasons/lib/screens/details.dart
RedBrogdon: A lot of people prefer to have the navigation info in the screens themselves, as you have it. I've seen plenty of apps that include route strings as constants in classes that define the Screens. I mostly wanted to make sure a) I understood what was going on, and b) whether this was a pattern you'd recommend for everyone.

No need to change it.

RedBrogdon (approved):

PR #435, opened 2020-05-08
Samples maintenance

domesticmouse:
Hey all, I just made a pass through the samples looking at warnings and following linter suggestions. There are some deprecation warnings in the `web` samples that I'm unsure how to fix. What is our policy on maintaining these samples? I'd advocate that they are either kept at latest revisions and warning free, or they get deleted/archived somewhere else.

Thoughts?


farr64:
Please:  Keep at latest revisions and warning free.

Thanks for all the work involved in doing so.

domesticmouse:
PTAL @RedBrogdon @johnpryan

johnpryan:
I vote to leave `web` directory as-is for now, it's likely we will have to either upgrade them to NNBD or drop them eventually.

place_tracker/lib/place_map.dart
```diff
@@ -47,7 +50,7 @@ class PlaceMapState extends State<PlaceMap> {
 
   LatLng _lastMapPosition;
 
-  Map<Marker, Place> _markedPlaces = Map<Marker, Place>();
+  final Map<Marker, Place> _markedPlaces = <Marker, Place>{};
```
johnpryan, line 53: nit: The type can be [inferred](https://dart.dev/guides/language/sound-dart#type-inference) with just `final _markedPlaces = <Marker, Place>{};` or `final Map<Marker, Place> _markedPlaces = {};`

veggieseasons/lib/data/preferences.dart
```diff
@@ -20,7 +20,7 @@ class Preferences extends Model {
 
   int _desiredCalories = 2000;
 
-  Set<VeggieCategory> _preferredCategories = Set<VeggieCategory>();
+  final Set<VeggieCategory> _preferredCategories = <VeggieCategory>{};
```
johnpryan, line 23: Same here

place_tracker/lib/place_map.dart
domesticmouse: I believe the current style (at least as enforced by the lint checks) is to turn method locals into `var`s, but to leave class fields as fully typed.

veggieseasons/lib/data/preferences.dart
domesticmouse: Yup, this is also a class field

place_tracker/lib/place_map.dart
johnpryan: I would vote for the second option then.

veggieseasons/lib/data/preferences.dart
johnpryan: same

place_tracker/lib/place_map.dart
domesticmouse: Wanna spin up a new PR with your suggestion? This PR has landed and I've already deleted the source repo =)

PR #441, opened 2020-05-13
change slide_puzzle to a submodule

johnpryan:


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "web/slide_puzzle"]
+	path = web/slide_puzzle
+	url = git@github.com:kevmoo/slide_puzzle.git
```
kevmoo, line 3: you'll have to use an https URI to make travis happy – or anyone else who is not authenticated

johnpryan: ah, good catch

RedBrogdon:
Do you mind providing some context for the change? Was @kevmoo already maintaining a copy of the app elsewhere, and that's the real source of truth?

If so, would it be possible to host it using that repo's hosting gh-pages, and merely link to it from the samples index?

domesticmouse:
> [...] would it be possible to host it using that repo's hosting gh-pages, and merely link to it from the samples index?

If my memory serves (and it usually doesn't), this was @kevmoo's sample originally, but the `/web/slide_puzzle` sample was updated to work with Flutter for web. I'd suggest we go down the path that we used for gallery and replace `/web/slide_puzzle/`'s content with a README.md that just point's at @kevmoo's repo.

johnpryan:
> Do you mind providing some context for the change? 

The samples index build fails whenever there's a breaking change in the SDK. https://github.com/kevmoo/slide_puzzle/pull/7 provides some context. Rather than dupe the changes in two places, it makes more sense to switch it to a submodule.

> If so, would it be possible to host it using that repo's hosting gh-pages, and merely link to it from the samples index?

> I'd suggest we go down the path that we used for gallery and replace /web/slide_puzzle/'s content with a README.md that just points at @kevmoo's repo.

I would prefer to make sure the web demos that we link to are built using the latest version of the Flutter SDK. I'd rather not bug maintainers to keep updating their gh-pages branch. (Except for the gallery.)

kevmoo:
FYI – I'm also happy to delete my personal copy and make this repo the canonical home!

domesticmouse:
Thanks @kevmoo that sounds brilliant 😁

johnpryan:
👍 Merging this for now

PR #442, opened 2020-05-14
[animations] Add widget tests for card_swipe

abd99:
This PR adds widget tests for Card Swipe in the Animations app.
@domesticmouse @RedBrogdon reviews, please.

### DIFF
--- a/animations/test/misc/card_swipe_test.dart
+++ b/animations/test/misc/card_swipe_test.dart
@@ -16,7 +16,7 @@ void main() {
     testWidgets('One card swiped out', (tester) async {
       await tester.pumpWidget(createCardSwipeScreen());
 
-      // Get the total number of cards availabe.
+      // Get the total number of cards available.
       var totalCards = tester.widgetList(find.byType(SwipeableCard)).length;
 
       // Swipe out one card.
--- a/animations/test/misc/card_swipe_test.dart
+++ b/animations/test/misc/card_swipe_test.dart
@@ -1,0 +1,66 @@
+// Copyright 2020 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:animations/src/misc/card_swipe.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+Widget createCardSwipeScreen() => MaterialApp(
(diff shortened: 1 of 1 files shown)


animations/test/misc/card_swipe_test.dart
```diff
@@ -0,0 +1,66 @@
...
+      // Swipe out one card.
+      await tester.drag(find.byType(SwipeableCard).first, Offset(100.0, 0.0));
+      await tester.pumpAndSettle();
+
+      // Check if removed properly.
+      expect(tester.widgetList(find.byType(SwipeableCard)).length,
+          lessThan(totalCards));
+    });
+
+    testWidgets('All cards swiped out', (tester) async {
```
abd99, line 31: I'm not sure if we should be adding a test to swipe out all the cards. What do you think?

domesticmouse:
PTAL @RedBrogdon =)

animations/test/misc/card_swipe_test.dart
domesticmouse: I'm curious how that'd work. It's worth trying =)

animations/test/misc/card_swipe_test.dart
```diff
@@ -0,0 +1,66 @@
...
+      }
+
+      // Check if any card is remaining.
+      expect(find.byType(SwipeableCard), findsNothing);
+    });
+
+    testWidgets('Stack refilled with cards', (tester) async {
+      await tester.pumpWidget(createCardSwipeScreen());
+
+      // Get the total number of cards availabe.
```
RedBrogdon, line 51: nit: There's a typo in "available."

animations/test/misc/card_swipe_test.dart
RedBrogdon: I think the question is whether this test adds value beyond what the previous test provides. If the previous test passes, we already know that swiping a stack of cards once will remove the top card, so this test could seen as verifying that:

* Multiple swipes in a row also work.
* The app doesn't crash or have another weird problem once there are no more cards.

Like @domesticmouse, I think this is probably fine as it is.

RedBrogdon: I'll leave this umerged for a day, @abd99, in case you want to make further changes. Looks ready to land to me, though.

RedBrogdon:
LGTM, other than the one typo.

abd99:
I've fixed the typo. @RedBrogdon, you can go ahead and merge it.
Thanks!

PR #447, opened 2020-05-27
[web_dashboard] add logout

johnpryan:
This adds a logout button, a widget to switch between the login page and app page, and makes the the sign in screen call the `onSuccess` callback immediately if the user is already signed in (for example if they sign in and refresh the page, they are taken directly to the app and not asked to click the sign in button.)

The `MockAuthService`'s method,  `isSignedIn`, always returns false, so the sign in button needs to be pressed each time the app loads if using this configuration.

### DIFF
--- a/experimental/web_dashboard/lib/src/app.dart
+++ b/experimental/web_dashboard/lib/src/app.dart
@@ -77,13 +77,17 @@ class _DashboardAppState extends State<DashboardApp> {
   void _handleSignIn(User user, BuildContext context, AppState appState) {
     appState.api = widget.apiBuilder(user);
 
-    _showPage(HomePage(), context);
+    _showPage(HomePage(onLogout: _handleLogout), context);
   }
 
   /// Navigates to the home page using a fade transition.
   void _showPage(Widget page, BuildContext context) {
     var route = _fadeRoute(page);
     Navigator.of(context).pushReplacement(route);
+  }
+
+  void _handleLogout() {
+
   }
 
   /// Creates a [Route] that shows [newPage] using a fade transition.
--- a/experimental/web_dashboard/lib/src/app.dart
+++ b/experimental/web_dashboard/lib/src/app.dart
@@ -65,10 +65,11 @@ class _DashboardAppState extends State<DashboardApp> {
     return Provider.value(
       value: _appState,
       child: MaterialApp(
-          home: SignInSwitcher(
-        appState: _appState,
-        apiBuilder: widget.apiBuilder,
-      )),
+        home: SignInSwitcher(
+          appState: _appState,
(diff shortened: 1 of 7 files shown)


experimental/web_dashboard/lib/src/pages/sign_in.dart
```diff
@@ -20,24 +20,48 @@ class SignInPage extends StatefulWidget {
...
+    var alreadySignedIn = await widget.auth.isSignedIn();
+    if (alreadySignedIn) {
+      var user = await widget.auth.signIn();
+      widget.onSuccess(user);
+    }
+    return alreadySignedIn;
+  }
+
+  Future<void> _signIn() async {
+    var user = await widget.auth.signIn();
```
RedBrogdon, line 44: Is it possible for `signIn`'s Future to resolve with an error rather than data?

experimental/web_dashboard/lib/src/pages/sign_in.dart
```diff
@@ -20,24 +20,48 @@ class SignInPage extends StatefulWidget {
...
   @override
   void initState() {
     super.initState();
+    _checkSignInFuture = _checkIfSignedIn();
+  }
+
+// Check if the user is signed in. If the user is already signed in (for
+// example, if they signed in and refreshed the page), invoke the `onSuccess`
+// callback right away.
+  Future<bool> _checkIfSignedIn() async {
```
RedBrogdon, line 34: Because this is a sample, you wouldn't be incorrect to remove this section for simplicity's sake and instead expect users to hit the button each time they want to login.

Totally optional, but I figured I'd put it out there.

experimental/web_dashboard/lib/src/app.dart
```diff
@@ -63,41 +65,61 @@ class _DashboardAppState extends State<DashboardApp> {
...
       value: _appState,
       child: MaterialApp(
-        home: Builder(
-          builder: (context) => SignInPage(
-            auth: _appState.auth,
-            onSuccess: (user) => _handleSignIn(user, context, _appState),
-          ),
-        ),
-      ),
+          home: SignInSwitcher(
```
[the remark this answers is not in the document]
johnpryan: Hmm, that might be a better way. Will take a look.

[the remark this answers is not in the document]
johnpryan: Will revisit this when we have improved routing

experimental/web_dashboard/lib/src/pages/sign_in.dart
johnpryan: This is certainly reaching the point of diminishing returns. It might be worth taking out and capturing this pattern as an article or something.

Issue #449, opened 2020-05-31
analysis_options.yaml link is broken in CONTRIBUTING.md


AyushBherwani1998:
The `analysis_options.yaml` is linked to the gallery sample, which does not exist anymore in the repo.

PR #450, opened 2020-06-01
Linking to a analysis_options.yaml that is present

domesticmouse:
fixes: https://github.com/flutter/samples/issues/449


domesticmouse:
Oddly, Travis is showing the required tests as all green. Unsure where in the test process it failed to notify. Merging as is.

PR #463, opened 2020-06-08
add form_app

johnpryan:
This adds a sample demonstrating some new features like autofill, and some other patterns common with forms (validation, using other form widgets, and sending HTTP requests). 

cc: @nturgut @LongCatIsLooong 

Screen Shot 2020-06-08 at 12 16 17 PM
Screen Shot 2020-06-08 at 12 16 24 PM
Screen Shot 2020-06-08 at 12 16 28 PM
Screen Shot 2020-06-08 at 12 16 31 PM
Screen Shot 2020-06-08 at 12 16 35 PM


experimental/form_app/lib/src/autofill.dart
```diff
@@ -0,0 +1,101 @@
...
+                        labelText: 'Last Name',
+                      ),
+                      autofillHints: [AutofillHints.familyName],
+                    ),
+                    TextField(
+                      decoration: InputDecoration(
+                        hintText: 'foo@example.com',
+                        labelText: 'Email',
+                      ),
+                      autofillHints: <String>[AutofillHints.email],
```
theacodes, line 48: Why is the `<String>` needed here but not in the two fields above?

experimental/form_app/lib/src/sign_in_http.dart
```diff
@@ -0,0 +1,122 @@
...
+  SignInHttpDemo({
+    this.httpClient,
+  });
+
+  @override
+  _SignInHttpDemoState createState() => _SignInHttpDemoState();
+}
+
+class _SignInHttpDemoState extends State<SignInHttpDemo> {
+  final _formKey = GlobalKey<FormState>();
```
theacodes, line 37: It'd be nice to explain why this is needed with a comment. It's the only thing I found surprising.

experimental/form_app/lib/src/validation.dart
```diff
@@ -0,0 +1,161 @@
...
+      appBar: AppBar(
+        title: Text('📖 Story Generator'),
+        actions: [
+          Padding(
+            padding: EdgeInsets.all(8),
+            child: FlatButton(
+              textColor: Colors.white,
+              child: Text('Submit'),
+              onPressed: () {
+                // When the form is submitted, validate it by getting the FormState from the the GlobalKey.
```
theacodes, line 27: typo: `the the`.

experimental/form_app/lib/src/form_widgets.dart
```diff
@@ -0,0 +1,208 @@
...
+                          children: [
+                            Text('Enable feature',
+                                style: Theme.of(context).textTheme.bodyText1),
+                            Switch(
+                              value: enableFeature,
+                              onChanged: (enabled) {
+                                setState(() {
+                                  enableFeature = enabled;
+                                });
+                              },                      ),
```
RedBrogdon, line 126: How did this formatting happen?

experimental/form_app/lib/src/form_widgets.dart
```diff
@@ -0,0 +1,208 @@
...
+        key: _formKey,
+        child: Scrollbar(
+          child: Align(
+            alignment: Alignment.topCenter,
+            child: Card(
+              child: SingleChildScrollView(
+                padding: EdgeInsets.all(16),
+                child: ConstrainedBox(
+                  constraints: BoxConstraints(maxWidth: 400),
+                  child: Column(
```
RedBrogdon, line 34: Is it possible to present these fields using a ListView rather than a Scrollable + Column? For performance reasons, I believe the best practice is to prefer ListView (In order for Column to size itself, it has to layout all its children, which a ListView can sometimes avoid).

experimental/form_app/lib/src/form_widgets.dart
```diff
@@ -0,0 +1,208 @@
...
+                            labelText: 'Description',
+                          ),
+                          onChanged: (value) {
+                            description = value;
+                          },
+                          maxLines: 5,
+                        ),
+                        _FormDatePicker(
+                          onChanged: (value) {
+                            date = value;
```
RedBrogdon, line 63: This is modifying state without calling setState, which is a problem. Widgets like this are weird, because they have complex UI state to represent the value, but need to still report the data state to their parent.

The pattern used in the framework for a lot of widgets like this is to have the parent widget keep track of the state and then inject it back into the child when it's changed. The Slider that's a few lines down does this:

```dart
Slider(
  min: 0,
  max: 500,
  divisions: 500,
  value: maxValue,
  onChanged: (value) {
    setState(() {
      maxValue = value;
    });
  },
),
```

When the user does a gesture to move the slider, there's a little dance that's done:

* Slider calls its onChanged method.
* The parent updates its field for the slider value in setState (`maxValue` in this case).
* The parent's build method is called, creating a new Slider with the new value.
* Slider rebuilds to display the new data, which causes the little knob to actually move onscreen.

I believe you could put the same flow to work here.

experimental/form_app/lib/src/http/mock_client.dart
```diff
@@ -0,0 +1,27 @@
+import 'dart:convert';
+
+import 'package:http/http.dart' as http;
+import 'package:mockito/mockito.dart';
+
+class MockClient extends Mock implements http.Client {
```
RedBrogdon, line 6: This is great. It's still simple enough for beginners, but way better than relying a real request to something external.

experimental/form_app/lib/src/sign_in_http.g.dart
```diff
@@ -0,0 +1,19 @@
+// GENERATED CODE - DO NOT MODIFY BY HAND
+
+part of 'sign_in_http.dart';
+
+// **************************************************************************
+// JsonSerializableGenerator
+// **************************************************************************
+
+FormData _$FormDataFromJson(Map<String, dynamic> json) {
```
RedBrogdon, line 9: Entirely your call, but consider rolling these by hand in order to remove the dependency on source_gen.

experimental/form_app/lib/src/validation.dart
```diff
@@ -0,0 +1,161 @@
...
+      appBar: AppBar(
+        title: Text('📖 Story Generator'),
+        actions: [
+          Padding(
+            padding: EdgeInsets.all(8),
+            child: FlatButton(
+              textColor: Colors.white,
+              child: Text('Submit'),
+              onPressed: () {
+                // Validate the form by getting the FormState from the GlobalKey
```
RedBrogdon, line 27: Great place for a comment. I would have needed to check the docs to make sure I understood line 29.

experimental/form_app/lib/src/validation.dart
```diff
@@ -0,0 +1,161 @@
...
+                      return 'You must agree to the terms of service.';
+                    }
+                    return null;
+                  },
+                  builder: (FormFieldState formFieldState) {
+                    return Column(
+                      crossAxisAlignment: CrossAxisAlignment.start,
+                      children: [
+                        Row(
+                          mainAxisAlignment: MainAxisAlignment.start,
```
RedBrogdon, line 122: nit: Start is the default, so this can be cut.

experimental/form_app/lib/src/validation.dart
```diff
@@ -0,0 +1,161 @@
...
+                          children: [
+                            Checkbox(
+                              value: agreedToTerms,
+                              onChanged: (value) {
+                                // When the value of the checkbox changes, we
```
RedBrogdon, line 127: nit: @kwalrath can correct me here (srsly!), but generally I try to avoid first person pronouns in comments like this. My usual cheat is to say, "the app..." in place of "we" or "I."

experimental/form_app/lib/main.dart
```diff
@@ -0,0 +1,90 @@
+import 'package:flutter/material.dart';
+import 'package:form_app/src/validation.dart';
```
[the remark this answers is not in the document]
kwalrath: FYI, we're now recommending relative paths for inside-package imports that don't cross the /lib boundary. (This isn't well known; we should perhaps update Effective Dart, and we should add the "What's new" page proposed in https://github.com/dart-lang/site-www/issues/2412.)

https://github.com/dart-lang/site-www/pull/2443 is the change that added the relative path recommendation to https://dart.dev/guides/libraries/create-library-packages#importing-library-files.

@munificent should we add a recommendation to prefer relative paths to Effective Dart?

experimental/form_app/lib/src/validation.dart
kwalrath: Agreed. You could even drop the subject:

```
// When the value of the checkbox changes, update
// the FormFieldState so the form is revalidated.
```

experimental/form_app/lib/main.dart
```diff
@@ -0,0 +1,90 @@
+import 'package:flutter/material.dart';
```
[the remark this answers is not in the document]
johnpryan: embraced the yak shave on this one: [package:copywriter](https://pub.dev/packages/copywrite)

experimental/form_app/lib/src/form_widgets.dart
johnpryan: I'm not sure either way is right or wrong here, since there's small number of children. In this case, doesn't this guarantee that the subtree is layed out once as opposed to potentially multiple builds / layouts with ListView?

experimental/form_app/lib/src/form_widgets.dart
johnpryan: That makes more sense.  d0aaf7add5133be9c753f4286bcf9789272abcba

experimental/form_app/lib/main.dart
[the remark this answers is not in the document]
RedBrogdon: We should socialize that. Is there a specific reason for the recommendation?

experimental/form_app/lib/src/form_widgets.dart
RedBrogdon: You're right. Now that I'm looking at this, I'm not sure. It's definitely not going to be noticeable with this few widgets, so it's only interesting from a "do we care enough to show it as the default technique every time" kind of thing.

With a Column, each child has to be given an element and a render object, laid out, and so on. With the ListView, that's done lazily, which can mean quicker first-frame time for long lists. Since those elements and render objects will be disposed once they're scrolled out of frame, though, it's possible that the Column with a small or medium amount of children would be faster at scrolling.

I may ask Goderbauer about this when we talk to him about the state restoration stuff. He knows package:widgets inside and out.

experimental/form_app/lib/main.dart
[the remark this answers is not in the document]
kwalrath: @mit-mit or @munificent would know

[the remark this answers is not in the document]
munificent: The only real reason is consistency, but consistency is good.

PR #465, opened 2020-06-09
Add a pigeon sample that demonstrates a "realistic" integration scenario with middleware and business logic

xster:
This sample shows what you would do to pass business logic data in and out of a Flutter screen using pigeon. The existing app has existing middleware that populates the business logic data. GSON, OkHttp and Google Books API are semi-arbitrarily selected here to represent middleware (partly based on go/popular-android-libraries-and-apis). 

Kapture 2020-06-08 at 18 48 46

Unfortunately I only have time to do the Android part before I go on leave. @gaaclarke, could you fill in an iOS part too after?


add_to_app/flutter_module_books/lib/main.dart
```diff
@@ -0,0 +1,191 @@
...
+      (Book book) {
+        setState(() {
+          // This book model is what we're going to return to Kotlin eventually.
+          // Keep it bound to the UI.
+          this.book = book;
+          titleTextController.text = book.title;
+          titleTextController.addListener(() {
+            this.book?.title = titleTextController.text;
+          });
+          subtitleTextController.text = book.subtitle ?? '';
```
RedBrogdon, line 77: The `??`operator sticks out a bit here. Is it possible to correct for this (substitute an empty string for null) when the book record is first downloaded?

add_to_app/flutter_module_books/lib/main.dart
```diff
@@ -0,0 +1,191 @@
...
+    FlutterBookApi.setup(FlutterBookApiHandler(
+      // The `FlutterBookApi` just has one method. Just give a closure for that
+      // method to the handler class.
+      (Book book) {
+        setState(() {
+          // This book model is what we're going to return to Kotlin eventually.
+          // Keep it bound to the UI.
+          this.book = book;
+          titleTextController.text = book.title;
+          titleTextController.addListener(() {
```
RedBrogdon, line 74: Could these listeners be set up outside the call to setState, since they don't depend on the book itself until they're run?

add_to_app/flutter_module_books/lib/main.dart
```diff
@@ -0,0 +1,191 @@
...
+  FocusNode textFocusNode = FocusNode();
+  TextEditingController titleTextController = TextEditingController();
+  TextEditingController subtitleTextController = TextEditingController();
+  TextEditingController authorTextController = TextEditingController();
+
+  @override
+  void initState() {
+    super.initState();
+
+    // This `HostBookApi` class instance lets us make outgoing calls to the platform.
```
RedBrogdon, line 60: nit: For *.dart, we generally stick to an 80-char line.

add_to_app/flutter_module_books/test/widget_test.dart
```diff
@@ -0,0 +1,30 @@
+// This is a basic Flutter widget test.
```
RedBrogdon, line 1: This test is broken, right? Is there a way to replace it with a working test?

I ask because the add-to-app apps and modules are currently all integrated into CI. In the `/tool` directory there's a script for pure Flutter stuff and two more for Android and iOS. This should be wired into those scripts before it's landed.

add_to_app/flutter_module_books/pubspec.yaml
```diff
@@ -0,0 +1,30 @@
+name: flutter_module_books
+description: A new Flutter project.
```
RedBrogdon, line 2: nit: new description

add_to_app/flutter_module_books/test/widget_test.dart
RedBrogdon: You don't have to go as far as the espresso tests that are in one of the Android apps, I should mention.

add_to_app/android_books/app/src/androidTest/java/dev/flutter/example/books/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
+    @Test
+    fun useAppContext() {
+        // Context of the app under test.
+        val appContext = InstrumentationRegistry.getInstrumentation().targetContext
+        assertEquals("dev.flutter.example.books", appContext.packageName)
+    }
+}
```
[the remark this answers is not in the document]
xster: Oh man, good catch. It should be on all the files.

add_to_app/android_books/app/src/main/java/dev/flutter/example/books/MainActivity.kt
```diff
@@ -0,0 +1,161 @@
...
+        for (jsonBook in jsonBooks.map { it.asJsonObject }) {
+            try {
+                // Here we're using GSON to populate a Pigeon data class directly. The Pigeon data
+                // class can be used not just as part of your IPC API's signature but also as a
+                // normal data class in your existing application.
+                //
+                // We could either push the Pigeon data class usage higher into the existing GSON
+                // "middleware" or lower, closer to the IPC.
+                val book = Api.Book()
+                val volumeInfoJson = jsonBook.getAsJsonObject("volumeInfo")
```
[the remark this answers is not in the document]
xster: That's not possible :)

This code is about breaking out data from the Google Books specific return data.

add_to_app/flutter_module_books/lib/main.dart
```diff
@@ -0,0 +1,191 @@
...
+import 'package:flutter_module_books/api.dart';
+
+void main() => runApp(MyApp());
+
+class MyApp extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      theme: ThemeData(
+        primaryColor: const Color(0xff6200ee),
```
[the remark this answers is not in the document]
xster: So that's the new material swatch colors. We don't have it in Flutter yet. i.e. it's not one of the Colors. It's in ColorSwatch now but widgets aren't reading it yet (like the AppBar which is what I want).

add_to_app/README.md
```diff
@@ -26,20 +28,26 @@ See https://guides.cocoapods.org/using/getting-started.html for more details.
...
+  platform and Flutter.
+
 
 Before using them, you need to resolve the Flutter modules' dependencies. Do so
 by running this command from within the `flutter_module` and
 `flutter_module_using_plugin` directories:
 
 ```bash
-flutter packages get
+flutter pub get
```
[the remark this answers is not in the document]
xster: I've been told that `flutter packages get` is "deprecated". But I don't have any docs. Maybe @jmagman knows.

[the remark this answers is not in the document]
jmagman: I think it was all renamed in https://github.com/flutter/flutter/pull/33041

add_to_app/android_books/app/src/main/java/dev/flutter/example/books/FlutterBookActivity.kt
```diff
@@ -0,0 +1,97 @@
...
+    // This {@link Api.HostBookApi} subclass will be called by Pigeon when the corresponding
+    // APIs are invoked on the Dart side.
+    inner class HostBookApiHandler: Api.HostBookApi {
+        override fun cancel() {
+            // Flutter calls cancel. Finish the activity with a cancel result.
+            setResult(Activity.RESULT_CANCELED)
+            finish()
+        }
+
+        override fun finishedEditingBook(book: Api.Book?) {
```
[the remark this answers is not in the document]
xster: ha it's a good way to test what happens if I rename the schema under the code

add_to_app/flutter_module_books/lib/main.dart
xster: the setState closure doesn't actually do anything :)

I think it's a bit clearer together since we're hooking in 2 directions per controller.

add_to_app/flutter_module_books/lib/main.dart
xster: ya, I thought about it a bit when writing too. Though I thing generally, expecting the contract to be in the other side's implementation isn't right. The contract should be in the schema. Though in this case, it means https://github.com/flutter/flutter/issues/59118 :D

xster:
I'm gonna be out for a month soon. If you want me to fix up anything on this we should do it soon :)

add_to_app/README.md
```diff
@@ -26,20 +28,26 @@ See https://guides.cocoapods.org/using/getting-started.html for more details.
...
+There are three Flutter modules included in the codebase:
 
 * `flutter_module` displays the dimensions of the screen, a button that
   increments a simple counter, and an optional exit button.
 * `flutter_module_using_plugin` does everything `flutter_module` does and adds
   another button that will open the Flutter documentation in a browser using the
   [`url_launcher`](https://pub.dev/packages/url_launcher) Flutter plugin.
+* `flutter_module_books` simulates an integration scenario with existing
+  platform business logic and middleware. It uses the [`pigeon`](https://pub.dev/packages/pigeon)
+  plugin to facilitate integration ergonomics by generating the platform channel
```
gaaclarke, line 40: nit: ~~facilitate integration ergonomics~~ make integration easier

add_to_app/android_books/app/src/main/java/dev/flutter/example/books/FlutterBookActivity.kt
```diff
@@ -0,0 +1,101 @@
...
+                .putExtra(
+                    // The Pigeon data class is useful not only between Kotlin/Java and Dart
+                    // but also within Kotlin/Java where activities must communicate with
+                    // each other via serializable data. The Pigeon data class is a
+                    // serializable class by definition.
+                    EXTRA_BOOK,
+                    // TODO(gaaclarke): the Pigeon generated data class should just implement
+                    // Serializable so we won't need 'toMap()' here
+                    // https://github.com/flutter/flutter/issues/58909
+                    book.toMap()
```
gaaclarke, line 46: On other platforms toMap and fromMap are private.  I should make them private or visible for testing only in java too, not package visible.  We don't want people to get reliant on them.

xster: +1. We shouldn't let people have access to this. Though we should make sure it implements Serializable first so that the data classes can be used in people's existing projects (needed to share data classes between activities) in addition to just the interop.

gaaclarke: I don't think so, once we provide a serialization format we have to be backward compatible.  By hiding the actual serialization behind functions we can change it at any point.

add_to_app/flutter_module_books/README.md
```diff
@@ -0,0 +1,42 @@
+# Books add-to-app sample
+
+This application simulates a mock scenario where an existing app with
+business logic and middleware already exists. This sample demonstrates how to
+do an add-to-app Flutter integration into existing conventions.
+
+This application also utilizes the [Pigeon](https://pub.dev/packages/pigeon)
+plugin to avoid manual platform channel wiring. The Pigeon autogenerates the
```
gaaclarke, line 8: ~~The Pigeon~~ Pigeon

xster: lol oops

add_to_app/android_books/app/src/main/java/dev/flutter/example/books/FlutterBookActivity.kt
xster: Discussed offline. This is a widely used Android pattern.

xster:
@RedBrogdon any thoughts? I got 1 business day left before I leave.

add_to_app/flutter_module_books/lib/main.dart
RedBrogdon: Yeah, this becomes a different thing entirely once NNBD lands.

xster: ya, hopefully it's a good thing. Since we don't want to create custom annotations for Kotlin and Swift currently to make those sides more ergonomic. We can then piggy back off of the dart NNBD syntax once it matures a bit.

add_to_app/README.md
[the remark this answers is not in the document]
RedBrogdon: TIL. Definitely keep as-is.

add_to_app/android_books/app/src/androidTest/java/dev/flutter/example/books/ExampleInstrumentedTest.kt
```diff
@@ -14,11 +14,14 @@ import org.junit.Assert.*
...
     @Test
     fun useAppContext() {
         // Context of the app under test.
         val appContext = InstrumentationRegistry.getInstrumentation().targetContext
         assertEquals("dev.flutter.example.books", appContext.packageName)
     }
+
+    // The app should be hermetic (with offline books JSON) before adding
+    // more tests.
 }
```
RedBrogdon, line 27: nit: newline

add_to_app/flutter_module_books/lib/main.dart
RedBrogdon: Fair enough.

PR #479, opened 2020-06-19
fix ios install

rubensdemelo:
To use this plugin on iOS you need to opt-in for the embedded views preview by adding a boolean property to the app's Info.plist file, with the key io.flutter.embedded_views_preview and the value YES.

<key>io.flutter.embedded_views_preview</key>
<true/>

### DIFF
--- a/place_tracker/ios/Runner/Info.plist
+++ b/place_tracker/ios/Runner/Info.plist
@@ -41,5 +41,7 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>io.flutter.embedded_views_preview</key>
+	<true/>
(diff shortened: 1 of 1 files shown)


RedBrogdon:
I want to say I remember running this app on iOS manually when it was first added to the repo. @cbracken or @goderbauer, do you know if the plist requirement was added later? Either way, could one of you aver that this is the current best practice (it certainly seems to be, but I figured I'd check).

domesticmouse:
This plist key is required by the [`google_maps_flutter`](https://pub.dev/packages/google_maps_flutter) plugin. The way to catch this would be to attempt to run this app on an iOS simulator, and have it fail to render. If memory serves, without this key the sample will fail to launch.

PR #490, opened 2020-07-08
[testing_app] Add State Management Tests

abd99:
State Management Tests using Flutter Driver added.

/cc: @domesticmouse @RedBrogdon

### DIFF
--- a/testing_app/test_driver/state_mgmt.dart
+++ b/testing_app/test_driver/state_mgmt.dart
@@ -1,0 +1,42 @@
+// Copyright 2020 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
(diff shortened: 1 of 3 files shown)


testing_app/test_driver/state_mgmt.dart
```diff
@@ -0,0 +1,42 @@
...
+  Favorites favoritesList;
+
+  // This line enables the extension.
+  enableFlutterDriverExtension(
+      // This handler gets data from the test file
+      // and then we use it to perform operations on the state object.
+      handler: (command) {
+    if (command == 'ADD') {
+      favoritesList.add(30);
+    } else if (command == 'REMOVE') {
```
abd99, line 23: I've used this method for communication between the application and the test file. It's working fine. Please let me know if there is any other way that I should be using.

PR #492, opened 2020-07-10
Updates samples for latest beta release of the Flutter SDK

RedBrogdon:
Landing some updates to the samples into the beta branch now that a new beta version of the Flutter SDK has been released:

* Updates to a bunch of `pubspec.lock` files.
* Fixes some casts in json example that were failing static analysis in the new version.
* Reruns source_gen for `built_value`-based classes.
* Adds a line to the Flutter test script to manually get packages before running tests.


jsonexample/lib/tab_pages.dart
```diff
@@ -43,20 +41,22 @@ class BasicsPage extends StatelessWidget {
     final boldStyle =
         localTheme.bodyText2.copyWith(fontWeight: FontWeight.w600);
 
-    final dynamic dynamicListOfInts = json.decode(JsonStrings.listOfInts);
+    final dynamicListOfInts = json.decode(JsonStrings.listOfInts) as Iterable;
```
johnpryan, line 44: nit: `as Iterable` works, but `json.decode` will always return Lists, right?

tool/travis_flutter_script.sh
```diff
@@ -48,6 +48,9 @@ do
     echo "== Testing '${PROJECT_NAME}' on Flutter's ${FLUTTER_VERSION} channel =="
     pushd "${PROJECT_NAME}"
 
+    # Grab packages.
+    "${LOCAL_SDK_PATH}/bin/flutter" pub get
```
johnpryan, line 52: Weird that `flutter analyze` runs `pub get` on a fresh clone for me. 

It's probably better not to rely on this behavior anyway, since it's not clear what criteria it's using.

RedBrogdon: Weird. Now that I've double-checked it, I'm getting the same result as you. My local copy must have somehow gotten into a weird state, which I now can't reproduce.

> It's probably better not to rely on this behavior anyway, since it's not clear what criteria it's using.

Agreed.

jsonexample/lib/tab_pages.dart
RedBrogdon: True.

I was using Iterable because that's what `list.from` has in its signature. Now that I'm looking with fresh eyes, though, List makes more sense and matches the variable names better, so I've updated the code.

RedBrogdon:
@johnpryan, I just rebased on master to pull any late-arriving changes into the PR, and then set master as the target branch. This code should be good to go.

johnpryan:
Sounds good, will merge tomorrow

PR #504, opened 2020-07-23
Resolve conflicts caused when trying to merge testing_sample into master

abd99:
Actually, the imports in `platform_channels/lib/main.dart` weren't sorted in `master`. So, I was accepting the current change. It was totally out of my mind that it would again cause a conflict when merging back to `master`.

@domesticmouse, it should probably work now =)

### DIFF
--- a/platform_channels/lib/main.dart
+++ b/platform_channels/lib/main.dart
@@ -4,9 +4,9 @@
 
 import 'package:flutter/material.dart';
 import 'package:platform_channels/src/add_pet_details.dart';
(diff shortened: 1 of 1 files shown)


AyushBherwani1998:
@abd99 Nice catch for unsorted import. It does have an unsorted import in `main.dart`. 

@domesticmouse @RedBrogdon shouldn't the analyzer throw a warning. I remember last time unsorted import in animations sample failed tests too.

domesticmouse:
@AyushBherwani1998 it totally depends on what lint checks you have enabled for your sample.

PR #500, opened 2020-07-20
Testing sample 

domesticmouse:
This PR lands @abd99's testing sample in `master`.

PR #508, opened 2020-07-30
Changes for testing_sample

abd99:
This PR contains changes requested in #500.

I guess this is the only way to do it as I don't have write access to the repo.

/cc @domesticmouse


abd99:
Thanks for the super quick response! Can you please re-request a review there.

PR #521, opened 2020-08-13
Upgrade provider shopper

filiph:


provider_shopper/analysis_options.yaml
```diff
@@ -7,7 +7,6 @@ analyzer:
 
 linter:
   rules:
-    - avoid_types_on_closure_parameters
```
johnpryan, line 10: We generally try to keep `analysis_options.yaml` files consistent for all the samples. Do we need to remove this lint?

johnpryan:
Just one question, otherwise LGTM

provider_shopper/analysis_options.yaml
filiph: Ah, this is a left over change and isn't actually needed. I re-added it.

filiph:
John, I can't merge until you give me approval in the github UI. Or, you can just merge yourself if that's easier.

PR #534, opened 2020-08-27
Added a build script that can help document how to build the sample.

gaaclarke:
relevant issue: https://github.com/flutter/flutter/issues/64675#issuecomment-681178517

### DIFF
--- a/add_to_app/ios_fullscreen/build.sh
+++ b/add_to_app/ios_fullscreen/build.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+###############################################################################
+# A convenience script that runs through the setup tasks for building this
+# sample.  A description of the steps can be found at:
+# https://github.com/flutter/samples/blob/master/add_to_app/README.md
+###############################################################################
+
+WHOAMI=`whoami`
+
+cd ../flutter_module
+sed -i .backup 's/dev.flutter.example.flutterModule/dev.flutter.'"$WHOAMI"'.flutterModule/g' pubspec.yaml
+flutter pub get
+cd -
+pod install
+xcodebuild -workspace ./IOSFullScreen.xcworkspace -scheme IOSFullScreen -sdk iphonesimulator


jmagman:
If this isn't usable from the README, we should update that instead of creating scripts that aren't mentioned in the [add-to-app guides](https://flutter.dev/docs/development/add-to-app/ios/project-setup), since this should be an example of how anyone can set up an add-to-app app, based on the docs.

\cc @RedBrogdon

gaaclarke:
This doesn't detract from the README, just provides contextual documentation with a link to the README.

add_to_app/ios_fullscreen/build.sh
```diff
@@ -0,0 +1,16 @@
...
+###############################################################################
+# A convenience script that runs through the setup tasks for building this
+# sample.  A description of the steps can be found at:
+# https://github.com/flutter/samples/blob/master/add_to_app/README.md
+###############################################################################
+
+WHOAMI=`whoami`
+
+cd ../flutter_module
+sed -i .backup 's/dev.flutter.example.flutterModule/dev.flutter.'"$WHOAMI"'.flutterModule/g' pubspec.yaml
```
RedBrogdon, line 12: Thanks for the PR. I'm not sure what the purposes are for some of these steps, though. The string replace on the pubspec, for example -- is that needed to run the app?

gaaclarke: Yep, it's needed to run the app.  The bundle identifier that is in the sample project is registered with the flutter development profile, so you can't run the app unless you change the bundle identifier to something you can register with your credentials (unless you are on the Flutter team).

gaaclarke: Check out apple's sample code, they usually put a random string in the bundle identifier so that there isn't collisions.  (I didn't dig into how they actually do implemented that).

jmagman: I think we should fix that instead of adding this build script.  It should be totally runnable from just the README since the point of these samples is to show how to get add-to-app working.

gaaclarke: Yea, someone should do that, doesn't change the utility of this PR.  Even if someone does that, the script is helpful.  I'm telling you it isn't clear how to run this sample.  The documentation isn't present in a visible location to help and prose isn't as helpful in the same way an executable script is.  If you want to close this PR and file some issues that represent deeper issues that's fine with me.  I think that would be a shame though since it would leave our users at a disadvantage until some undetermined date when someone has the bandwidth to address the deeper issues.

jmagman: I have a PR to fix this the bundle identifier issue at https://github.com/flutter/samples/pull/535.

If this line can be removed, the rest of the script:
1. Runs `flutter pub get` in the module
2. Runs `pod install` in the host app.

That's described here:
https://github.com/flutter/samples/tree/master/add_to_app#the-important-bits
and
https://flutter.dev/docs/development/add-to-app/ios/project-setup#option-a---embed-with-cocoapods-and-the-flutter-sdk

If that's still not clear enough, we should update those docs to make it more obvious.

gaaclarke: Updating the bundle identifiers so they don't collide is good, thanks for looking into that.

It's not that those documents are not clear, they are not **contextual** and they aren't **executable**.  They never say `IOSFullScreen` depends on `flutter_module` and where you can find `flutter-module`, for example.  Also, if you are linked directly to the IOSFullScreen sample, like I was, there so no way of knowing there is relevant information 2 directories higher.  Furthermore, parsing through prose that is generally relevant is not the same as seeing explicit steps relevant to this project that you can execute.

I know there isn't many steps in the script but they are all still required steps in order to successfully run the sample code.

RedBrogdon: You're raising some fair points, and they're much appreciated.

I *do* think that it would be better to avoid adding a build script here as a means of explaining the structure of this sample, for a few reasons:

* I don't think we use build scripts for that purpose anywhere else in the repo.
* We (meaning DevRel) funnel people to this sample via our [samples index](https://flutter.github.io/samples/add_to_app.html), which points at the main README for add-to-app. If we're going to do something to help people understand project structure and build steps, it'd be best to do it there rather than splitting the responsibilty.

Still, like I said, you've pointed out some problems. I suggest we:

1. Work to land @jmagman's change in #535.
2. In this PR, move the commands you've got in the build script (minus the `sed` switch) into the [README](https://github.com/flutter/samples/tree/master/add_to_app) as a tl;dr at the top. That README's gotten really big, and having a section at the top that says "If you're just looking to quickly understand how these work and how to get them running, here's the commands" would help a lot of readers.
3. Outside this PR (and I can do this bit if you like), add small README files to each of the modules and subprojects directing people to the main README file.

What do y'all think? 

CC: @xster

xster: I agree with @gaaclarke's general sentiment, that specific scenarios might be linked to as an example to illustrate something specific like cocoapods integration etc and each subdirectory should have enough documentation to stand on its own, describing which other folder it's paired with and how to run it. 

I don't have a strong opinion either way for scripts. Adding one is fine as long as 
1- We don't make the experience different between a sample and a real add-to-app scenario. i.e. we can never possibly write a script to run the user's pair of host/module projects since we don't know where things are or how to build the host. Therefore we make clear that the script is purely to save you from typing; and 
2- The script doesn't do anything magical. It's just to save you from typing the steps but you need to understand the steps from the docs Jenn listed to do add-to-app (since you have to do them manually in real add-to-app). So the script is just there to document those steps and type them for you. i.e. echo "I'm running the step 1 of the cocoapods integration described at https://... for you now."; echo "cd into the flutter module directory at ... and running flutter pub get to generate the module wrapper for iOS" etc etc.

gaaclarke: @RedBrogdon There are 2 types of people when it comes to "assembly required" toys: people who read over the instructions then build the toy, and people who jump right into it and only check the instructions if they get stuck.

I think a script matches the approach of the latter person, of which I am one.  It might be hard to understand the other person if you aren't one of them.  If you think an appropriately annotated script detracts somehow, I'd disagree but I'm beyond the point where I'll argue the point any further =)

At a minimum we should have a README.md in directories like `flutter_samples/add_to_app/ios_fullscreen`.  The script solved 2 issues: the lack of local documentation in those directories and the lack of concise programmer-friendly documentation.  We should at least solve the locality problem if we don't like the script.

RedBrogdon: Got it. I have a new proposal:

* Work to land @jmagman's change in #535.
* Add a README to each app and Flutter module
  - Each of these will point back to the main README in /add_to_app
  - For the apps, each README will also contain cut-and-paste commands to init any required Flutter modules and run the app from the command line.

I'm happy to do the above, or it can be done in this PR.

RedBrogdon: Created #549.

RedBrogdon:
Closing now that #549 has landed. Thanks for inspiring a good change, @gaaclarke!

PR #535, opened 2020-09-02
Use generic iOS bundle IDs instead of ones reserved by the Flutter Apple Developer account

jmagman:
When you try to build the sample add-to-app module apps with a non-Flutter development team, it will fail and complain that the bundle identifier isn't available (it's registered with the Flutter Apple Developer account).

Change to a generic `com.example` bundle identifier to match what Apple sample apps do (`com.example.apple-samplecode.something`)

### DIFF
--- a/add_to_app/flutter_module/pubspec.yaml
+++ b/add_to_app/flutter_module/pubspec.yaml
@@ -30,4 +30,4 @@ flutter:
   module:
     androidX: true
     androidPackage: dev.flutter.example.flutter_module
-    iosBundleIdentifier: dev.flutter.example.flutterModule
+    iosBundleIdentifier: com.example.flutter-sample.flutterModule
--- a/add_to_app/flutter_module_books/pubspec.yaml
+++ b/add_to_app/flutter_module_books/pubspec.yaml
@@ -30,4 +30,4 @@ flutter:
   module:
     androidX: true
     androidPackage: dev.flutter.example.flutter_module_books
-    iosBundleIdentifier: dev.flutter.example.flutterModuleBooks
+    iosBundleIdentifier: com.example.flutter-sample.flutterModuleBooks
--- a/add_to_app/flutter_module_using_plugin/pubspec.yaml
+++ b/add_to_app/flutter_module_using_plugin/pubspec.yaml
@@ -28,4 +28,4 @@ flutter:
   module:
     androidX: true
     androidPackage: dev.flutter.example.flutter_module_using_plugin
-    iosBundleIdentifier: dev.flutter.example.flutterModuleUsingPlugin
+    iosBundleIdentifier: com.example.flutter-sample.flutterModuleUsingPlugin


add_to_app/flutter_module/pubspec.yaml
```diff
@@ -30,4 +30,4 @@ flutter:
   module:
     androidX: true
     androidPackage: dev.flutter.example.flutter_module
-    iosBundleIdentifier: dev.flutter.example.flutterModule
```
xster, line 33: This PR SGTM. How does this work mechanically? The only app identifiers I see us having created in our account that's close are
dev.flutter.VeggieSeasons, dev.flutter.test.autofill and dev.flutter.test.FidelityTextInput. Does Apple somehow reduce those to the first 2 parts of the rdn and claim that ID space?

xster:
That sounds like it's based on an exact bundle id string collision which would be an issue right. In this case, no one claimed the exact dev.flutter.example.flutterModule string with a distribution provisioning profile that got uploaded to the app store connect (I'm assuming that's the only way to claim an id?) right? 

Though this whole signing thing is complicated and I might be misunderstanding things. 

Is there something special about add-to-app? Since I assume the same bundle ids are reused by the same people who download our other samples.

xster:
i.e. is there something special about this one or should all of these https://cs.opensource.google/search?q=%22dev.flutter%22%20(path:.xcodeproj%20OR%20path:pubspec.yaml)&ss=flutter theoretically be broken as well? If Apple applies some wildcard somewhere in the bundle id path for us.

gaaclarke:
@xster I'm not sure when Apple registers a bundle identifier with a development team.  I know that I got an error when trying to build the add-to-app sample `dev.flutter.example.flutterModule` with my personal account.  I don't think we ever shipped that but it got registered at some point with Apple.

That's how Apple's workaround works, by appending the development team to the bundle identifier there is a guarantee there is no collisions, doesn't matter if it's add-to-app or not.

xster:
Though that sounds like you're saying all of our example code everywhere would have collision issues after the first person runs it in dev unless the bundle id is dynamic. But it doesn't seem like it's happening. I'm curious what we did "wrong" according to the apple signing mechanism here.

gaaclarke:
@xster Nothing special about add-to-app.  This should be done for every sample we ship for iOS.  The fact that users haven't complained speaks to their generosity and savvy =)

jmagman:
I have no idea how it was registered, but I also see the error with my developer account.  Same with the gallery app.

Apple's solution is to append `DEVELOPMENT_TEAM` to the bundle identifier, but unfortunately for add-to-app it's set in the pubspec.yaml and don't have access to those build settings.

With non-add-to-app samples, like the gallery, we do have control of the build settings, so we could change `PRODUCT_BUNDLE_IDENTIFIER` there if we wanted.

add_to_app/flutter_module/pubspec.yaml
jmagman: I'm really not sure how it works, I was surprised too.

jmagman:
> This will only work for the first person. 

I don't think so, `com.example` seems special (reserved in RFC 5735).  Flutter uses `com.example.` in the create template, I have never seen someone complain about collisions there.

RedBrogdon:
Bear with me; I'm trying to keep up with a deep, long-running conversation that doesn't relate to my particular expertise. :)

I've got some questions for y'all:

* Do we still need to update the other apps in the repo, or is it just the add-to-app samples that need fixing?
* Does this matter for debug builds, or just release ones?
* How is it that Travis is able to build these apps? Is it due to the unusual code signing options we have in place there?
  - https://github.com/flutter/samples/blob/master/tool/travis_ios_script.sh

gaaclarke:
@jmagman Maybe, it depends on how Apple implemented their provisioning.  I'm not sure when a app bundle is registered with Apple so I can't verify it.  Have you figured out how to verify it?

@RedBrogdon 
1) We should be fixing every xcodeproj that has sample code.
2) It matters for debug and release builds if you are putting them on a device.
3) Travis just compiles things, it doesn't actually install the code onto a device.

xster:
I have some degree of unconfirmed suspicion this has something to do with the latest Xcode. I've been getting 

`error: Failed to register bundle identifier: The app identifier "io.flutter.demo.gallery" cannot be registered to your development team because it is not available. Change your bundle identifier to a unique
    string to try again. (in target 'Runner' from project 'Runner')`

on the flutter gallery too which never happened before.

RedBrogdon:
Discussed offline. Mobile platform team will do some investigating to make sure we understand the exact nature of the problem and let us know.

It looks likely, though, that we may need to rename the bundle IDs in all of our samples so that they incorporate the development team identifier into the final Bundle ID.

jmagman:
Closing this, from our testing there's evidently nothing special about `com.example`

PR #538, opened 2020-09-09
Add an iOS app clip sample

xster:
Don't know if I've crossed the threshold of acceptable scope 😅

I've made a sample to show using an App Clip target in iOS 14. 

https://github.com/flutter/flutter/issues/60029


ios_app_clip/.metadata
```diff
@@ -0,0 +1,10 @@
+# This file tracks properties of this Flutter project.
+# Used by Flutter tool to assess capabilities and perform upgrades etc.
+#
+# This file should be version controlled and should not be manually edited.
+
+version:
+  revision: 7e480edc7e531c400aff4f08afeebe04c0f63e2e
+  channel: cupertino-icon-5
```
RedBrogdon, line 8: Do we have new channels?

ios_app_clip/README.md
```diff
@@ -0,0 +1,17 @@
+# ios_app_clip
```
RedBrogdon, line 1: Can you flesh this out just a bit using the format from the other READMEs? Nothing huge, just the same headers and footer, really:

https://github.com/flutter/samples/tree/master/jsonexample

RedBrogdon (commented):
I'm not sure if this repo is the right place for this. I get that the goal for the sample is that it's set up to build out of the box as an AppClip, but what's the content? Like, if you diffed this against the Counter app from `flutter create`, what's going to be different?

The reason I ask is that if it's not that much code, you might be better off just sticking with the guide on flutter.dev by itself.

ios_app_clip/.metadata
xster: doh, fixed it to what it should have been.

ios_app_clip/README.md
xster: ok

xster:
Ya, it's definitely debatable. The goal is to have an out of the box thing that builds like you're saying so we can test. Also a known working case that users can validate against after following through the guide and perhaps running into problems. It's a way to bisect the problem. 

The difference with a blank flutter created template is going to be mostly in Runner.xcodeproj file. It's a bit like the prebuilt iOS host app example in add-to-app. There isn't too much code. It's A) to let you spin up an add-to-app example quickly to see it running, and to prove that it works, and B) to double check your own integration steps (which is also manual on flutter.dev). In that case too, a lot of the diff vs a blank Xcode project is in the Xcode project hook-up itself.

xster:
Gentle ping

ios_app_clip/lib/main.dart
```diff
@@ -0,0 +1,132 @@
+import 'package:device_info/device_info.dart';
```
RedBrogdon, line 1: Copyright header needed.

ios_app_clip/lib/main.dart
```diff
@@ -0,0 +1,132 @@
...
+import 'package:flutter/material.dart';
+
+void main() {
+  runApp(MyApp());
+}
+
+class MyApp extends StatelessWidget {
+  // This widget is the root of your application.
```
RedBrogdon, line 9: I would recommend cutting all of these comments rom the template. Also, consider stripping this down to just the plugin output and some kind of graphic (like a spinning Flutter logo or something).

ios_app_clip/pubspec.yaml
```diff
@@ -0,0 +1,77 @@
+name: ios_app_clip
```
RedBrogdon, line 1: Cut these comments.

ios_app_clip/test/widget_test.dart
```diff
@@ -0,0 +1,30 @@
...
+// gestures. You can also use WidgetTester to find child widgets in the widget
+// tree, read text, and verify that the values of widget properties are correct.
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:ios_app_clip/main.dart';
+
+void main() {
+  testWidgets('Counter increments smoke test', (WidgetTester tester) async {
```
RedBrogdon, line 14: If you revise the main page of the app, feel free to replace this with a stub test. It does need to have one, though, and get listed in `/tool/travis_flutter_script.sh`

RedBrogdon: Actually, consider adding it to `/tool/travis_ios_script.sh` as well.

ios_app_clip/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,871 @@
...
+				GCC_C_LANGUAGE_STANDARD = gnu11;
+				INFOPLIST_FILE = AppClip/Info.plist;
+				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/Frameworks",
+				);
+				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
+				MTL_FAST_MATH = YES;
+				PRODUCT_BUNDLE_IDENTIFIER = com.example.iosAppClip.Clip;
```
RedBrogdon, line 502: Should we do the ${DEVELOPMENT_TEAM} trick for this project as well?

RedBrogdon: And is this the right place, or should it be done in `AppFrameworkInfo.plist`? I don't actually know, and I need to update the other samples.

ios_app_clip/.metadata
```diff
@@ -0,0 +1,10 @@
+# This file tracks properties of this Flutter project.
+# Used by Flutter tool to assess capabilities and perform upgrades etc.
+#
+# This file should be version controlled and should not be manually edited.
+
+version:
+  revision: 65ae30c8cc99b7d48b5bd08d3726f5669b819be0
+  channel: master
```
RedBrogdon, line 8: If this is targeting master (and I'm guessing it should), it should go in `/experimental` rather than the root directory.

RedBrogdon (changes requested):
Sounds good to land it here. I've added a few nitpicks, but nothing significant. 😄

ios_app_clip/ios/Runner.xcodeproj/project.pbxproj
RedBrogdon: I now realize it's probably weird to use your PR to work out my own research, but there it is. 😄

ios_app_clip/lib/main.dart
xster: Done

ios_app_clip/lib/main.dart
xster: Ah ok, SG. Done

ios_app_clip/pubspec.yaml
xster: done

ios_app_clip/test/widget_test.dart
xster: ok, I think I did. I'm not sure I understood the travis_ios_script. Did you mean whether I had an XCTest for this?

ios_app_clip/ios/Runner.xcodeproj/project.pbxproj
xster: Ah very good point. Done. PRODUCT_BUNDLE_IDENTIFIER is the thing to use, rather than CFBundleIdentifier.

ios_app_clip/.metadata
xster: Oh good point. This will work fine on 1.20 actually. I just happened to have made this on master.

xster:
Gentle ping

ios_app_clip/test/widget_test.dart
RedBrogdon: No, I just want it to build. We have some full integration tests for the add-to-app Android apps, but haven't had bandwidth to do anything for iOS.

RedBrogdon (approved):

RedBrogdon:
LGTM. Adding it to the iOS test script felt like a thing to do because this app mostly exists to illustrate iOS configuration and iOS-related stuff. It's not completely necessary though. I'll have a think on it. Comments welcome in the meantime.

PR #541, opened 2020-09-12
Replaced Long Press to icon button

mohak852:
#540 
Solved the issue by replacing Long Press to Icon button

PR #545, opened 2020-09-15
[Animations] Add remove-button

### LINKED ISSUE #540: [animations] AnimatedList demo should have a delete IconButton to delete the item
## Description

Currently, the list item is deleted upon the long-press which is not the general pattern people use to delete the item. It's like finding an easter egg in the demo 😄 

## Suggestion

Have a delete IconButton which will delete the item instead of a long press.

Not submitting a PR and keeping this open for the first time contributors since it's an easy fix. If we can have an easy fix label it will be easily noticeable for them.

mohak852:
fixes #540 
#541 I was unable to get #541 working so I created a new pr.

@AyushBherwani1998 @abd99 @RedBrogdon

### DIFF
--- a/animations/lib/src/misc/animated_list.dart
+++ b/animations/lib/src/misc/animated_list.dart
@@ -57,7 +57,10 @@ class _AnimatedListDemoState extends State<AnimatedListDemo> {
       leading: CircleAvatar(
         child: Icon(Icons.person),
       ),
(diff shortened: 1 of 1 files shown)


abd99 (approved):
LGTM, thanks!

RedBrogdon (approved):

PR #549, opened 2020-09-18
Updates add-to-app README strategy

RedBrogdon:
* Adds a bash script to the top of the main README with commands for prepping all the modules and getting them ready to run.
* Adds mini READMEs to all the apps that didn't have them, briefly explaining what they're for and pointing people back to the main README
* Does a proof-read run on the main README to correct various errors/typos/malaprops that have crept in over the past few quarters.


gaaclarke (approved):
Definitely an improvement, thanks!

PR #552, opened 2020-09-24
[provider_counter] Linux added in provider_counter

yash1200:
## Description

Linux folder added in provider_counter example, which makes it able to run on linux platform.

### DIFF
--- a/provider_counter/linux/.gitignore
+++ b/provider_counter/linux/.gitignore
@@ -1,0 +1,1 @@
+flutter/ephemeral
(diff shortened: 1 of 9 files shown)


AyushBherwani1998:
Hey @yash1200, thanks for the contribution. The samples in the repo are focused on the stable channel, that's why most of the samples are missing other platforms. You can check the [contributing guide](https://github.com/flutter/samples/blob/master/CONTRIBUTING.md#a-few-ground-rules).

yash1200:
Oh didn't see that. Thanks for the info.

PR #555, opened 2020-09-26
[infinite_list] Web folder added in infinite_list

yash1200:
In this PR, I have added the web folder in infinite_list.


domesticmouse:
@RedBrogdon do we have a position on adding web directory to stable samples? Looking at this PR, I see we are winding up with `pubspec.lock` with Dart 2.10 level deps. I'm also concerned that some of the travis tests are failing due to Android licenses not being accepted?

RedBrogdon:
Several of our existing stable samples have web (and desktop) folders. It looks like you were involved in creating them. 😄 

https://github.com/flutter/samples/pull/223/files

I don't think it's a problem (nor is the failed test, which seems to be corrected), but we *do* need to keep the pubspec locked to the stable version. @yash1200, can you run `flutter channel stable`, then `flutter pub get` and update your pull request?

yash1200:
It's done @RedBrogdon .

RedBrogdon (approved):
LGTM. Thanks!

PR #557, opened 2020-09-28
Added example plugin for ios and android that uses pigeon with kotlin / swift.

gaaclarke:


gaaclarke:
Moved to https://github.com/gaaclarke/pigeon_plugin_example

PR #563, opened 2020-10-10
[platform_channels] adds ios implementation for MethodChannel demo

AyushBherwani1998:
## Description
Adds iOS implementation for MethodChannel Demo.

cc:// @RedBrogdon let me know if other file changes I pushed are meant to be pushed or not. I fairly new to iOS, so let me know if my swift implementation needs to be improved.


AyushBherwani1998:
googlebot I fixed it.

platform_channels/ios/Runner/AppDelegate.swift
```diff
@@ -3,11 +3,31 @@ import Flutter
...
+        _ application: UIApplication,
+        didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
+    ) -> Bool {
+        let flutterViewController = window.rootViewController as! FlutterViewController
+        FlutterMethodChannel(name: "methodChannelDemo", binaryMessenger: flutterViewController.binaryMessenger).setMethodCallHandler({
+            (call: FlutterMethodCall, result: FlutterResult) -> Void in
+            
+            let count = (call.arguments as? NSDictionary)?["count"] as? Int
+            
+            if count == nil {
```
domesticmouse, line 16: This probably should use `guard let`.

domesticmouse (approved):

platform_channels/ios/Runner/AppDelegate.swift
AyushBherwani1998: Something like this

```swift
guard let count = (call.arguments as? NSDictionary)?["count"] as? Int else {
    // Throw invalid argument error
}
```

AyushBherwani1998: Done @domesticmouse

platform_channels/ios/Runner/AppDelegate.swift
```diff
@@ -3,11 +3,30 @@ import Flutter
...
+        FlutterMethodChannel(name: "methodChannelDemo", binaryMessenger: flutterViewController.binaryMessenger).setMethodCallHandler({
+            (call: FlutterMethodCall, result: FlutterResult) -> Void in
+            
+            guard let count = (call.arguments as? NSDictionary)?["count"] as? Int else {
+                result(FlutterError(code: "INVALID_ARGUMENT", message: "Value of count cannot be null", details: nil))
+                return
+            }
+            
+            switch call.method {
+            case "increment":
```
RedBrogdon, line 20: I did a double-take at the indentation here, but it's correct. 😄

RedBrogdon (approved):
LGTM. Thanks for putting that new laptop to work for Flutter. 😄

Issue #564, opened 2020-10-15
[Animations] Add title property (AppBar) for All the Routes.


Raj-Dave368:
### Description

There is no title on some Route Page which can show, What type of Animation Page is Demonstrating.
![image](https://user-images.githubusercontent.com/54175432/96062870-5e10e600-0eb4-11eb-99c2-ad971c28941f.png)

* When I was testing the app I was Just clicking `ListTile`(main.dart) Randomly,

![image](https://user-images.githubusercontent.com/54175432/96061699-84815200-0eb1-11eb-8bc9-36fe68e11d5d.png)

* And I was navigating to Corresponding Page But After Navigating to the Screen  **I didn't know which type of Animation I am Trying** (Because: Page Doesn't Contain title On The `AppBar`).

![image](https://user-images.githubusercontent.com/54175432/96062339-0f168100-0eb3-11eb-9160-fa3a1d8ba2d4.png)

* Before Navigating to Any Page, I had to first Remember the `ListTile.title`. [a few pages have a title, but I'll like to have for all]

![image](https://user-images.githubusercontent.com/54175432/96062171-ad561700-0eb2-11eb-92c4-4a7658f5844b.png)

But I would Like to Make it More User Friendly and Easy to Use.
Something like this:

![image](https://user-images.githubusercontent.com/54175432/96062431-484ef100-0eb3-11eb-98f2-813ddb57fb27.png)

This is my first issue and I would like to Solve this Issue.

Raj-Dave368:
@RedBrogdon @AyushBherwani1998 Please Review my issue, if the corresponding issue does not exist already, then I would like to Make a PR Because I have done this minor Changes, which have Huge Impact(80/20 principle).

Thanks,
                                                                                                                                                                         - Your new contributor.

AyushBherwani1998:
Hey @Raj-Dave368, thanks for opening the issue. IMO, it will make the UX better. We still have to wait for Andrew's take on this. Till then, you can make changes in your system, and if this is something the samples need, you can submit a PR.

RedBrogdon:
I agree, and I imagine @johnpryan would as well (though he's free to disagree with me 😄 ).

Alabhya268:
@Raj-Dave368 are you still working on this issue? If not I would like to work on this issue.

Alabhya268:
@RedBrogdon , I want to work on this issue could you please assign it to me ?

adarsh-technocrat:
[...]

PR #567, opened 2020-10-16
[Animations] Adding the title property in AppBar for each Page

Raj-Dave368:
##Description 

Fixes: #564

I have added `title `property of `AppBar` for each Route Page, Which Represents the Animation Type.


Raj-Dave368:
@googlebot I fixed it.

AyushBherwani1998:
Your commits are not signed, try signing the commits. Make sure your email and username in the git config are same.

Raj-Dave368:
Yes, that was an issue, At that time My email was different in GitHub and Git config. 
Any other options, rather than closing this PR and create new PR with the correct email?

AyushBherwani1998:
Usually you can use `--amend` with `git commit` to change the sign, commit message and other things of the most recent commit. Since you have multiple commits, I'm not sure of a convenient way. As far as I remember there's also -c flag to specify a commit to amend but I'm not able to find it in docs. I did find [something on SO](https://stackoverflow.com/questions/41882919/is-there-a-way-to-gpg-sign-all-previous-commits), but I'm not used those before so I'll leave it up-to you. If I'll get the convenient way, I'll share it.

Raj-Dave368:
@AyushBherwani1998 Sir I have found the following solution on StackOverflow, can you please, just check it before I run it.

```
git filter-branch --env-filter '

OLD_EMAIL="raj@OMNS.com"
CORRECT_NAME="Raj-Dave368"
CORRECT_EMAIL="rajdaveomns1@gmail.com"

if [ "$GIT_COMMITTER_EMAIL" = "$OLD_EMAIL" ]
then
    export GIT_COMMITTER_NAME="$CORRECT_NAME"
    export GIT_COMMITTER_EMAIL="$CORRECT_EMAIL"
fi
if [ "$GIT_AUTHOR_EMAIL" = "$OLD_EMAIL" ]
then
    export GIT_AUTHOR_NAME="$CORRECT_NAME"
    export GIT_AUTHOR_EMAIL="$CORRECT_EMAIL"
fi
' --tag-name-filter cat -- --branches --tags
```

animations/lib/src/basics/05_animated_builder.dart
```diff
@@ -15,7 +15,7 @@ class _AnimatedBuilderDemoState extends State<AnimatedBuilderDemo>
     with SingleTickerProviderStateMixin {
   static const Color beginColor = Colors.deepPurple;
   static const Color endColor = Colors.deepOrange;
-  Duration duration = Duration(milliseconds: 800);
+  Duration duration = Duration(milliseconds: 1000);
```
RedBrogdon, line 18: Any particular reason for this change? It doesn't seem to relate to the title changes.

RedBrogdon (commented):
Thanks for sending a PR.

This looks good to me, though I do have one question about the duration change.

RedBrogdon:
As @AyushBherwani1998 was nice enough to point out, the commit author emails are a problem we need to fix before merging the PR. 

Because the code changes here are fairly small, the easiest thing to do might be just to undo these commits, remake the changes in a new commit, and force-push that to your fork. If you'd prefer just to close this PR and open a new one, though, that works, too.

animations/lib/src/basics/05_animated_builder.dart
Raj-Dave368: Thanks for reviewing PR.
Yes, it is not related to title, But because of long duration user can see more clearly that how colors are changing. 
Not a big change but a bit slow animation creates huge impact.

Raj-Dave368:
Gotcha. 
Fine I'll close this PR but before that, I would like to wait for your opinion on duration change.
Is it fine to change duration from 800 to 1000?

Issue #568, opened 2020-10-16
[Improvement] Size of Material Button is small


Raj-Dave368:
### Description 
if the size of the `MaterialButton` in `AnimatedBuilder` [here ](https://github.com/flutter/samples/blob/master/animations/lib/src/basics/05_animated_builder.dart#L49 )Could be large, then the user could more clearly "How beautifully Flutter Turns One Color into Another".

### **instead of this**

![image](https://user-images.githubusercontent.com/54175432/96218543-91ca3980-0fa2-11eb-8b4a-b0c546ab824c.png)

### **it can look like this.**

![image](https://user-images.githubusercontent.com/54175432/96218673-dfdf3d00-0fa2-11eb-87ca-549d57896c22.png)  
![image](https://user-images.githubusercontent.com/54175432/96218688-e79ee180-0fa2-11eb-8404-679410493e3e.png)

RedBrogdon:
Thanks for filing this issue, @Raj-Dave368. I actually wouldn't recommend this change, mostly because having a single button expanded to take up the whole isn't something you'd normally see in an app (nor is it something the Material team would recommend).

Making the button a *little* bigger, though, could be a good change. Doubling it in size, for example.

Raj-Dave368:
Thanks 
Sorry, As I was excited to contribute as a newbie I haven't thought about material design.
But I think issue is still there that it will be better to change colour of large area instead of colour of button.
So what if instead of applying color on button, we apply color on scaffold(background color) and keeping button Size normal.

tusharojha:
@domesticmouse I think this issue can be closed now.

AyushBherwani1998:
Closing, the issue have been fixed in #600

PR #569, opened 2020-10-16
[Animations] Increasing Meterial Button Size And changing beginColor

Raj-Dave368:
# Description

Fixes: #568 

This issue(#568) is made by me because During testing I was trying different colors to see how beautifully flutter changes colors from one to another but I forced to increase the `MaterialButton `size to see Tween Animation more clearly, And this may happen **with other persons as well**. 

I have increased the size Of `MaterialButton `and Gave `Padding `around it and Also have Changed the Color.
also, I have increased `FontSize `of `Text `.

and now it looks like this,

image

Previously, 

image

### DIFF
--- a/animations/lib/src/basics/05_animated_builder.dart
+++ b/animations/lib/src/basics/05_animated_builder.dart
@@ -13,7 +13,7 @@ class AnimatedBuilderDemo extends StatefulWidget {
 
 class _AnimatedBuilderDemoState extends State<AnimatedBuilderDemo>
     with SingleTickerProviderStateMixin {
-  static const Color beginColor = Colors.deepPurple;
+  static const Color beginColor = Colors.indigoAccent;
   static const Color endColor = Colors.deepOrange;
   Duration duration = Duration(milliseconds: 800);
   AnimationController controller;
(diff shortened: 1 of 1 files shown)


animations/lib/src/basics/05_animated_builder.dart
```diff
@@ -46,16 +46,21 @@ class _AnimatedBuilderDemoState extends State<AnimatedBuilderDemo>
...
-              color: animation.value,
-              child: child,
-              onPressed: () {
-                if (controller.status == AnimationStatus.completed) {
-                  controller.reverse();
-                } else {
-                  controller.forward();
-                }
-              },
+            return Container(
```
RedBrogdon, line 49: As I mentioned in the related issue, a full-screen button isn't a UX pattern we'd want to show here. Increasing the height, width, and font size to make the button slightly bigger could work, however.

animations/lib/src/basics/05_animated_builder.dart
```diff
@@ -13,7 +13,7 @@ class AnimatedBuilderDemo extends StatefulWidget {
 
 class _AnimatedBuilderDemoState extends State<AnimatedBuilderDemo>
     with SingleTickerProviderStateMixin {
-  static const Color beginColor = Colors.deepPurple;
+  static const Color beginColor = Colors.indigoAccent;
```
RedBrogdon, line 16: Any particular rationale for changing the color here?

RedBrogdon (changes requested):

animations/lib/src/basics/05_animated_builder.dart
Raj-Dave368: No @RedBrogdon . It was just same colour as a AppBar color so I thought to apply new color!
If it's not good then We can undo the changes!
What do you think?

RedBrogdon: It's not a critical thing -- I just wanted to know the rationale. If anything, indigo is more of an opposite to orange, so this change seems fine to me:

https://www.google.com/search?q=color+wheel+indigo

RedBrogdon:
@Raj-Dave368, just checking in -- is this PR something you're still interested in working on?

Raj-Dave368:
> @Raj-Dave368, just checking in -- is this PR something you're still interested in working on?

No sir, actually I am really busy, which I don't wanted to be, but now I can't work on this issue anymore. You can close it, if you want.
Thank you and sorry again.

RedBrogdon:
Understood. Thanks for getting back to us.

Issue #278, opened 2019-04-03
[Veggieseasons] Let the search page feed slide behind the search/tab bar


xster:

RedBrogdon:
Notes:

Right now the list on the search page is clipped by the bottom edge of the search field. Instead, it should slide underneath the list.

AyushBherwani1998:
I would like to work on this issue @RedBrogdon 
We can have a stack where we can place the search bar above the list view. Is there any alternate solutions that can be tried to achieve this? I'll be happy to get some leads.

xster:
I think this got fixed recently?

RedBrogdon:
#279, which was the status bar-related scroll, got fixed. This is still outstanding.

tusharojha:
@RedBrogdon One possible way could be to wrap the items inside a column and provide the singlechildscrollview at the very top of both search bar and feeds.

Can I do this way or would you like to suggest something?

RedBrogdon:
I think this would need something slightly more sophisticated, unless I'm misunderstanding you.

The goal here would be for the search bar to remain in place at the top of the page, and for the search results to slide underneath it when scrolled up.

tusharojha:
I would like to work on this issue!
@RedBrogdon I have tried to use stack instead of column inside the `search.dart` (output as shown below). Although I don't think its the exact result but would like to know your thoughts on the same.

![ezgif com-video-to-gif (3)](https://user-images.githubusercontent.com/36601970/93668224-c680c900-faa8-11ea-9460-3083cc4d10d5.gif)

tusharojha:
@RedBrogdon I would like to know your suggestions for its implementation. Please have a look into it.

kjain333:
There seems to be a simple bug in screen recording you have shared that the first item of the list is already covered by the search bar may increasing the size of the listview by one and adding a SizedBox at the starting can fix this issue.

kjain333:
I request @xster and @RedBrogdon to kindly review the pull request I have created recently

RedBrogdon:
This has been fixed in #572.

PR #572, opened 2020-10-22
Fixed issue #278 by making list scrollable under search bar

kjain333:
A small contribution to fix issue #278 making the list slide under the search bar. There is some quality problem in the uploaded file and no colors have been changed when compared to the previous commits.
veggie_season_fix


veggieseasons/lib/screens/search.dart
```diff
@@ -79,12 +85,16 @@ class _SearchScreenState extends State<SearchScreen> {
       builder: (context) {
         return SafeArea(
           bottom: false,
-          child: Column(
+          child: Stack(
             children: [
-              _createSearchBox(),
-              Expanded(
-                child: _buildSearchResults(model.searchVeggies(terms)),
+              Column(
```
RedBrogdon, line 90: Is this Column still necessary, now that it only has a single child? Could you instead use Positioned instead of Column/Expanded together?

RedBrogdon (changes requested):
One minor change requested. @LongCatIsLooong, do you mind taking a look at this as well, to make sure it's the correct approach?

veggieseasons/lib/screens/search.dart
```diff
@@ -61,12 +61,18 @@ class _SearchScreenState extends State<SearchScreen> {
...
 
     return ListView.builder(
-      itemCount: veggies.length,
+      itemCount: veggies.length + 1,
       itemBuilder: (context, i) {
-        return Padding(
-          padding: EdgeInsets.only(left: 16, right: 16, bottom: 24),
-          child: VeggieHeadline(veggies[i]),
-        );
+        if (i == 0) {
```
LongCatIsLooong, line 66: I'm a bit concerned about hardcoding the height (64) the search bar occupies here. Is it possible to use a [SliverPersistentHeader](https://master-api.flutter.dev/flutter/widgets/SliverPersistentHeaderDelegate-class.html) to achieve this?

LongCatIsLooong: Doesn't seem easy to do with `SliverPersistentHeader` since it currently requires us to specify `minExtent` and `maxExtent`. /cc @Piinks

veggieseasons/lib/screens/search.dart
kjain333: Respected Sir,
It's good to know that you responded so quick on my pr. Actually I was a bit unsure as to why the team was using Expanded in the previous code as similar thing can also be achieved without using the same. I had to use Column as well because documentation says Expanded should be a direct child of a Column or Row. Nevertheless I will me more than happy to make this change so that it can be merged with the current project.

veggieseasons/lib/screens/search.dart
kjain333: Respected sir,
I am just a budding developer and am glad that I can ask for your guidance at this step. Sir my thinking was since other things used to create the search bar like paddings,margins have been hard coded and default size of icon which remains same across platforms has been used wouldn't the size of searchbar automatically get fixed and if the height is fixed then hard coding this value should not be a problem as if it works fine in one device it will work similarly in others as well.

LongCatIsLooong: hi kjain333, indeed the `SearchBar` has a fixed intrinsic height, but the height value may change over time. Other contributors may adjust the paddings/margins of the `SearchBar` widget, without realizing that could break the UI of the search page. IMO we should promote a more robust approach.

There doesn't seem to be a readily available widget that does exactly what we want, unfortunately. The use case here doesn't seem to be niche so maybe we should introduce a new `Sliver` widget in flutter?

kjain333: Respected Sir,
I completely agree with the comment you have made and accept my mistake for not actually thinking about the same. It would be really nice if we would create a new Sliver widget for the same and hopefully I would be able to contribute to that as well. Till then I will also try to think of a work around which can hopefully solve this issue.

kjain333: Respected sir,
Sorry to bother you again. I understand that you are trying to build a robust approach towards solving this problem but using the tools we presently have will it be a good approach to replace the SizedBox with the SearchBox itself but with opacity 0 so that it occupies the same space and wouldn't affect the design as well?

veggieseasons/lib/screens/search.dart
```diff
@@ -79,12 +86,10 @@ class _SearchScreenState extends State<SearchScreen> {
       builder: (context) {
         return SafeArea(
           bottom: false,
-          child: Column(
+          child: Stack(
```
Piinks, line 89: Making this a Stack is the right change here. 👍   I think re-factoring for a sliver solution would overcomplicate things. This is a very simple use case that shouldn't require slivers.
The Expanded widget would also no longer be necessary since the ListView will take up all of the available space.

veggieseasons/lib/screens/search.dart
Piinks: Could we just remove the SizedBox? It does not appear to be necessary.

If we wrap the _createSearchBox in a`Visibility` widget, that will achieve the invisible placeholder, with the added benefit that we know it won't be interacted with.

```dart
return ListView.builder(
      itemCount: veggies.length,
      itemBuilder: (context, i) {
        if (i == 0) {
          return Visibility(
            child: _createSearchBox(),
            visible: false,
            maintainSize: true,
            maintainAnimation: true,
            maintainState: true,
          );
        }
        return Padding(
          padding: EdgeInsets.only(left: 16, right: 16, bottom: 24),
          child: VeggieHeadline(veggies[i]),
        );
      },
    );
```

I've run this, and it appears to WAI. :)

Piinks (commented):
Hi @kjain333 welcome! Always happy to work with a new contributor!

I think we can take a different approach here to address some of the concerns @LongCatIsLooong  mentioned.

The scroll effect you've created is really cool! I would recommend a slight aesthetic update though, for a native feel and good readability. Since the scrollbar already has an opacity effect on it, its a bit hard to see clearly.

This will update the search bar to have the same effect as the bottom navigation bar, where you can see the items scrolling below, but with a slight blur effect.
```dart
Widget _createSearchBox() {
    return ClipRect(
      child: Container(
        color: CupertinoTheme.of(context).barBackgroundColor,
        child: BackdropFilter(
          filter: ImageFilter.blur(sigmaX: 10.0, sigmaY: 10.0),
          child:Padding(
            padding: const EdgeInsets.all(8),
            child: SearchBar(
              controller: controller,
              focusNode: focusNode,
            ),
          ),
        ),
      ),
    );
  }
```
Try this out and let me know what you think, experiment with it if you like. Thank you for your contribution to this sample app! 🎉

veggieseasons/lib/screens/search.dart
kjain333: Respected Ma'am,
It's really good to know that we are on the same page with the change of things. With the latest change I have made to my request I have used the Opacity widget which indeed makes the search bar invisible but the part I missed was that it should not even be click able even though very less people would be able to get it that there is a search bar there as well. 
Thanks a lot for pointing this out and I will make these changes right away.

LongCatIsLooong: That's clever! I'm on board with this approach. But still I think it's a little bit hacky, and this is a common enough use case that `SliverPersistentHeader` should probably support. Maybe we can retrofit this when we have the right sliver.

kjain333:
@Piinks Ma'am the feature you are referring to about blurring the widgets scrolled behind the search bar seems to be really cool but I am bit unsure about the approach you are suggesting as what it does is that it blurs the content in a rectangular container including the paddings and the search bar which I am sure must not be desired product. Maybe wrapping the search bar instead of Padding might bring the correct results.
Would be great to hear from you soon on this.

veggieseasons/lib/screens/search.dart
Piinks: True! I don't know that we are missing a sliver here. It can definitely be done with a SliverPersistentHeader, a SliverAppBar most likely, it would just mean that we would need to convert the ListView to a SliverList with a CustomScrollView, etc. All that seems a bit more than necessary.

kjain333:
Adding more to my previous comment:
1) a sample of the ui as per changes requested by @Piinks 
Screenshot_20201024-005320
2) a sample of the hi as per my suggestions
Screenshot_20201024-011213
I would be delighted if the team could provide there suggestions which ui should be used and recommend any changes to be made if required.

Piinks:
> @Piinks Ma'am the feature you are referring to about blurring the widgets scrolled behind the search bar seems to be really cool but I am bit unsure about the approach you are suggesting as what it does is that it blurs the content in a rectangular container including the paddings and the search bar which I am sure must not be desired product. Maybe wrapping the search bar instead of Padding might bring the correct results.
> Would be great to hear from you soon on this.

Hey @kjain333 the effect was intended in the snippet I provided. Feel free to modify it if you prefer and update your PR description above. I think @RedBrogdon has the final say on the appearance. :)

veggieseasons/lib/screens/search.dart
```diff
@@ -79,12 +89,10 @@ class _SearchScreenState extends State<SearchScreen> {
...
             children: [
+              _buildSearchResults(model.searchVeggies(terms)),
               _createSearchBox(),
-              Expanded(
-                child: _buildSearchResults(model.searchVeggies(terms)),
-              ),
```
kjain333, line 87: Removed the Expanded Widget as the ListView itself handles taking up all the space provided.

veggieseasons/lib/widgets/search_bar.dart
```diff
@@ -19,45 +21,51 @@ class SearchBar extends StatelessWidget {
...
-            ),
-            Expanded(
-              child: CupertinoTextField(
-                controller: controller,
-                focusNode: focusNode,
-                decoration: null,
-                style: Styles.searchText(themeData),
-                cursorColor: Styles.searchCursorColor,
-              ),
+    return ClipRRect(
```
kjain333, line 24: Improvement in the ui design so that scrolled list is blurred under the search bar and not visible due to it's opacity.

veggieseasons/lib/screens/search.dart
RedBrogdon: It seems like this can't be solved elegantly without creating some kind of render object or other class that would sit above both the ListView and the search box, know how big the latter is, and pass that information down to whatever's building the ListView.

Therefore it's cool by me to merge this as-is, though I'd recommend adding a one line comment above line 68 that explains why the code is building a second, invisible search box.

RedBrogdon:
@kjain333, my only remaining request here is for an inline comment on the second search box. Looks great otherwise!

kjain333:
Sure sir, I will do that right away.

RedBrogdon (approved):

RedBrogdon:
The only test left running is an Android one, which this PR shouldn't be able to break.

RedBrogdon:
Thanks for the pull request, @kjain333. It's a good change. 👍

PR #579, opened 2020-11-01
[federated_plugin] add iOS implementation

AyushBherwani1998:
## Description

Adds iOS implementation for federated_plugin

cc:// @RedBrogdon @domesticmouse  I don't have Apple Developer account to test it on my iPhone, and on simulator it throws exception. I read that you can get battery level of simulator on reddit. It would be great if you can check it when ever you get time.

### DIFF
--- a/experimental/federated_plugin/federated_plugin/ios/Classes/SwiftFederatedPlugin.swift
+++ b/experimental/federated_plugin/federated_plugin/ios/Classes/SwiftFederatedPlugin.swift
@@ -3,12 +3,24 @@ import UIKit
 
 public class SwiftFederatedPlugin: NSObject, FlutterPlugin {
   public static func register(with registrar: FlutterPluginRegistrar) {
-    let channel = FlutterMethodChannel(name: "federated_plugin", binaryMessenger: registrar.messenger())
+    let channel = FlutterMethodChannel(name: "battery", binaryMessenger: registrar.messenger())
     let instance = SwiftFederatedPlugin()
     registrar.addMethodCallDelegate(instance, channel: channel)
   }
 
   public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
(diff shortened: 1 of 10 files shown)


experimental/federated_plugin/federated_plugin/ios/Classes/SwiftFederatedPlugin.swift
```diff
@@ -3,12 +3,24 @@ import UIKit
...
   public static func register(with registrar: FlutterPluginRegistrar) {
-    let channel = FlutterMethodChannel(name: "federated_plugin", binaryMessenger: registrar.messenger())
+    let channel = FlutterMethodChannel(name: "battery", binaryMessenger: registrar.messenger())
     let instance = SwiftFederatedPlugin()
     registrar.addMethodCallDelegate(instance, channel: channel)
   }
 
   public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
-    result("iOS " + UIDevice.current.systemVersion)
+    guard  call.method == "getBatteryLevel"  else {
```
RedBrogdon, line 12: nit: it looks like there are some extra whitespace characters here. Is that intentional?

RedBrogdon (approved):
LGTM, though I also lack a hardware iOS device to test with.

jmagman:
Tried on a real device, it works (`device.batteryState` returns `full`)
Screen Shot 2020-11-02 at 2 10 54 PM

RedBrogdon:
Much obliged, @jmagman!

experimental/federated_plugin/federated_plugin/ios/Classes/SwiftFederatedPlugin.swift
AyushBherwani1998: Opps! I'll fix this in the next commit for MacOS support.

PR #588, opened 2020-11-07
Re-generate Linux, macOS and Windows runners

domesticmouse:


domesticmouse:
PTAL @johnpryan, this is to update the runners as part of pushing to an app store.

johnpryan (approved):

Issue #586, opened 2020-11-06
experimental/desktop_photo_search doesn't handle utf8 correctly.


domesticmouse:
@mit-mit who should I chat to about `built_value` and decoding utf8 data correctly? I'm guessing I need to use [`characters`](https://pub.dev/packages/characters) with `built_value` somehow?

mit-mit:
cc @davidmorgan @lrhn

davidmorgan:
Could you give a bit more info on the issue you're seeing, please?

domesticmouse:
30,000' explanation is that I'm incorrectly rendering search requests that contain non-latin1 characters:

<img width="806" alt="Screen Shot 2020-11-07 at 11 42 38 am" src="https://user-images.githubusercontent.com/30503/98426785-5973d200-20ee-11eb-9634-9fdd0a054d4c.png">

In digging into the problem (see https://github.com/flutter/samples/pull/587/files) I figured out a couple of things. First, our `http` package will default to decoding an `application/json` stream with the latin1 parser. Unfortunately for me, Unsplash API is returning API results using UTF8 without declaring it as such.

davidmorgan:
Thanks--indeed, looks like the issue is outside `built_value`.

lrhn:
And outside `characters` too. You'll have to set the encoding on the HTTP request manually.
```dart
import "package:http/http.dart";
...
  var response = await get(uri);
  var data = utf8.fuse(json).decode(response.bodyBytes);
```
(Do use the `http` package for HTTP requests, because using `dart:io` is not compatible with Flutter-for-Web).

domesticmouse:
With the above integrated in https://github.com/flutter/samples/pull/590/commits/8e0f5c441fc2d4eaf7b3b4526c424336f6bd0202, it still trips up on the following on macOS (but renders fine in Chrome).

```json
      "user": {
        "id": "JzhDKu_NfNs",
        "updated_at": "2020-11-08T00:33:05-05:00",
        "username": "arusfly",
        "name": "Arusfly 🌿",
        "first_name": "Arusfly 🌿",
        "last_name": null,
        "twitter_username": "arusfly",
        "portfolio_url": "http://www.instagram.com/vallartadiarios",
        "bio": null,
        "location": "Puerto Vallarta",
        "links": {
          "self": "https://api.unsplash.com/users/arusfly",
          "html": "https://unsplash.com/@arusfly",
          "photos": "https://api.unsplash.com/users/arusfly/photos",
[...]

PR #590, opened 2020-11-10
UTF-8 fix

domesticmouse:
This is an attempt at fixing https://github.com/flutter/samples/issues/586. This PR is based on https://github.com/flutter/samples/pull/588 which is why it contains a lot of changes. It'll become a one file change once #588 lands.


RedBrogdon (approved):
Just to make sure I understand what's going on here: the issue is that Unsplash is returning UTF8 text in its JSON responses, but not using headers to correctly declare that, which causes problems for the default decoding used by dart:convert and built_value?

Also: LGTM.

domesticmouse:
Yup, that summary is correct =)

PR #593, opened 2020-11-12
Styling revamp using CupertinoThemeData and CupertinoTextThemeData

ADVavvas:
This was done to handle styling in a manner more in keeping with current SDK capabilities. (Issue #322 )

### DIFF
--- a/experimental/veggieseasons/lib/screens/details.dart
+++ b/experimental/veggieseasons/lib/screens/details.dart
@@ -28,12 +28,10 @@ class ServingInfoChart extends StatelessWidget {
       builder: (context, snapshot) {
         final target = snapshot?.data ?? 2000;
         final percent = standardPercentage * 2000 ~/ target;
-        final themeData = CupertinoTheme.of(context);
 
         return Text(
           '$percent% DV',
           textAlign: TextAlign.end,
-          style: Styles.detailsServingValueText(themeData),
         );
       },
     );
@@ -54,7 +52,6 @@ class ServingInfoChart extends StatelessWidget {
             ),
             child: Text(
               'Serving info',
-              style: Styles.detailsServingHeaderText,
             ),
           ),
         ),
@@ -79,7 +76,6 @@ class ServingInfoChart extends StatelessWidget {
                         child: Text(
                           veggie.servingSize,
                           textAlign: TextAlign.end,
-                          style: Styles.detailsServingValueText(themeData),
                         ),
                       ),
                     ],
@@ -96,7 +92,6 @@ class ServingInfoChart extends StatelessWidget {
                         child: Text(
                           '${veggie.caloriesPerServing} kCal',
                           textAlign: TextAlign.end,
-                          style: Styles.detailsServingValueText(themeData),
                         ),
                       ),
                     ],
@@ -184,7 +179,7 @@ class InfoView extends StatelessWidget {
                     style: (snapshot.hasData &&
                             snapshot.data.contains(veggie.category))
                         ? Styles.detailsPreferredCategoryText(themeData)
-                        : Styles.detailsCategoryText(themeData),
+                        : themeData.textTheme.textStyle,
                   );
                 },
               ),
@@ -210,7 +205,6 @@ class InfoView extends StatelessWidget {
           SizedBox(height: 8),
           Text(
             veggie.shortDescription,
-            style: Styles.detailsDescriptionText(themeData),
           ),
           ServingInfoChart(veggie, prefs),
           SizedBox(height: 24),
@@ -249,7 +243,8 @@ class DetailsScreen extends StatefulWidget {
   static Route<void> _routeBuilder(BuildContext context, Object arguments) {
     final veggieId = arguments as int;
     return CupertinoPageRoute(
-      builder: (context) => DetailsScreen(id: veggieId, restorationId: 'details'),
+      builder: (context) =>
+          DetailsScreen(id: veggieId, restorationId: 'details'),
       fullscreenDialog: true,
     );
   }
--- a/experimental/veggieseasons/lib/screens/details.dart
+++ b/experimental/veggieseasons/lib/screens/details.dart
@@ -31,6 +31,7 @@ class ServingInfoChart extends StatelessWidget {
 
         return Text(
           '$percent% DV',
+          style: CupertinoTheme.of(context).textTheme.textStyle,
           textAlign: TextAlign.end,
         );
       },
@@ -52,6 +53,7 @@ class ServingInfoChart extends StatelessWidget {
             ),
             child: Text(
               'Serving info',
+              style: CupertinoTheme.of(context).textTheme.textStyle,
             ),
           ),
         ),
@@ -76,6 +78,7 @@ class ServingInfoChart extends StatelessWidget {
                         child: Text(
                           veggie.servingSize,
                           textAlign: TextAlign.end,
+                          style: CupertinoTheme.of(context).textTheme.textStyle,
                         ),
                       ),
                     ],
@@ -91,6 +94,7 @@ class ServingInfoChart extends StatelessWidget {
                       TableCell(
                         child: Text(
                           '${veggie.caloriesPerServing} kCal',
+                          style: CupertinoTheme.of(context).textTheme.textStyle,
                           textAlign: TextAlign.end,
                         ),
                       ),
@@ -205,6 +209,7 @@ class InfoView extends StatelessWidget {
           SizedBox(height: 8),
           Text(
             veggie.shortDescription,
+            style: CupertinoTheme.of(context).textTheme.textStyle,
           ),
           ServingInfoChart(veggie, prefs),
           SizedBox(height: 24),
@@ -319,8 +324,14 @@ class _DetailsScreenState extends State<DetailsScreen> with RestorationMixin {
                   SizedBox(height: 20),
                   CupertinoSegmentedControl<int>(
                     children: {
-                      0: Text('Facts & Info'),
-                      1: Text('Trivia'),
+                      0: Text(
+                        'Facts & Info',
+                      ),
+                      //style: CupertinoTheme.of(context).textTheme.textStyle),
+                      1: Text(
+                        'Trivia',
+                      )
+                      //style: CupertinoTheme.of(context).textTheme.textStyle),
                     },
                     groupValue: _selectedViewIndex.value,
                     onValueChanged: (value) {
--- a/experimental/veggieseasons/lib/screens/details.dart
+++ b/experimental/veggieseasons/lib/screens/details.dart
@@ -327,11 +327,9 @@ class _DetailsScreenState extends State<DetailsScreen> with RestorationMixin {
                       0: Text(
                         'Facts & Info',
                       ),
-                      //style: CupertinoTheme.of(context).textTheme.textStyle),
                       1: Text(
                         'Trivia',
                       )
-                      //style: CupertinoTheme.of(context).textTheme.textStyle),
                     },
                     groupValue: _selectedViewIndex.value,
                     onValueChanged: (value) {
--- a/experimental/veggieseasons/lib/screens/list.dart
+++ b/experimental/veggieseasons/lib/screens/list.dart
@@ -42,8 +42,8 @@ class ListScreen extends StatelessWidget {
         final prefs = Provider.of<Preferences>(context);
         final themeData = CupertinoTheme.of(context);
         return AnnotatedRegion<SystemUiOverlayStyle>(
-          value:
-              SystemUiOverlayStyle(statusBarBrightness: themeData.brightness),
+          value: SystemUiOverlayStyle(
+              statusBarBrightness: MediaQuery.of(context).platformBrightness),
           child: SafeArea(
             bottom: false,
             child: ListView.builder(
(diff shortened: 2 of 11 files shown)


RedBrogdon (commented):
Thanks for the PR, @ADVavvas. Has this been tested with both light and dark mode?

ADVavvas:
Hi, forgot to include an important addition to my PR. Should be working fine now for both light and dark mode.

There are still 2 issues with the light/dark theme (although I'm not too sure how to deal with them).

1. Changing IOS theme while the app is running doesn't update the "In Season Today" headline color (Probably because the widget is not rebuilding?)
2. On the 'Home' and 'Search' tabs the status bar color (time, wifi, battery) doesn't change (if you change theme while the app is running). It does change if you go to 'My Garden' or 'Settings' (probably because they have a `NavigationBar`)

experimental/veggieseasons/lib/screens/details.dart
```diff
@@ -324,8 +324,14 @@ class _DetailsScreenState extends State<DetailsScreen> with RestorationMixin {
                   SizedBox(height: 20),
                   CupertinoSegmentedControl<int>(
                     children: {
-                      0: Text('Facts & Info'),
-                      1: Text('Trivia'),
+                      0: Text(
+                        'Facts & Info',
+                      ),
+                      //style: CupertinoTheme.of(context).textTheme.textStyle),
```
RedBrogdon, line 330: nit: remove any commented-out code.

experimental/veggieseasons/lib/screens/list.dart
```diff
@@ -32,14 +32,14 @@ class ListScreen extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final themeData = CupertinoTheme.of(context);
```
RedBrogdon, line 35: This is the reason the "In Season Today" font is not re-rendering with the correct color when you switch between dark and light mode. I'll explain:

When you call one of these `of` methods on an InheritedWidget and give it a context, the InheritedWidget remembers the context you give it, and will automatically rebuild the widget at that point in the tree if it needs to (such as when CupertinoTheme detects that the device has switched from dark to light mode or vice versa).

In this bit of code, you're using the context of ListScreen itself. That's fine for most cases, but here, ListScreen is building a CupertinoTabView. CupertinoTabView has as one of its children, a Navigator widget.

CupertinoTabView gives that Navigator the builder method on line 38 to use when building its children. If you rebuild the CupertinoTabView and give it a *different* builder function, the Navigator will use that new function any time it needs to build its child widgets in the future, **but** it will not immediately trigger a rebuild of them. This is because Navigators need to hold multiple routes simultaneously, stacked on top of each other. They need to remember what routes have already been pushed. Those routes have their own widgets and state, which could be lost if they were automatically rebuilt in response to the new builder method.

This can be slightly confusing to wrap your head around. Fortunately, the fix is easy: move the call to the InheritedWidget to be *inside* the builder method that the Navigator is using. That way, the context that's used will *also* be inside the Navigator's subtree of widgets, which means there's no change to the builder method, and instead the right widget will be rebuilt automatically, further down the tree.

In this case, that means taking the call that's now on line 35, and move it back to line 42. The font will begin to change correctly after that.

experimental/veggieseasons/lib/widgets/trivia.dart
```diff
@@ -150,21 +148,15 @@ class _TriviaViewState extends State<TriviaView> with RestorationMixin {
               ),
               Padding(
                 padding: const EdgeInsets.symmetric(horizontal: 16),
-                child: Text(
-                  ' of ',
-                  style: Styles.triviaFinishedText(themeData),
-                ),
+                child: Text(' of ', style: themeData.textTheme.textStyle),
```
RedBrogdon, line 151: In testing your change, I discovered a pre-existing issue on line 146. Would you mind changing it from:

```
'$score',
```

to:

```
'${score.value}',
```

RedBrogdon (changes requested):
Thanks, @ADVavvas. This PR looks pretty good. I just have a few comments, which are inline.

For #2, I am able to replicate the issue, but I'm not sure whether it's something wrong with this app or an SDK bug. @LongCatIsLooong, would you mind taking a look? Steps to reproduce are:

* Launch Veggieseasons in light mode.
* Switch to the "My Garden" screen using the bottom navigation bar.
* Switch back to the "Home" screen.
* Change from light mode to dark mode using iOS developer settings.
* Return to the app, and the status bar across the top will have no visible text for time, battery, etc.
* Switch to the "My Garden" screen, and white text will appear in the status bar.
* Return to the "Home" screen, and the white text will remain.

This is on an iPhone SE Simulator running iOS 14, with this `flutter doctor` output:

```
[✓] Flutter (Channel master, 1.24.0-8.0.pre.267, on Mac OS X 10.15.6 19G2021 darwin-x64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 12.0)
[✓] Chrome - develop for the web
[✓] Android Studio (version 4.0)
[✓] IntelliJ IDEA Community Edition (version 2020.1.2)
[✓] VS Code (version 1.50.1)
[✓] Connected device (4 available)
```

Is that WAI and there's something we're doing wrong in this app, or is the SDK not behaving itself?

experimental/veggieseasons/lib/screens/details.dart
```diff
@@ -28,12 +28,11 @@ class ServingInfoChart extends StatelessWidget {
       builder: (context, snapshot) {
         final target = snapshot?.data ?? 2000;
         final percent = standardPercentage * 2000 ~/ target;
-        final themeData = CupertinoTheme.of(context);
 
         return Text(
           '$percent% DV',
+          style: CupertinoTheme.of(context).textTheme.textStyle,
```
LongCatIsLooong, line 34: nit: is it needed to explicitly specify the text style? I'd expect the `Text` widget to inherit the style from the theme.

experimental/veggieseasons/lib/styles.dart
```diff
@@ -3,171 +3,115 @@
...
-  static const minorText = TextStyle(
-    color: Color.fromRGBO(128, 128, 128, 1),
-    fontFamily: 'NotoSans',
-    fontSize: 16,
-    fontStyle: FontStyle.normal,
-    fontWeight: FontWeight.normal,
-  );
+  static TextStyle minorText(CupertinoThemeData themeData) =>
+      themeData.textTheme.textStyle.copyWith(
+        color: Color.fromRGBO(128, 128, 128, 1),
```
LongCatIsLooong, line 34: It seems like we're using a regular color. Is "minorText" supposed to use the same color when you switch from light mode to dark mode? Same with "cardTitleText" and "cardCategoryText", etc,.

experimental/veggieseasons/lib/screens/details.dart
```diff
@@ -324,8 +324,14 @@ class _DetailsScreenState extends State<DetailsScreen> with RestorationMixin {
...
-                      0: Text('Facts & Info'),
-                      1: Text('Trivia'),
+                      0: Text(
+                        'Facts & Info',
+                      ),
+                      //style: CupertinoTheme.of(context).textTheme.textStyle),
+                      1: Text(
+                        'Trivia',
+                      )
+                      //style: CupertinoTheme.of(context).textTheme.textStyle),
```
LongCatIsLooong, line 334: I vaguely remember `CupertinoSegmentedControl` doesn't work well with dark mode, because the UI component no longer exists on iOS 13 and above.

experimental/veggieseasons/lib/screens/list.dart
ADVavvas: That makes a lot of sense. Thanks for the insight!

experimental/veggieseasons/lib/screens/details.dart
ADVavvas: Perhaps I've done something wrong in assigning the theme to the whole app, but removing the `style` makes the text black regardless of the IOS brightness.

experimental/veggieseasons/lib/styles.dart
ADVavvas: `minorText` is used for the date in the Home Screen. It looks okay with both themes since it's grey.

The cards are all white with black text, which also works fine since they have a picture (and shadows).
I haven't changed any of these values. I assume it was intentional.

experimental/veggieseasons/lib/screens/details.dart
ADVavvas: Seems okay on IOS 13.1. 
Blue background with black letters for selected. Black background with blue letters for unselected.
Have not applied any styling to the its children.
Not sure what it's supposed to look like.

ADVavvas:
That should be all the changes (except #2).

Thanks you both! Appreciate the feedback.
Let me know if there's anything else.

RedBrogdon (approved):
This LGTM. @LongCatIsLooong, WDYT?

experimental/veggieseasons/lib/screens/details.dart
LongCatIsLooong: Does it work if you wrap the `DetailsScreen` widget in a `DefaultTextStyle(style: CupertinoTheme.of(context).textTheme.textStyle)` widget? Anyways not a big deal it just makes the code a little bit more verbose.

experimental/veggieseasons/lib/screens/details.dart
LongCatIsLooong: On iOS 13+ `UISegmentedControl`s have a different look so I'm not sure either what it is supposed to look like. I'm okay with this since there's no actual code change (could you remove the lines that are commented out?)

experimental/veggieseasons/lib/styles.dart
LongCatIsLooong: I see. The cards all have very bright backgrounds because of the images they used. Make sense to opt-out of dark mode on the cards.

experimental/veggieseasons/lib/widgets/veggie_card.dart
```diff
@@ -5,6 +5,7 @@
 import 'dart:ui';
 
 import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
```
LongCatIsLooong, line 8: nit: not sure why the material library needs to be imported here. Can this be removed?

experimental/veggieseasons/lib/styles.dart
```diff
@@ -3,171 +3,115 @@
 // found in the LICENSE file.
 
 import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
```
LongCatIsLooong, line 6: nit: can we remove this import?

LongCatIsLooong (commented):
LGTM with nits.

LongCatIsLooong:
Ah didn't notice this

>  On the 'Home' and 'Search' tabs the status bar color (time, wifi, battery) doesn't change (if you change theme while the app is running). It does change if you go to 'My Garden' or 'Settings' (probably because they have a NavigationBar)

I thought this bug has long been fixed, but the status bar icons are not visible on my android device either. Taking a look.

ADVavvas:
Added the `AnnotatedRegion<SystemUiOverlayStyle>` on both list.dart and search.dart. Not sure where exactly to place it in the tree, so I'd appreaciate if you could take a look @LongCatIsLooong . Works fine though. Also removed the unnecessary imports.

experimental/veggieseasons/lib/screens/list.dart
```diff
@@ -40,43 +41,48 @@ class ListScreen extends StatelessWidget {
...
-                var relativeIndex =
-                    index - (appState.availableVeggies.length + 2);
-                return _generateVeggieRow(
-                    appState.unavailableVeggies[relativeIndex], prefs,
-                    inSeason: false);
-              }
-            },
+        return AnnotatedRegion<SystemUiOverlayStyle>(
+          value:
+              SystemUiOverlayStyle(statusBarBrightness: themeData.brightness),
```
LongCatIsLooong, line 46: `CupertinoThemeData.brightness` is the brightness override and it's nullable. I think this should be `MediaQuerty.brightnessOf(context)`. Same with the search page.

ADVavvas: Done! Ignore the first commit (wrote `MediaQuery.of(context).platformBrightness` instead. Not sure if that makes any difference)

LongCatIsLooong: IIRC they're the same.

LongCatIsLooong (approved):
Thank you!

ADVavvas:
Quick question since this is my 1st PR: Do I need to do anything else or is it going to be merged at some point?

LongCatIsLooong:
@ADVavvas sorry for the late reply. I'm not particularly familiar with the samples repo procedures. @RedBrogdon is it good to merge this PR?

RedBrogdon:
Yes, this is ready to land. I would have done it last week, but was out for Thanksgiving. 🦃

RedBrogdon:
@ADVavvas, thanks for sticking with this PR through all the review. It's definitely an improvement. I wrote that original text styling two years ago -- it was seriously out of date, but I had no time to fix it. 😄

PR #596, opened 2020-11-16
[samples_index] Fix search bug when using filters

### LINKED ISSUE #392: [visual samples index] Inaccurate search results when using filters and search
e.g. https://flutter.github.io/samples/#?search=ios&type=cookbook

cazci:
Fixes #392 

Currently, the search results are inaccurate when type filters are used. I've fixed this by adding an additional type check before the current checks. This loop will check whether a type parameter is present in the search query, and return false if the current self type mismatches the query type. Hope this helps :)

I've tested this against a set of custom queries, and the following is a screenshot when tested against the query mentioned in the issue.
`?search=ios&type=cookbook`

sample_index

### DIFF
--- a/web/samples_index/lib/src/search.dart
+++ b/web/samples_index/lib/src/search.dart
@@ -9,6 +9,17 @@ bool matchesQuery(String query, String sampleAttributes) {
   var queryWords = query.split(' ')..removeWhere((s) => s.isEmpty);
   var attributes = sampleAttributes.split(' ')..removeWhere((s) => s.isEmpty);
 
+  // Test for type filter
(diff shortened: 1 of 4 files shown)


johnpryan (approved):
Looks good, want to add a test to [this file](https://github.com/flutter/samples/blob/master/web/samples_index/test/samples_index_test.dart#L99)?

cazci:
Sure, I'll add

cazci:
I've added three tests and squashed the commits. Test output is as follows, 

tests

johnpryan (approved):
looks good to me, thanks!

PR #598, opened 2020-11-19
[samples_index] Implement carousel

### LINKED ISSUE #393: [visual sample index] Add Carousel
Design spec:

Screen Shot 2020-03-23 at 2 18 19 PM

cazci:
Fixes #393 

I've implemented a simple animated carousel to list screenshots. I got inspired by [this](https://codepen.io/flowrome/pen/abowVRE) codepen project. I've used the base code and customized it to match the proposed design. 

* Kudos to [flowrome](https://codepen.io/flowrome)
* I've used font awesome icons, which is fully [open source](https://fontawesome.com/v4.7.0/license)

A sample screen recording is as follows,
preview

A high-quality preview can be accessed through the following link, thanks!
https://drive.google.com/file/d/1BABOk0WCNgUHaUFl0KSO78fyDzHTeWqE/view?usp=sharing

### DIFF
--- a/web/samples_index/lib/src/templates.dart
+++ b/web/samples_index/lib/src/templates.dart
@@ -51,6 +51,7 @@ String _descriptionHeader = '''
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <script src="packages/mdc_web/material-components-web.min.js"></script>
+  <script src="https://kit.fontawesome.com/16cc04762e.js"></script>
   <script defer src="description.dart.js"></script>
 </head>
 ''';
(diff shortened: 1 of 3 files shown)


RedBrogdon:
Thanks for the PR, @cazci. There are a couple issues here I need to cover, though.
 
* While CodePen examples are usually MIT licensed by default, we wouldn't be able to merge the code here unless the original author signs the Google Contributors License agreement just like you have. 
* The samples index is written in Dart so that it, too, can function as a sample application. That's an important design choice we'd like to maintain, and landing a JS file would work against it.

The latter of these is a solvable problem (port the code to Dart), but I'm not sure about the first one. Do you know the original author of that CodePen?

cazci:
Thanks for replying, @RedBrogdon 

On the 1st issue, I'm sorry I didn't know that, and unfortunately I don't know the original author either. So, unless he has already signed the Google CLA, I guess this code can't be used. 

However, I can try to implement another carousel using dart which will eventually solve both issues. WDYT? :)

RedBrogdon:
That would solve both those issues. 😄 

I'll close this PR. Feel free to open another one if you have time to try your hand at a new implementation.

PR #603, opened 2020-11-25
let add-to-app ios demo page transitions be animated

xster:
Not sure why we left in un-animated originally. API's on stable.

### DIFF
--- a/add_to_app/flutter_module/lib/main.dart
+++ b/add_to_app/flutter_module/lib/main.dart
@@ -147,7 +147,6 @@ class Contents extends StatelessWidget {
                 if (showExit) ...[
                   SizedBox(height: 16),
                   RaisedButton(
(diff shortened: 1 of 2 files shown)


add_to_app/flutter_module/lib/main.dart
```diff
@@ -148,6 +148,7 @@ class Contents extends StatelessWidget {
                   SizedBox(height: 16),
                   RaisedButton(
                     onPressed: () => SystemNavigator.pop(),
```
domesticmouse, line 150: Delete this line? The duplicate `onPressed:` is triggering a build failure on travis.

xster: oops, thanks

domesticmouse (approved):

PR #606, opened 2020-11-30
Add navigator 2 samples

johnpryan:
These are the samples from [Learning Flutter's new Navigation and Routing System](https://medium.com/flutter/learning-flutters-new-navigation-and-routing-system-7c9068155ade)

Since the framework APIs for routing are available in the stable channel, these don't need to be in the experimental directory I don't think (Deep linking and web routing are only supported in the beta channel, so it's also valid to argue this is experimental)

cc: @chunhtai @goderbauer


navigation_and_routing/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Samples
+
+**Navigator 1.0 samples**
+
+* `nav_1/anonymous_routes.dart` - how to use a Navigator to push and pop
+  anonymous routes (e.g. MaterialPageRoute)
+* `nav_1/named_routes.dart` - how to use define named routes via the `routes`
+  parameter on MaterialApp, and navigate using Navigator.pushNamed
+* `nav_1/on_generate_route.dart` - Shows how to handle arbitrary named routes
```
RedBrogdon, line 14: nit: `Show` is capitalized, but `how` on lines 10 and 12 was not.

navigation_and_routing/pubspec.yaml
```diff
@@ -0,0 +1,76 @@
+name: navigation_and_routing
```
RedBrogdon, line 1: nit: Remove autogenerated comments. Would be good to change description as well.

RedBrogdon (approved):
Discussed offline. I have a couple of nits, but otherwise LGTM.

This is great code and we definitely want to bring it to people's attention, but I think we should continue to think about whether we can find a better, more easily consumable home for it than here.

In the meantime, though: great stuff, ready to merge after fixing a couple nits.

PR #607, opened 2020-12-02
[animation] Replaced depricated property(overflow) with new one(clipBehavior)

Alabhya268:
overflow property was deprecated after v1.22.0-12.0.
'clipBehaviour' is new replacement of 'overflow'.

### DIFF
--- a/animations/lib/src/misc/card_swipe.dart
+++ b/animations/lib/src/misc/card_swipe.dart
@@ -45,6 +45,7 @@ class _CardSwipeDemoState extends State<CardSwipeDemo> {
               Expanded(
                 child: ClipRect(
                   child: Stack(
+                    clipBehavior: Clip.hardEdge,
                     children: <Widget>[
                       for (final fileName in fileNames)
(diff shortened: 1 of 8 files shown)


RedBrogdon (changes requested):
Interesting. The current stable version doesn't include the deprecation warning, which is why our automated tests weren't catching any analyzer issues.

This is a good catch, so thanks @Alabhya268. If you check the current SDK code, though, I believe you'll find `hardEdge` is the default `clipBehavior` for a Stack, so you can simply remove line 48 entirely.

https://github.com/flutter/flutter/pull/66305/files#diff-437969bbec991153bb60d13acb966e6fbb8283bc2ae3c5f1c4abbe4b0160c38fL3278

Alabhya268:
@RedBrogdon I've now entirely removed depricated line, Please have a look .

RedBrogdon (approved):
Always nice to see unnecessary code taken out of a project. 😄

PR #608, opened 2020-12-03
[samples_index] Fix carousel edge cases and make mobile responsive

cazci:
Currently, the carousel doesn't render well when no images or only one image is present. This will solve this issue by 
* Skipping the carousel container space when no images are present
* Rendering only the image without carousel options when only one image is present

I missed this case in the implementation, and fixed asap I noted

### DIFF
--- a/web/samples_index/lib/src/carousel.dart
+++ b/web/samples_index/lib/src/carousel.dart
@@ -14,6 +14,9 @@ class Carousel {
 
   Element prevSlide, currentSlide, nextSlide;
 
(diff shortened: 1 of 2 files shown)


cazci:
I updated the carousel to be mobile responsive and implemented a swipe gesture listener to detect touch events.

Please refer the following preview, :) 

preview_mobile

johnpryan (approved):

johnpryan:
Cool, thanks for following up!

Issue #573, opened 2020-10-22
[Veggie Season] The veggies saved to Garden are deleted when app is closed


kjain333:
A possible bug in the Veggie Season application is when we close our application the data we save to the garden gets deleted. A possible fix for this may be to use shared preferences to store favorite veggies instead of the app state.

RedBrogdon:
That's definitely not the desired behavior. VeggieSeason should be using shared_preferences to store the data on device between executions.

Can you provide a little more information? What device and iOS version were you using when you saw this happen, for instance?

kjain333:
Respected sir,
I am really sorry for not providing this information before but due to system constraints I was presently just able to run this code on my android with android version 10 but looking at the code itself since we are using shared preferences to store the preferred calorie count and category of veggie why not we use the same for garden as well?

rj-since-2000:
Unlike the persistent store of data in the shared_preference for [_caloriesKey] and [_preferredCategoriesKey] the data for My Garden is not stored or saved anywhere! (Tested on both iOS and android)
Below code reference --> lib/data/preferences.dart/[line 57]

```
  Future<void> _saveToSharedPrefs() async {
    final prefs = await SharedPreferences.getInstance();
    await prefs.setInt(_caloriesKey, _desiredCalories);

    // Store preferred categories as a comma-separated string containing their
    // indices.
    await prefs.setString(_preferredCategoriesKey,
        _preferredCategories.map((c) => c.index.toString()).join(','));
  }
```

Here only the values of calories target and preferred categories are stored and saved after state of the app is closed. The data of [My Garden] is not stored anywhere that's why the favourite data for my garden is not saved and gets renewed to null every time the app restarts. This is probably not what a user would want!

     1-User marks veggies as      ->      2-Data gets saved and shows in       ->      3-User restarts the app and the data 
           [Save to Garden]                                 [My Garden]                for [My Garden] gets renewed to null                        
<img src="https://user-images.githubusercontent.com/68644104/97758211-64de6080-1b24-11eb-822d-ce0b5103f950.jpg" width="200">---------------><img src="https://user-images.githubusercontent.com/68644104/97758220-6871e780-1b24-11eb-91de-f134db6e43b2.jpg" width="200">---------------><img src="https://user-images.githubusercontent.com/68644104/97758221-690a7e00-1b24-11eb-953c-aaf3a7d4357f.jpg" width="200">


@RedBrogdon 
[...]

PR #617, opened 2020-12-10
[Veggie Seasons] Storing my garden to shared_preferences 

tusharojha:
## Description

The Veggie Seasons app doesn't store my garden which contains favorite veggies to shared_preferences due to which the user cannot see the data after every execution. Related Issue #573.

**Note:** As veggie seasons app is moved to the experimental folder to implement the state restoration as discussed in #433 by @goderbauer. I have worked on the beta channel and tested for all the tests & working on the app. 

The following gif shows the working:
ezgif com-gif-maker

### DIFF
--- a/experimental/veggieseasons/lib/main.dart
+++ b/experimental/veggieseasons/lib/main.dart
@@ -30,16 +30,8 @@ class VeggieApp extends StatefulWidget {
   State<StatefulWidget> createState() => _VeggieAppState();
 }
 
-class _VeggieAppState extends State<VeggieApp> with RestorationMixin {
-  final _RestorableAppState _appState = _RestorableAppState();
-
-  @override
-  String get restorationId => 'wrapper';
-
-  @override
-  void restoreState(RestorationBucket oldBucket, bool initialRestore) {
-    registerForRestoration(_appState, 'state');
-  }
+class _VeggieAppState extends State<VeggieApp> {
+  final _appState = AppState();
 
   @override
   void dispose() {
@@ -52,7 +44,7 @@ class _VeggieAppState extends State<VeggieApp> with RestorationMixin {
     return MultiProvider(
       providers: [
         ChangeNotifierProvider.value(
-          value: _appState.value,
+          value: _appState,
         ),
         ChangeNotifierProvider(
           create: (_) => Preferences()..load(),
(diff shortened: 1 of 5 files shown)


tusharojha:
@RedBrogdon & @domesticmouse kindly review and suggest changes if any.

experimental/veggieseasons/lib/main.dart
```diff
@@ -80,6 +82,7 @@ class _RestorableAppState extends RestorableListenable<AppState> {
     final favorites = (data as List<dynamic>).cast<int>();
     for (var id in favorites) {
       appState.setFavorite(id, true);
+      preferences.setFavioriteVeggie(id, true);
```
RedBrogdon, line 85: The approach that you've taken looks designed to minimize the changes to the existing codebase, which is an understandable goal. I think, though, that it's resulted in a solution that's spreading information around more than it needs to be.

If you look at the AppState class, you'll notice that it doesn't store any of the information currently kept in preferences (desired calories and preferred categories). Likewise, the Preferences class didn't store any information about the list of veggies.

In this change, however, information about favorite veggies is getting spread among both classes, which is resulting in strange situations like this one (setFavorite and setFavoriteVeggie both need to be called, even though the concern the same information).  

A better solution would be to find a way to remove the favorites information from AppState entirely and keep it solely in Preferences, or devise some other way the maintain separation of concern.

This would likely result in a much more complicated PR than you had originally anticipated, so it's perfectly reasonable if you'd prefer to just close this.

RedBrogdon (changes requested):

experimental/veggieseasons/lib/main.dart
tusharojha: Actually, As you said I tried to solve the issue with minimum changes in the existing code since I haven't discussed the right approach to solve it. 

I would love to learn and contribute to solving this issue. That's why I have set it to WIP and working on separating favorites information from the AppState.

tusharojha:
I tried to separate the favorite veggies information from the `AppState` class but in that case, there will be no use of the `_RestorableAppState` class because the favorites data which it is sending and receiving from primitives is not stored in `AppState` anymore and we can fetch & store favorites using `Preferences` & FutureBuilder just like Calories & Preferred Categories.

`_VeggieAppState` would look like this:

```dart
class _VeggieAppState extends State<VeggieApp> {
  final _appState = AppState();

  @override
  void dispose() {
    _appState.dispose();
    super.dispose();
  }

  @override
    return MultiProvider(
      providers: [
        ChangeNotifierProvider.value(
          value: _appState,
        ),
        ChangeNotifierProvider(
          create: (_) => Preferences()..load(),
        ),
      ],
      child: CupertinoApp(
        theme: Styles.veggieThemeData,
        debugShowCheckedModeBanner: false,
        home: HomeScreen(restorationId: 'home'),
        restorationScopeId: 'app',
      ),
    );
  }
}

```

Although, after changing to this and separating the favorites from the app state it's working I am confused about its effect on the state restoration of the app.

tusharojha:
Closing this PR.

Issue #399, opened 2020-03-25
[animations] Improve unit/widget tests in the app.


AyushBherwani1998:
## Description
Currently, in the widget test, we are only checking whether the `Text` is rendered correctly or not. Adding more widget/unit tests in the application can help the Flutter enthusiast to get more detailed examples of the writing unit/widget test.

aksayushx:
Hi @AyushBherwani1998 ,
The demonstration of writing unit tests and widget tests is aimed to be done with a new Flutter testing sample app GSoC project which is listed in the [GSoC project ideas list](https://github.com/dart-lang/sdk/wiki/Dart-GSoC-2020-Project-Ideas) and will be taken up this summer.

jddeep:
@AyushBherwani1998 Thanks for filing the issue. Make sure you add tests for the different animation demos themselves.

AyushBherwani1998:
@jddeep Yes, having something similar in mind.

@RedBrogdon in most of the samples. we are generating values randomly. I think checking the values before and after animations will not be the best choice. It might be the case, that the same value is generated again. Well, it will happen in rare case. WDYT?

RedBrogdon:
@AyushBherwani1998, writing tests for our sample apps can sometimes be tricky, since they're often designed to show off the basic widgets present in the SDK. While you could certainly add a test for the ListView widget itself, that test really belongs in the SDK, rather than here.

What you *could* do here is add tests for the demos that are complicated enough that it would make sense. The carousel demo, for example, is complex enough, as would be the expand card. If you're interested, I'd start with a small pull request for just one of them, as it'll be easier to review, revise, and merge.

AyushBherwani1998:
> 
> 
> @AyushBherwani1998, writing tests for our sample apps can sometimes be tricky, since they're often designed to show off the basic widgets present in the SDK. While you could certainly add a test for the ListView widget itself, that test really belongs in the SDK, rather than here

Makes Senses.

AyushBherwani1998:
> What you _could_ do here is add tests for the demos that are complicated enough that it would make sense. The carousel demo, for example, is complex enough, as would be the expand card. If you're interested, I'd start with a small pull request for just one of them, as it'll be easier to review, revise, and merge.

Yes, I'm interested in working on complex examples.

AyushBherwani1998:
[...]

PR #620, opened 2020-12-14
[animations] added test for HeroAnimation

arteevraina:
Added widget test for `HeroAnimation`

This is with regard to issue #399

### DIFF
--- a/animations/test/misc/hero_animation_test.dart
+++ b/animations/test/misc/hero_animation_test.dart
@@ -12,19 +12,26 @@ Widget createHeroAnimationDemoScreen() => MaterialApp(
 
 void main() {
   group('Hero Animation Tests', () {
-    testWidgets('Heigth of Container changes on Tap', (tester) async {
+    testWidgets('Size of Container changes on Tap', (tester) async {
       await tester.pumpWidget(createHeroAnimationDemoScreen());
 
       // Get the initial Container.
       final initialContainer =
           tester.firstWidget(find.byType(Container)) as Container;
 
-      // Get height of Container.
+      // Get initial height of Container.
       // In this case minHeight is equal to maxHeight.
-
       var initialHeight =
           initialContainer.constraints.heightConstraints().maxHeight;
 
+      // Get initial width of Container.
+      // In this case minWidth is equal to maxWidth.
+      var initialWidth =
+          initialContainer.constraints.widthConstraints().maxWidth;
+
+      // Calculate initial area of Container
+      var initialArea = initialHeight * initialWidth;
+
       // Tap on the GestureDetector.
       await tester.tap(find.byType(GestureDetector));
       await tester.pumpAndSettle();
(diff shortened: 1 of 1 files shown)


arteevraina:
@googlebot I signed it!

animations/test/misc/hero_animation_test.dart
```diff
@@ -0,0 +1,127 @@
...
+      await tester.pumpWidget(createHeroAnimationDemoScreen());
+
+      // Get the initial Container.
+      final initialContainer =
+          tester.firstWidget(find.byType(Container)) as Container;
+
+      // Get height of Container.
+      // In this case minHeight is equal to maxHeight.
+
+      var initialHeight =
```
tusharojha, line 25: nit: No need for this extra line gap between code & comments.

animations/test/misc/hero_animation_test.dart
```diff
@@ -0,0 +1,127 @@
...
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+Widget createHeroAnimationDemoScreen() => MaterialApp(
+      home: HeroAnimationDemo(),
+    );
+
+void main() {
+  group('Hero Animation Tests', () {
+    testWidgets('Heigth of Container changes on Tap', (tester) async {
```
tusharojha, line 15: nit: typo in height.

animations/test/misc/hero_animation_test.dart
```diff
@@ -0,0 +1,127 @@
...
+      var finalHeight =
+          finalContainer.constraints.heightConstraints().maxHeight;
+
+      expect(
+        initialHeight,
+        lessThan(finalHeight),
+      );
+    });
+
+    testWidgets('Width of Container changes on Tap', (tester) async {
```
tusharojha, line 48: I think comparing both height & width in separate test cases is not required. It would be more efficient to check for complete size instead.

tusharojha (changes requested):
Hi @arteevraina 
I have suggested a few changes and @RedBrogdon will review your PR before merging.

animations/test/misc/hero_animation_test.dart
arteevraina: Since, the `Container` is constrained on both sides. I think a better approach will be to take the area and test it against final area of `Container`. What are your view about this?

animations/test/misc/hero_animation_test.dart
arteevraina: Fixed this.

arteevraina:
@tusharojha Thanks for the review :)

animations/test/misc/hero_animation_test.dart
```diff
@@ -0,0 +1,109 @@
...
+      home: HeroAnimationDemo(),
+    );
+
+void main() {
+  group('Hero Animation Tests', () {
+    testWidgets('Size of Container changes on Tap', (tester) async {
+      await tester.pumpWidget(createHeroAnimationDemoScreen());
+
+      // Get the initial Container.
+      final initialContainer =
```
RedBrogdon, line 19: This isn't actually getting the size of the Container. It's getting the constraints the Container will use when sizing its child, which is a slightly different property. See `expand_card_test.dart` for an example of how to get a widget's size during testing.

animations/test/misc/hero_animation_test.dart
```diff
@@ -0,0 +1,109 @@
...
+      // Calculate final area of Container
+      var finalArea = finalHeight * finalWidth;
+
+      expect(
+        initialArea,
+        lessThan(finalArea),
+      );
+    });
+
+    testWidgets('Color of Container changes on Tap', (tester) async {
```
RedBrogdon, line 61: This test isn't just checking that the color changes. It's checking that it starts as a specific color and ends up being another specific color. Why not try simply comparing the before and after colors to see if they match?

RedBrogdon (changes requested):

animations/test/misc/hero_animation_test.dart
arteevraina: Yes sure, fixing this in the next commit.

animations/test/misc/hero_animation_test.dart
arteevraina: Fixed this in new commit.

arteevraina:
@RedBrogdon Made the changes. Let me know if there are still more changes to be done.

RedBrogdon (approved):
LGTM. Thanks for a good PR!

PR #627, opened 2020-12-17
flutter pub update & various fixes

domesticmouse:


RedBrogdon (approved):
LGTM with one exception: some of the older projects don't have `.last_build_id` in their `/ios/.gitignore` files. It's a new file, and while the Flutter tool will create it, it doesn't go back and update anyone's .gitignore files for fear they've been hand-modified in the meantime.

Should be a quick fix, and it'll keep three unnecessary files out of the repo.

PR #629, opened 2020-12-17
use source script.sh instead of ./script.sh

johnpryan:
This allows the exit code to be returned to GitHub Actions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: subosito/flutter-action@v1
         with:
           channel: ${{ matrix.flutter_version }}
-      - run: source tool/flutter_ci_script.sh
+      - run: ./tool/flutter_ci_script.sh
   android-build:
     runs-on: ubuntu-latest
     strategy:
@@ -40,7 +40,7 @@ jobs:
       - uses: subosito/flutter-action@v1
         with:
           channel: stable
-      - run: source tool/android_ci_script.sh
+      - run: ./tool/android_ci_script.sh
   ios-build:
     runs-on: macos-latest
     strategy:
@@ -53,4 +53,4 @@ jobs:
       - uses: subosito/flutter-action@v1
         with:
           channel: stable
-      - run: source tool/ios_ci_script.sh
+      - run: ./tool/ios_ci_script.sh
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - flutter_version: dev
-            allow_failure: true
+            allow_failure: false
           - flutter_version: beta
-            allow_failure: true
+            allow_failure: false
           - flutter_version: stable
             allow_failure: false
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: subosito/flutter-action@v1
         with:
           channel: ${{ matrix.flutter_version }}
-      - run: ./tool/flutter_ci_script.sh
+      - run: source tool/flutter_ci_script.sh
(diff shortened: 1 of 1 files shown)


johnpryan:
@domesticmouse @RedBrogdon I think we should remove the `allow_failures` from dev and beta, and rather reconfigure this repos branch protection rules to ignore these checks:

Screen Shot 2020-12-17 at 10 07 21 AM

That way, we will see when `beta` and `dev` fail, but will be able to still merge pull requests.

RedBrogdon:
(I should have commented here earlier, rather than offline.)

This is an option, and at one point it's how I had things working (though I wasn't using `source` to run the scripts, just leaving out the continue-on-error flag, IIRC).

The advantage here seems to be that you see the failure right away, and the disadvantage is that we'd need to get used to accepting PRs with failing tests (since there will be times that we can't get code to work across all the channels). We could always put something in the name to call them out as optional or allowed to fail.

The downside of leaving it as it is now is that we really won't see dev/beta failures without effort. That's kind of the case right now, though, since we were previously required to click all the way through to travis.org in order to see which jobs failed.

At any rate, I'm not particularly wedded to either approach. The main thing to me is that we all agree on one or the other.

domesticmouse:
Personally, I prefer the option where we surface breaks in the Git PR UI and explictly accept them, as opposed to hiding them deep in Github Actions UI. The first approach means that I won't be blindsided when beta becomes stable and everything catches on fire.

RedBrogdon:
We should get this closed out. It sounds like both of you are in favor of surfacing the failure right in the GitHub UI, so let's go that way.

I added a couple of inline comments, but nothing major. Let me know if either of you don't have the ability to change the branch security settings, since that's something we'd want to fix.

.github/workflows/main.yml
```diff
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - flutter_version: dev
-            allow_failure: true
+            allow_failure: false
```
RedBrogdon, line 20: With all three set to false, there's no need for this matrix variable any more. You can just take it out.

.github/workflows/main.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - uses: subosito/flutter-action@v1
         with:
           channel: ${{ matrix.flutter_version }}
-      - run: ./tool/flutter_ci_script.sh
+      - run: source tool/flutter_ci_script.sh
         continue-on-error: ${{ matrix.allow_failure }}
```
RedBrogdon, line 34: I believe continue-on-error defaults to false, so this line can be taken out.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

.github/workflows/main.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - uses: subosito/flutter-action@v1
         with:
           channel: ${{ matrix.flutter_version }}
-      - run: ./tool/flutter_ci_script.sh
+      - run: source tool/flutter_ci_script.sh
```
RedBrogdon, line 33: Can you give me a little detail on this particular change? I know `source` changes whether the script is run in a new process or not, but beyond that I don't know the effect.

johnpryan: If I remember correctly, I had to use `source` to get Actions to recognize the non-zero exit code as a failure. 

Using`./` runs a script in a separate shell, whereas `source` runs the script in the current shell session, so I was able to work around it using `source`. 

I think this may have been an issue with Actions, so I reverted this change in 8f4a365 and the failures on dev and beta seem to be reporting correctly now.

RedBrogdon (approved):

RedBrogdon:
Initially I wasn't all that enthused about having the red X's on every PR when dev/beta is failing, but now that I've had more time to think, I agree with y'all's reasoning.

PR #632, opened 2020-12-19
Add a FlutterView level integration example on Android for add-to-app

xster:
Flutter cells (that appear one at a time) in an Android RecyclerView. 

Video (with don't keep activities):
demo

Currently built on stable and serves the less ambitious role of answering some challenges when accomplishing https://github.com/flutter/flutter/issues/21847. 

Though this paves the way to serve as a second role soon of demonstrating multi-engine integration once we finish work on https://github.com/flutter/flutter/issues/72009. At which point, we'll add an iOS version (which doesn't have the base challenge similar to https://github.com/flutter/flutter/issues/21847 since FlutterViewController which can integrate at view level is already more self-contained) but also make cells appear more often than at most once per screen.


add_to_app/android_view/app/src/androidTest/java/dev/flutter/example/androidView/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
+package dev.flutter.example.androidView
```
gaaclarke, line 1: probably not needed, right?

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/FlutterViewEngine.kt
```diff
@@ -0,0 +1,151 @@
...
+ * you must decide when the FlutterView is active based on your own application's
+ * intent.
+ */
+class FlutterViewEngine(val engine: FlutterEngine) : LifecycleObserver{
+    private var flutterView: FlutterView? = null
+    private var activity: ComponentActivity? = null
+    private var platformPlugin: PlatformPlugin? = null
+
+    // This is the intersection of an available activity and of a visible
+    // FlutterView. This is where Flutter would start rendering.
```
gaaclarke, line 48: Is the idiomatic docstring format for kotlin?  This docstring doesn't exactly say what this method does.  It more describes what it is.

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/FlutterViewEngine.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    fun attachFlutterView(flutterView: FlutterView) {
+        this.flutterView = flutterView
+        if (activity != null) {
+            hookActivityAndView()
+        }
+    }
+
+    fun detachFlutterView() {
```
gaaclarke, line 103: For sample code should we have full docstring documentation?

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/ListAdapter.kt
```diff
@@ -0,0 +1,94 @@
...
+import android.view.View
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import dev.flutter.example.androidView.databinding.AndroidCardBinding
+import io.flutter.embedding.android.FlutterView
+import io.flutter.plugin.common.MethodChannel
+import java.util.*
+import kotlin.random.Random
+
+class ListAdapter(private val context: Context, private val flutterViewEngine: FlutterViewEngine) : RecyclerView.Adapter<ListAdapter.Cell>() {
```
gaaclarke, line 19: Probably should have a class docstring.

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/ListAdapter.kt
```diff
@@ -0,0 +1,94 @@
...
+    override fun onBindViewHolder(cell: Cell, position: Int) {
+        // While scrolling forward, if no Flutter is presently showing, let the next one have a 1/3
+        // chance of being Flutter.
+        //
+        // While scrolling backward, let it be deterministic, and only show cells that were
+        // previously Flutter cells as Flutter cells.
+        if (previousFlutterCells.contains(position)
+            || ((previousFlutterCells.isEmpty() || position > previousFlutterCells.last())
+                && flutterCell == null
+                && random.nextInt(3) == 0)) {
```
gaaclarke, line 51: You could be more deterministic here by using the position with a prime number: `(position * prime) % 3 == 0`

gaaclarke (approved):
Looks good to me.  Seems like someone could really find FlutterViewEngine really useful.  I think we should make sure the docstrings are better fleshed out.  All classes should have one, they should be in the idiomatic format and we should flesh out the ones for the methods by making sure we are defining what the method is does.

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/ListAdapter.kt
xster: TIL Kotlin docs now does markdown and completely did away with the half tags half html stuff from javadocs. This is a huge improvement!

add_to_app/android_view/app/src/androidTest/java/dev/flutter/example/androidView/ExampleInstrumentedTest.kt
xster: ya, removed

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/FlutterViewEngine.kt
xster: it's a private method. But for consistency, I made everything use the kdoc format (which provides nice highlighting etc)

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/FlutterViewEngine.kt
xster: done

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/ListAdapter.kt
xster: done

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/ListAdapter.kt
xster: ya, it's a bit arbitrary, but I'm attempting to add a bit of immersion too to make it seem a bit like it could be server directed and not entirely pre-scripted.

xster:
> Seems like someone could really find FlutterViewEngine really useful.

Agree. There's also some challenges trying to generalize it for everyone since there are some app-specific decisions to be made. This being a v2 fix beyond docs is ok for now I think.

xster:
Not too sure what I need to do for stable. Verbose log isn't on on CI. I tried running `flutter analyze` on stable and it was passing.

RedBrogdon:
Looks interesting! I mostly have nits.

add_to_app/android_view/app/src/main/java/dev/flutter/example/androidView/FlutterViewEngine.kt
```diff
@@ -0,0 +1,220 @@
...
+import android.content.Intent
+import androidx.activity.ComponentActivity
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleObserver
+import androidx.lifecycle.OnLifecycleEvent
+import io.flutter.embedding.android.FlutterView
+import io.flutter.embedding.engine.FlutterEngine
+import io.flutter.plugin.platform.PlatformPlugin
+
+/**
```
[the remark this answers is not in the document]
xster: No, it's a bespoke file that users will have to write themselves. With a lot of wisdom and thought, it's probably possible to create one that works for 90% of the cases, but we can't get to it right now.

add_to_app/flutter_module_using_plugin/lib/cell.dart
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @override
+  void didChangeAppLifecycleState(AppLifecycleState state) {
+    setState(() { appLifecycleState = state; });
+  }
+
+  // Show a random bright color.
+  Color randomLightColor() {
+    if (randomColor == null) {
```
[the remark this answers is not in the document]
xster: lol, I haven't written dart for so long 😭

add_to_app/flutter_module_using_plugin/lib/cell.dart
```diff
@@ -0,0 +1,131 @@
...
+    return Color.fromARGB(255, randomColor.nextInt(50) + 205,
+        randomColor.nextInt(50) + 205, randomColor.nextInt(50) + 205);
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      // The Flutter cells will be noticeably different (due to background color
+      // and the Flutter logo). The banner breaks immersion.
+      debugShowCheckedModeBanner: false,
```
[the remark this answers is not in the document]
xster: It looks weird. Since it's a cell that's interwoven into native. But the background is a lot bigger than the cell, so this banner shows up in the middle of nowhere.

xster:
@RedBrogdon you want to give it another pass?

add_to_app/flutter_module_using_plugin/lib/cell.dart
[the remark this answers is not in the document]
RedBrogdon: Fair point. You're right to take it out.

RedBrogdon (approved):

RedBrogdon:
> @RedBrogdon you want to give it another pass?

LGTM!

PR #633, opened 2020-12-26
[testing_app] Migrate the sample to the integration_test package

### LINKED ISSUE #622: [testing_app] Migrate to integration_test package
It has been a few days since the [`integration_test`](https://pub.dev/packages/integration_test) package was released and we should probably migrate the sample to use it.

The [`e2e`](https://pub.dev/packages/e2e) package has also been deprecated and it's recommended to use the new package instead.

So, should we just remove the existing `e2e` and `flutter_driver` tests and replace them with the new tests as the developers writing new tests should be using the new package anyway or should we keep them for a while before we remove them?

WDYT @domesticmouse @RedBrogdon?

/cc @CareF

abd99:
- [x] Migrated integration tests to use the new [integration_test](https://pub.dev/packages/integration_test) package
- [x] Removed old e2e and driver tests
- [x] Updated the testing guide

closes #622

/cc @domesticmouse @RedBrogdon @johnpryan

PS - Feel free to take your time and review this PR after you're back from the holidays.


domesticmouse (approved):
Looks good to me, but I'm deferring to @johnpryan on this one =)

RedBrogdon:
@johnpryan, can you give this a look and merge it when you have a chance?

abd99:
@johnpryan, I would also like to know why did you suggest to follow this directory structure and keep the `driver.dart` file in the `integration_test` directory instead of the `test_driver` directory? I need to migrate the [codelab](https://codelabs.developers.google.com/codelabs/flutter-app-testing/) once this PR lands and it would be really beneficial for me while doing it.

johnpryan:
It's okay to put the `driver.dart` file in the `integration_test/` directory. This is what the counter template currently does:

```
integration_test/
  app_test.dart
  driver.dart
```

In the [integration testing](https://flutter.dev/docs/testing/integration-tests) docs we recommend a `test_driver` directory, but either is okay since both files are passed to the `flutter drive` command:

```
flutter drive \
  --driver=test_driver/integration_test.dart \
  --target=integration_test/foo_test.dart \
  -d <DEVICE_ID>
```

testing_app/integration_test/driver.dart
```diff
@@ -0,0 +1,23 @@
...
+
+Future<void> main() {
+  return integrationDriver(
+    responseDataCallback: (data) async {
+      // If the tests reported any data, save it to the disk.
+      if (data != null) {
+        for (var entry in data.entries) {
+          print('Writing ${entry.key} to the disk.');
+          // Default storage destination is the 'build' directory.
+          await writeResponseData(
```
johnpryan, line 15: The docs say that the default `responseDataCallback` is `writeResponseData` - is there a reason you are overriding it to write separate files?

testing_app/README.md
```diff
@@ -53,9 +48,9 @@ The Flutter SDK can run unit tests and widget tests in a virtual machine, withou
 - Open `coverage/index/index.html` in your preferred browser.
     
 ### CI/CD
-- Refer [.travis.yml](../.travis.yml) and the [tool](../tool) directory to see how to test Flutter projects using Travis-CI.
+- Refer [.github](../.github) and the [tool](../tool) directory to see how to test Flutter projects using GitHub Actions.
 
-Note that we aren't performing Flutter Driver tests using the Travis tool in this repo. That is because it's recommended to use physical devices to run Driver tests. You can use [Firebase Test Lab](https://firebase.google.com/docs/test-lab), [Codemagic](https://codemagic.io/) or any platform of your choice to do that.
+Note that we aren't performing Flutter Integration tests using the GitHub Actions in this repo. That is because it's recommended to use physical devices to run Driver tests. You can use [Firebase Test Lab](https://firebase.google.com/docs/test-lab), [Codemagic](https://codemagic.io/) or any platform of your choice to do that.
```
johnpryan, line 53: Technically, Firebase Test Lab doesn't use Flutter Driver. It uses the integration_test package:

```suggestion
Note that tools like GitHub Actions can't run tests on a physical device, which is required to run integration tests. Instead, you can use [Firebase Test Lab](https://firebase.google.com/docs/test-lab), [Codemagic](https://docs.codemagic.io/testing/aws/) or any platform of your choice to do that.
```

johnpryan (changes requested):

testing_app/integration_test/driver.dart
abd99: For the performance tests, I have recorded the timeline as well as its summary. If I don't manually separate the data into different files, it gets stored into one single file which becomes huge. My current approach also allows the timeline to be opened using Chrome's tracing tools to further evaluate the performance.

abd99:
> In the [integration testing](https://flutter.dev/docs/testing/integration-tests) docs we recommend a `test_driver` directory, but either is okay since both files are passed to the `flutter drive` command:

Great! I knew that it works both ways, but didn't know if there was a reason behind following one way or the other.
Thanks for the explanation!

testing_app/integration_test/driver.dart
johnpryan: You should be able to load a trace into chrome without any additional code here. @dnfield does this sound correct?

dnfield: The raw trace should be loadable in `chrome://tracing`.

abd99: For some reason, it isn't working for me.
Here's what I did to my current [integration-test](https://github.com/abd99/samples/tree/integration-test) branch:

1. Removed everything passed to `integrationDriver()` and kept everything else as it is.
2. Tried running the perf_test and then loading the `integration_response_data.json` file into `chrome://tracing`. 

The file was loaded, but no charts or data appeared.

I also tried removing `watchPerformance()` to make the reported data contain only timelines and not timeline summaries.

abd99: I think the tracing tools aren't able to load multiple timelines at once. Am I missing something here?

johnpryan: @abd99 Can you file an issue in the Flutter issue tracker?

abd99: Sure. I'll create one shortly.

johnpryan: Thanks

johnpryan (approved):

PR #634, opened 2020-12-29
isolate_example: Update filenames in README.md

mgalgs:
Some files were renamed in f325e685d148
but the rename never made it to the README.


RedBrogdon (approved):
LGTM. Nice catch!

Issue #546, opened 2020-09-15
[VeggieSeason] We should add a exit dialog in veggieseason app.


mohak852:
[VeggieSeason] In this app we should add an exit dialog in the homepage.

@abd99 @AyushBherwani1998 @RedBrogdon

mohak852:
@abd99 @AyushBherwani1998 Have you go through this issue.

RedBrogdon:
Can you give us a little more info on your rationale? I don't know that we have exit dialogs in any of our other samples, for instance.

mohak852:
![maxresdefault](https://user-images.githubusercontent.com/58987025/93591941-e80d8200-f9ce-11ea-8b9c-93f4e1458de8.jpg)
Like this Dialog as shown in this picture If we want to exit the app, we should have an alert dialog that confirms that are you sure you want to exit the app or not

mohak852:
@RedBrogdon Should I go with the Dialog Box.

RedBrogdon:
I think that showing a dialog box could be a good thing. It's the idea of an exit button that's giving me pause. Is there another location in the app where it would make sense to add a dialog?

mohak852:
No sir i think the homepage will be fine Because if we implement this inside any other page it will not make any sense. @RedBrogdon

RedBrogdon:
Sorry, what I meant was, is there another type of dialog that might make sense for the app? A dialog to confirm some other choice?

kjain333:
Respected Contributors,
It's been few months since I have contributing to flutter applications and now wanted to dive deeper into it by contributing to open source projects. I found this issue really interesting to begin with but since it's my first time here I just wanted some guidance if this issue will be fine to begin with and a little more about how the checks are performed before merging any pull request.

tusharojha:
Hi @kjain333 <img src="https://user-images.githubusercontent.com/1303154/88677602-1635ba80-d120-11ea-84d8-d263ba5fc3c0.gif" width="24px" alt="hi">
Thanks for your excitement in contributing to Flutter Samples. Also, Flutter Community always respects the new contributors.

Regarding this issue, @RedBrogdon is trying to say that we need a perfect use-case for implementing a bottom sheet from a user's perspective. So, I would suggest using the app and find the right slot for a bottom sheet.

Also regarding the checks in pull requests, the following things you need to keep in mind
- You have signed Google's CLA (for this google bot will help you after creating your first pr)
- Travis CI Build (tip: make sure to format all your dart files)
[...]

PR #639, opened 2021-01-06
[VeggieSeasons] added restore defaults feature

tusharojha:
## Description

Added the `Restore Defaults` option in the settings menu.  The idea is to incorporate a feature that makes sense for adding a `CupertinoAlertDialog` in the app.

Related Issue #546 

Screenshot: 
WhatsApp Image 2021-01-06 at 1 35 07 PM

### DIFF
--- a/experimental/veggieseasons/lib/screens/settings.dart
+++ b/experimental/veggieseasons/lib/screens/settings.dart
@@ -23,7 +23,8 @@ class VeggieCategorySettingsScreen extends StatelessWidget {
 
   static Route<void> _routeBuilder(BuildContext context, Object argument) {
     return CupertinoPageRoute(
-      builder: (context) => VeggieCategorySettingsScreen(restorationId: 'category'),
+      builder: (context) =>
+          VeggieCategorySettingsScreen(restorationId: 'category'),
       title: 'Preferred Categories',
     );
   }
@@ -136,9 +137,10 @@ class CalorieSettingsScreen extends StatelessWidget {
                       label: cals.toString(),
                       icon: SettingsIcon(
                         icon: Styles.checkIcon,
-                        foregroundColor: snapshot.hasData && snapshot.data == cals
-                            ? CupertinoColors.activeBlue
-                            : Styles.transparentColor,
+                        foregroundColor:
+                            snapshot.hasData && snapshot.data == cals
+                                ? CupertinoColors.activeBlue
+                                : Styles.transparentColor,
                         backgroundColor: Styles.transparentColor,
                       ),
                       onPress: snapshot.hasData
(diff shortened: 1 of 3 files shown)


tusharojha:
@domesticmouse & @RedBrogdon kindly review and suggest changes if any.

experimental/veggieseasons/lib/styles.dart
```diff
@@ -216,12 +216,20 @@ abstract class Styles {
 
   static const Color iconGold = Color(0xffdba800);
 
+  static const Color iconRed = Color(0xffFF0000);
```
RedBrogdon, line 219: nit: Use lowercase letters here, since that's what the other colors are doing.

Actually, one of the colors on line 209 *isn't* doing that, so if you have a second, change that one as well :)

experimental/veggieseasons/lib/screens/settings.dart
```diff
@@ -211,6 +213,44 @@ class SettingsScreen extends StatelessWidget {
...
+        icon: Styles.resetIcon,
+      ),
+      content: SettingsNavigationIndicator(),
+      onPress: () {
+        showCupertinoDialog<void>(
+          context: context,
+          builder: (context) => CupertinoAlertDialog(
+            title: Text('Are you sure?'),
+            content: Text(
+              'Are you sure you want to reset the current preferences?',
```
RedBrogdon, line 231: nit: Rather than "preferences," how about "settings," since that's the word used in the title for this screen?

RedBrogdon (commented):
I like this idea. It definitely seems like a realistic way to work a dialog into the application.

@LongCatIsLooong or @xster, would either of you mind giving this a quick look to make sure it aligns with iOS UI guidelines? I don't know, for example, whether it's common to have a settings item act like a button this way.

experimental/veggieseasons/lib/styles.dart
tusharojha: Okay, changed it and also at line 249 :)

experimental/veggieseasons/lib/screens/settings.dart
tusharojha: sounds good

LongCatIsLooong (approved):
The `CupertinoAlertDialog` looks good to me. Thank you!

experimental/veggieseasons/lib/styles.dart
```diff
@@ -216,12 +216,20 @@ abstract class Styles {
 
   static const Color iconGold = Color(0xffdba800);
 
+  static const Color iconRed = Color(0xffff0000);
```
domesticmouse, line 219: Using a pure red for Apple is probably wrong. According to Apple's [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/) `systemRed` is R 255, G 59, B 48 on a light background, and R 255, G 69 and B 58 on a dark background.

domesticmouse (approved):
Do we expose Apple's `UIColor` system colours somewhere? We should probably be using those, not hard coded constants...

experimental/veggieseasons/lib/styles.dart
tusharojha: yes, I just changed it!

domesticmouse (approved):

domesticmouse:
PTAL @RedBrogdon

RedBrogdon (approved):

PR #640, opened 2021-01-07
Replaced deprecated property with new one

Alabhya268:
In showDialog(), child property is deprecated and builder should be used instead.


RedBrogdon (approved):
Nice catch, @Alabhya268. 👍

Issue #635, opened 2020-12-29
Migrate desktop_photo_search to file_selector


stuartmorgan:
I've published [unendorsed implementations of the new `file_selector` plugin](https://github.com/google/flutter-desktop-embedding/tree/master/plugins/file_selector), which is the flutter/plugins replacement for FDE's file_chooser, and am therefore discontinuing file_chooser. The sample should be migrated to that new plugin.

PR #645, opened 2021-01-11
Migrate to file_selector

domesticmouse:
fixes: https://github.com/flutter/samples/issues/635

Stuart, can you please review? Thanks!


experimental/desktop_photo_search/lib/main.dart
```diff
@@ -101,20 +102,21 @@ class UnsplashHomePage extends StatelessWidget {
...
-                          if (!result.canceled) {
-                            final bytes =
-                                await photoSearchModel.download(photo: photo);
-                            await File(result.paths[0]).writeAsBytes(bytes);
-                          }
+                          final fileData =
+                              await photoSearchModel.download(photo: photo);
+                          final photoFile =
+                              XFile.fromData(fileData, mimeType: 'image/jpeg');
+                          photoFile.saveTo(path);
```
stuartmorgan, line 119: What happens if the user cancels the save dialog? You've removed the logic that handled that.

domesticmouse: It wasn't obvious from the types that the returned string was nullable, and that is how file save cancellation is now being communicated. This will be better when null safety is a thing.

stuartmorgan: > It wasn't obvious from the types that the returned string was nullable, and that is how file save cancellation is now being communicated.

Yes, file_chooser had an explicit cancel status since I thought the added clarity was worth the slightly more complex return type, but 🤷

stuartmorgan (approved):

PR #647, opened 2021-01-15
Enable State Restoration for iOS

goderbauer:
Feature was implemented in https://github.com/flutter/engine/pull/23495.

This enables it for the veggie seasons app according to the instructions to be published in https://github.com/flutter/flutter/pull/73585.

To test state restoration on iOS:

1. Open `ios/Runner.xcworkspace/` in xcode.
2. (iOS 14+ only): Switch to build in profile or release mode (launching an app from the home screen is not supported in debug mode), see screenshots in https://github.com/flutter/flutter/issues/51673#issuecomment-597215128.
2. Press the Play button in xcode to build and run the app.
3. Create some in-memory state in the app on the phone, e.g. by navigating to a different screen.
4. Background the app on the phone, e.g. by goign back to the home screen.
5. Press the Stop button in xcode to terminate the app while running in the background.
6. Open the app again on the phone (not via xcode), it will restart and restore its state.

### DIFF
--- a/experimental/veggieseasons/ios/Runner/Base.lproj/Main.storyboard
+++ b/experimental/veggieseasons/ios/Runner/Base.lproj/Main.storyboard
@@ -8,7 +8,7 @@
         <!--Flutter View Controller-->
         <scene sceneID="tne-QT-ifu">
             <objects>
-                <viewController id="BYZ-38-t0r" customClass="FlutterViewController" sceneMemberID="viewController">
+                <viewController restorationIdentifier="main" id="BYZ-38-t0r" customClass="FlutterViewController" sceneMemberID="viewController">
                     <layoutGuides>
                         <viewControllerLayoutGuide type="top" id="y3c-jy-aDJ"/>
                         <viewControllerLayoutGuide type="bottom" id="wfy-db-euE"/>


shihaohong:
I tested the above workflow on the `gallery` app. 

It worked on iOS simulators, but when I try it on an iOS 14.3 device, I get the following screen after on step 6: 



I am not able to test on an iOS device <14 since I do not have a physical device on me, but I assume it should work there. 

On the other hand, I also found the steps in [this article](http://pinkstone.co.uk/how-to-kill-your-app-when-it-enters-background-mode/) seem to work on a physical iOS 14+ device. Basically, there's a bool to kill an app the second it's backgrounded, similar to Android. However, this modifies the app's default behavior when it is backgrounded instead of the target device's behavior for all apps, so this runs the risk of the developer forgetting to flip the bool or removing it altogether once they're done testing state restoration on iOS.

I think that your suggestion is safer for Flutter devs, but maybe we should recommend testing state restoration on iOS simulators instead of physical devices?

goderbauer:
Alternatively, you could just run the app in profile or debug mode to test it.

shihaohong (approved):
LGTM

PR #648, opened 2021-01-15
Fix restoration of search page

goderbauer:
Restoration of the search page broke in https://github.com/flutter/samples/pull/433/commits/c7d3f1a55a7606d5442ff4141fb1754ad469d2eb due to a bad merge. Embarrassingly, that part of the search page wasn't covered by the test, so the breakage went unnoticed at the time. This PR fixes the breakage and extends the test so it will hopefully never break again.


shihaohong (approved):
LGTM

PR #661, opened 2021-01-19
Fix generic type placement of Provider.of call

robertpiosik:
Wrong usage of Provider.of<T> may be misleading for beginners.


domesticmouse:
Thank you!

domesticmouse (approved):

Issue #649, opened 2021-01-16
Headers missing in code in  Place Tracker Files


arteevraina:
As mentioned in [guidelines](https://github.com/flutter/samples/blob/master/CONTRIBUTING.md) that all files in the project must start with the following header.

```
// Copyright 2020 The Flutter team. All rights reserved.
// Use of this source code is governed by a BSD-style license that can be
// found in the LICENSE file.
```

This is missing in [place_tracker](https://github.com/flutter/samples/tree/master/place_tracker) sample app. If this is not done on some specific purpose. I would like to fix this up.

WDYT @RedBrogdon @domesticmouse ?

ditman:
> If this is not done on some specific purpose. I would like to fix this up.

Yeeet! No, these were probably forgotten by mistake. (there might be other apps in the repo with the same issue).

@arteevraina if you want to contribute a patch, it'll be very welcome, for sure!

johnpryan:
That would be great. We might want to also add some [tooling](https://pub.dev/packages/copywriter) to check this too.

arteevraina:
> > If this is not done on some specific purpose. I would like to fix this up.
> 
> Yeeet! No, these were probably forgotten by mistake. (there might be other apps in the repo with the same issue).
> 
> @arteevraina if you want to contribute a patch, it'll be very welcome, for sure!

Yes, sure thing. I will make a PR for this. After rechecking I found out that this header was missing in the files of the following repos- 
* [web](https://github.com/flutter/samples/tree/master/web).
* [platform-design](https://github.com/flutter/samples/tree/master/platform_design).
* [platform-channel](https://github.com/flutter/samples/tree/master/platform_channels).
* [ios_app_clip (only one file)](https://github.com/flutter/samples/blob/master/ios_app_clip/test/widget_test.dart).

arteevraina:
> That would be great. We might want to also add some [tooling](https://pub.dev/packages/copywriter) to check this using GH Actions.

Yes, I am always open to contributing to the community.

arteevraina:
> That would be great. We might want to also add some [tooling](https://pub.dev/packages/copywriter) to check this using GH Actions.

Can I create a separate issue for this and start working on it?

ditman:
> Can I create a separate issue for this and start working on it?

Yes, that'd be the best way to go about it. That way we can close this issue with the PR you sent fixing headers, without waiting for the tooling fix :)

PR #664, opened 2021-01-20
fix: added headers

arteevraina:
Fixes: #649 

I have added the headers that were missing in the files. Also, most of the apps in [web](https://github.com/flutter/samples/tree/master/web) are having their own licenses. So, I have not added this header in the files of those apps. Let me know if is is required I will fix that up also. 

Please review and suggest changes @ditman @johnpryan :)


ditman (approved):
In some of these the copyright should be probably older, but I think this is probably fine.

johnpryan (approved):

ditman:
Again, thanks for the contribution @arteevraina!

PR #668, opened 2021-01-22
Migrate to the New Material Buttons

domesticmouse:
Using [flutter.dev/go/material-button-migration-guide](https://flutter.dev/go/material-button-migration-guide)

PR #662, opened 2021-01-19
Bump ndk

goderbauer:
To fix build issues uncovered in https://github.com/flutter/samples/pull/648 and https://github.com/flutter/samples/pull/647.

This is likely only a short-term fix until the NDK version changes again...

PR #669, opened 2021-01-22
Revert "Bump ndk"

domesticmouse:
Reverts flutter/samples#662

Looks like the Github Action platform has reverted the available NDK version: https://github.com/flutter/samples/pull/668/checks?check_run_id=1746329483#step:5:418


goderbauer (approved):
lol, we should find a way to stabilize this. Can we pin the version that's available or something like that?

LGTM

domesticmouse:
@godofredoc I don't see the ability to pin GitHub actions to a specific revision.

PR #670, opened 2021-01-23
Added sample code for multiple flutters on iOS

gaaclarke:
issue: https://github.com/flutter/flutter/issues/74530


gaaclarke:
@xster I pushed this up before the weekend in case you want to start looking at this for android.  Things are looking good now, I have the first screen done with data sharing between flutter and the host.  I was thinking we can make the app infinite, you can push as many views as you want.  We can even keep a visible count and see how far we can go before the app is booted =)

add_to_app/multiple_flutters_module/lib/main.dart
```diff
@@ -0,0 +1,94 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+void main() => runApp(MyApp(Colors.blue));
+
+void topMain() => runApp(MyApp(Colors.green));
```
xster, line 6: These need @pragma('vm:entry-point')

add_to_app/multiple_flutters_module/lib/main.dart
```diff
@@ -0,0 +1,94 @@
...
+
+void bottemMain() => runApp(MyApp(Colors.purple));
```
xster, line 8: typo

gaaclarke:
@xster I finished up all the technical work, I just have to clean it up.  It ended up being really slick seeing the add button update both instances of flutter and popping and seeing the shared state.  I think following this pattern will handle a lot of peoples data sharing needs.  It requires a bit of discipline though, where the STM would hopefully make it easier.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+import Flutter
+
+class SingleFlutterViewController : FlutterViewController, DataModelObserver {
+  var _channel : FlutterMethodChannel?
```
xster, line 5: this is one area kotlin does so much better. Though here this can still be 
`private lazy var channel: FlutterMethodChannel = { FlutterMethodChannel(name: "multiple-flutters", binaryMessenger: self.engine!.binaryMessenger) }` so it's just non-null

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+    DataModel.shared.removeObserver(observer: self)
+  }
+  
+  required init(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+  
+  func onCountUpdate(newCount: Int64) {
+    if let channel = _channel {
+      channel.invokeMethod("setCount", arguments: newCount);
```
xster, line 25: isn't this circular with the `count` setter below?

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+  
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    _channel = FlutterMethodChannel(name: "multiple-flutters", binaryMessenger: self.engine!.binaryMessenger)
+    _channel!.invokeMethod("setCount", arguments: DataModel.shared.count)
```
xster, line 32: why not do these 2 steps in the init too?

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+    _channel!.setMethodCallHandler { (call : FlutterMethodCall, result : @escaping FlutterResult) in
+      if (call.method == "incrementCount") {
+        DataModel.shared.count = DataModel.shared.count + 1
+        result(nil)
+      } else if (call.method == "next") {
+        let storyboard = UIStoryboard(name: "Main", bundle: nil)
+        let vc = storyboard.instantiateViewController(withIdentifier: "Count")
```
xster, line 39: Can we call this something else? Like "nativeViewControllerWithCount" or something?

add_to_app/MultipleFluttersIos/Pods/Pods.xcodeproj/xcuserdata/aaclarke.xcuserdatad/xcschemes/Flutter.xcscheme
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
xster, line 1: gitignore

add_to_app/MultipleFluttersIos/.gitignore
```diff
@@ -0,0 +1 @@
+MultipleFluttersIos.xcworkspace/xcuserdata
```
xster, line 1: be consistent in the folder name with the other samples (e.g. snake notation)

add_to_app/multiple_flutters_module/README.md
```diff
@@ -0,0 +1,11 @@
+# multiple_flutters_module
+
+A new flutter module project.
+
+## Getting Started
+
+For help getting started with Flutter, view our online
```
xster, line 7: add some description

add_to_app/multiple_flutters_module/lib/main.dart
```diff
@@ -0,0 +1,93 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+@pragma('vm:entry-point')
```
xster, line 4: you don't need it for main

add_to_app/multiple_flutters_module/lib/main.dart
```diff
@@ -0,0 +1,93 @@
...
+          _counter = call.arguments as int;
+        });
+      } else {
+        throw Exception('not implemented ${call.method}');
+      }
+    });
+  }
+
+  void _incrementCounter() async {
+    _channel.invokeMethod("incrementCount", _counter);
```
xster, line 58: oh I see. This is a bit fragile. I'd just have the same setters on both sides whether being set locally or from the other side. Just have the same exit conditions on the setter. If being set to the same, just drop out. So no one has to worry about this implicit agreement and accidentally break it while refactoring.

add_to_app/multiple_flutters_module/test/widget_test.dart
```diff
@@ -0,0 +1,30 @@
...
+  testWidgets('Counter increments smoke test', (WidgetTester tester) async {
+    // Build our app and trigger a frame.
+    await tester.pumpWidget(MyApp());
+
+    // Verify that our counter starts at 0.
+    expect(find.text('0'), findsOneWidget);
+    expect(find.text('1'), findsNothing);
+
+    // Tap the '+' icon and trigger a frame.
+    await tester.tap(find.byIcon(Icons.add));
```
xster, line 23: This shouldn't pass anymore right? Though if you do what I'm suggesting, then it'll pass again :)

add_to_app/MultipleFluttersIos/MultipleFluttersIos.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
+</dict>
+</plist>
```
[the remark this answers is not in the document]
gaaclarke: It didn't in my case.  Maybe because I made the project in a git repository and I had the checkmark off for adding a git repo.  Moved all of them and updated git ignores.

add_to_app/MultipleFluttersIos/MultipleFluttersIos.xcodeproj/xcuserdata/aaclarke.xcuserdatad/xcschemes/xcschememanagement.plist
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
gaaclarke: done

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+
+
+protocol DataModelObserver : AnyObject {
```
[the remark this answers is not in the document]
gaaclarke: oldest trick in the book!  https://en.wikipedia.org/wiki/Design_Patterns

gaaclarke:
@xster Thanks for the review, that will help me get moving quicker this morning.  The one last thing I want to do is add a warning that the performance of release builds is different.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,47 @@
...
+  func onCountUpdate(newCount : Int64)
+}
+
+struct AnyDataModelObserver {
+  weak var base: DataModelObserver?
```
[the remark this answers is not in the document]
gaaclarke: You can't use `let` here.  Weak variables have to be be `var` because they get nil'd out after the last object is gone.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,47 @@
...
+class DataModel {
+  private var _count : Int64 = 0
+  var count : Int64 {
+    get {
+      return self._count
+    }
+    set {
+      self._count = newValue
+      for observer in _observers {
+        if let base = observer.base {
```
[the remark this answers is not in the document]
gaaclarke: This tripped me up for a long time and found the pattern described elsewhere.  They have to be weak references, so you cant create them directly.  The other key is you can't make weak references to protocols directly, they have to inherit from `AnyObject`.

add_to_app/MultipleFluttersIos/MultipleFluttersIos.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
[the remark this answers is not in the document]
xster: Ah, could be the case. I vaguely remember this happening to me too.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DoubleFlutterViewController.swift
```diff
@@ -0,0 +1,18 @@
+import UIKit
+
+class DoubleFlutterViewController : UIViewController {
```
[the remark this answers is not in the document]
xster: I don't mean specifically in this arrangement either, but we should find some way to show this off a bit more. Since we can handle something like 5 Flutters in a screen.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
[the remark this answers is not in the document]
gaaclarke: There is also some weird stuff about protocols not being interchangeable with objects.  For example, you can't use generics with protocols.  It's tricky and my swift-fu isn't that strong for such esoteric matters.  I tried to replace AnyDataModelObserver with a generic but I couldn't figure it out.

xster:
2 more thoughts. You built something fairly powerful in terms of sync'ing. Might be nice to let it show the data binding with something more realistic than just an int. Like a name/email/date picker or something than just a single int. Also, when writing the README, don't make it sound like it's the only, permanent arrangement for people who might not want to route everything through the platform. Say it's one of the viable options and more might come without platform coordination once https://github.com/flutter/flutter/issues/72030 is node.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,47 @@
...
+
+
+protocol DataModelObserver : AnyObject {
+  func onCountUpdate(newCount : Int64)
+}
+
+struct AnyDataModelObserver {
+  weak var base: DataModelObserver?
+  
+  init(_ base: DataModelObserver ) {
```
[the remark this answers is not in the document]
gaaclarke: Just nomenclature I saw used by the person that described the pattern.  I don't know if I can come up with a better name.  If I call it "observer" like it is, you might see code like `observer.observer.foo()` `observer.base.foo()` looks a bit better and communicates the wrapper struct a bit better.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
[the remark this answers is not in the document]
xster: Can we call this something more descriptive than base perhaps? And add some comments for why this indirection exists.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DoubleFlutterViewController.swift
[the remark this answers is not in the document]
gaaclarke: That isn't what we discussed doing.  It doesn't really provide a sample of something that doesn't exist and as the screens get smaller you won't even be able to press the buttons.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
gaaclarke: Yea, it's a pain but since the initialization has a dependency on `self.engine` I think this is better to have explicit initialization instead of lazy initialization.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
gaaclarke: I don't think so, I don't see it and it doesn't deadlock.  What cycle were you thinking?  This tells dart to update its count, there isn't a message that comes back from dart.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
gaaclarke: This is slightly more efficient, doing all the flutter bindings only when we are actually going to view something.  Subscribing to the datamodel isn't as heavy but it could be moved here too.  Updates are ignored anyways until we have a view.

add_to_app/MultipleFluttersIos/Pods/Pods.xcodeproj/xcuserdata/aaclarke.xcuserdatad/xcschemes/Flutter.xcscheme
gaaclarke: done

add_to_app/MultipleFluttersIos/.gitignore
gaaclarke: I don't believe that is idiomatic rules for naming iOS projects.  I've never seen it in my 10 years of iOS development.

add_to_app/multiple_flutters_module/lib/main.dart
gaaclarke: Done

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DoubleFlutterViewController.swift
```diff
@@ -0,0 +1,18 @@
...
+  let _topFlutter : SingleFlutterViewController = SingleFlutterViewController(withEntrypoint: "topMain")
+  let _bottomFlutter : SingleFlutterViewController = SingleFlutterViewController(withEntrypoint: "bottomMain")
+
+  override func viewDidLoad() {
+    addChild(_topFlutter)
+    addChild(_bottomFlutter)
+    let halfHeight = self.view.bounds.height/2.0
+    _topFlutter.view.frame = CGRect(x: 0, y: 0, width: self.view.bounds.width, height: halfHeight);
```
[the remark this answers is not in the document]
gaaclarke: Done

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,47 @@
...
+    set {
+      self._count = newValue
+      for observer in _observers {
+        if let base = observer.base {
+          base.onCountUpdate(newCount: self._count)
+        }
+      }
+    }
+  }
+  var _observers: [AnyDataModelObserver] = []
```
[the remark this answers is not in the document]
gaaclarke: Done, you do need it if you are writing your own getter and setters: https://stackoverflow.com/a/24215191  I was completely inconsistent here.

add_to_app/MultipleFluttersIos/.gitignore
xster: Screen Shot 2021-01-25 at 10 32 04 AM

Sure I agree, but we don't want this to be the only folder with a different name either since there are many platforms involved.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
xster: SG

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
xster: Let's move thread. This is a subset of the comment in https://github.com/flutter/samples/pull/670#discussion_r563205815.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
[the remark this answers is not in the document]
xster: Your AnyDataModelObserver isn't an actual observer. Maybe we can fix it there? e.g. for weakObserverHolder in observerList, weakObserverHolder.observer.on...

Or make AnyDataModelObserver implement DataModelObserver and let its onCountUpdate pipe it through to the weak reference if it's still alive?

add_to_app/MultipleFluttersIos/MultipleFluttersIos/SingleFlutterViewController.swift
gaaclarke: done

add_to_app/multiple_flutters_module/lib/main.dart
gaaclarke: This is required if you want a single source of truth, the native datamodel.  If we removed this native views would stop updating and when you pop a view controller they would be out of date.  You'd have to create a whole new mechanism to avoid that.  If you were concerned about breaking this requirement you could make a handshake message.

add_to_app/multiple_flutters_module/test/widget_test.dart
gaaclarke: Making it function without a native datamodel is a non-goal.  If we wanted to update this we'd assert that it is sending messages which isn't as easy as it should be.  I'll just deleted it.

add_to_app/MultipleFluttersIos/.gitignore
gaaclarke: Okay, moved the folder, i can rename that folder fine and keep the project named idiomatically.  I got it twisted.

add_to_app/multiple_flutters_module/README.md
gaaclarke: Done

add_to_app/multiple_flutters_module/lib/main.dart
gaaclarke: I've added a description of the data synchronization mechanism in `multiple_flutters_module/README.md`

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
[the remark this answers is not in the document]
gaaclarke: We could do that.  The problem with that is that it requires more code and I don't think it is much of an improvement at the cost of more code to maintain (every method declaration would have to be written twice, wrapping the null checks won't allow clients to coalesce null checks).

I made the struct private to make sure it doesn't confuse any client of the data model, I've also added docstrings.  I think that will address most of your concerns.

add_to_app/MultipleFluttersIos/MultipleFluttersIos/DataModel.swift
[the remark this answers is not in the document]
gaaclarke: Docstring was added and the class was made private.

gaaclarke:
@xster 

> 2 more thoughts. You built something fairly powerful in terms of sync'ing. Might be nice to let it show the data binding with something more realistic than just an int. Like a name/email/date picker or something than just a single int.

I've added more details on how this mechanism works in code comments and mostly in `multiple_flutters_module/README.md`.

I think given the example and the documentation I've written they can extrapolate it to their needs.  Even with a complicated data model, someone may want to still have a single observable integer.  I guess I think having the more simple example is more useful.

> Also, when writing the README, don't make it sound like it's the only, permanent arrangement for people who might not want to route everything through the platform. Say it's one of the viable options and more might come without platform coordination once [flutter/flutter#72030](https://github.com/flutter/flutter/issues/72030) is done.

Done, check out `multiple_flutters_module/README.md`.

add_to_app/multiple_flutters_module/README.md
```diff
@@ -0,0 +1,50 @@
...
+data between the host platform and multiple instances of Flutter by combining
+the [Observer design pattern](https://en.wikipedia.org/wiki/Observer_pattern)
+and [Flutter platform
+channels](https://flutter.dev/docs/development/platform-integration/platform-channels).
+Here is how it works:
+
+* The definitive source of truth for the data lives in the host platform data
+  model.
+* Every host view displaying Flutter content maintains: a Flutter engine, a
+  bidirectional platform channel, a subscription to the host data model.
```
RedBrogdon, line 38: nit: "channel, and a subscription..."

RedBrogdon: nit: "...data they are interested, and this data is..."

add_to_app/multiple_flutters_module/README.md
```diff
@@ -0,0 +1,50 @@
...
+* Flutter instances maintain a copy of the data it is interested in reading,
+  this data is seeded by the host when the instance is first displayed.
+* Mutations from Flutter code are sent to the host platform via the channel who
```
RedBrogdon, line 41: nit: "the channel which performs the mutations, and then notifies"

add_to_app/multiple_flutters_module/lib/main.dart
```diff
@@ -0,0 +1,98 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(widget.title),
+      ),
+      body: Center(
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
```
RedBrogdon, line 75: nit: No `<Widget>` required here.

RedBrogdon (changes requested):

RedBrogdon:
@gaaclarke I have a few nits about the README and code style, but otherwise this looks like a great sample.

The only real concern I have is the directory names it's putting code in. Right now, everything in `/add-to-app` is named first by platform, then by subject or function (e.g. `ios_using_plugin`, `android_using_plugin`, `flutter_module_using_plugin`). Can you rename the directories here to match that pattern, maybe something like `ios_multiple_engines`, etc.?

xster:
proposed naming sounds reasonable

add_to_app/multiple_flutters_module/README.md
gaaclarke: Done, done.

add_to_app/multiple_flutters_module/README.md
gaaclarke: Done.

add_to_app/multiple_flutters_module/lib/main.dart
gaaclarke: Not required, but a linter error if it isn't present.  This is also how the template we provide to users does things.

gaaclarke:
> @gaaclarke I have a few nits about the README and code style, but otherwise this looks like a great sample.

Thanks!

> The only real concern I have is the directory names it's putting code in. Right now, everything in `/add-to-app` is named first by platform, then by subject or function (e.g. `ios_using_plugin`, `android_using_plugin`, `flutter_module_using_plugin`). Can you rename the directories here to match that pattern, maybe something like `ios_multiple_engines`, etc.?

The reason I did this is because `[multiple_flutters_ios, multiple_flutters_android, multiple_flutters_module]` are all part of the same example.  If they share the same prefix they will be listed together on Github and and locally.  I think having the module next to the clients is ideal.  People browsing the directory don't have to decode the naming structure, things are just next to each other.

What about this, do you want me to put all 3 of my directories in another directory here called `multiple_flutters`?  That seems even better.

add_to_app/multiple_flutters_ios/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,55 @@
...
+  func addObserver(observer : DataModelObserver) {
+    observers.append(AnyDataModelObserver(observer))
+  }
+  
+  func removeObserver(observer : DataModelObserver) {
+    observers.removeAll { (element : AnyDataModelObserver) -> Bool in
+      if let base = element.base {
+        return base === observer
+      } else {
+        return true
```
xster, line 51: add comment

add_to_app/multiple_flutters_ios/MultipleFluttersIos/ViewController.swift
```diff
@@ -0,0 +1,40 @@
+// Copyright 2021 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import UIKit
+import Flutter
+
+/// UIViewController associated with the NativeViewCount storyboard scene.
+class ViewController: UIViewController, DataModelObserver {
```
xster, line 9: NativeViewController? It's too generic.

add_to_app/multiple_flutters_ios/MultipleFluttersIos/ViewController.swift
```diff
@@ -0,0 +1,40 @@
...
+    onCountUpdate(newCount: DataModel.shared.count)
+  }
+  
+  @IBAction func onAddCount() {
+    DataModel.shared.count = DataModel.shared.count + 1
+  }
+  
+  @IBAction func onNext() {
+    let navController = self.navigationController!
+    if navController.viewControllers.count % 4 == 1 {
```
xster, line 32: add comment

add_to_app/multiple_flutters_module/lib/main.dart
xster: Ah, I see what you're trying to do. Documenting that this is just one option of doing this SG.

add_to_app/multiple_flutters_module/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Getting Started
+
+For iOS instructions, see:
+[multiple_flutters_ios](../multiple_flutters_ios/README.md).
+
+For Android instructions, see:
+[multiple_flutters_android](../multiple_flutters_android/README.md).
+
+## Flutter Engine Group
+
```
xster, line 15: Add a brief context for why might someone want to have multiple engines (probably in the parent's README).

add_to_app/multiple_flutters_module/README.md
```diff
@@ -0,0 +1,50 @@
...
+* The definitive source of truth for the data lives in the host platform data
+  model.
+* Every host view displaying Flutter content maintains: a Flutter engine, a
+  bidirectional platform channel, and a subscription to the host data model.
+* Flutter instances maintain a copy of the data they are interested in reading,
+  and this data is seeded by the host when the instance is first displayed.
+* Mutations from Flutter code are sent to the host platform via the channel. The
+  host platform performs the mutations, and then notifies all host view
+  controllers and Flutter engines of the new value.
```
xster, line 43: Describe what source the Flutter UI is following

add_to_app/README.md
```diff
@@ -223,6 +223,14 @@ Once you've understood the basics of add-to-app with `android_fullscreen` and
 `ios_fullscreen`, this is a good sample to demonstrate how to integrate Flutter
 in a slightly more realistic setting with existing business logic.
 
+### `multiple_flutters_ios` and `multiple_flutters_android (TODO)`
+
+These are examples that show how to embed multiple lightweight engines into an
```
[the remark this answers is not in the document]
gaaclarke: That isn't scaling well.  I'd like to trim that information down and put it in their respective README's.  This should be a brief overview of all the samples in this directory IMO.

add_to_app/multiple_flutters_ios/MultipleFluttersIos.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,457 @@
...
+				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
+				CODE_SIGN_STYLE = Automatic;
+				DEVELOPMENT_TEAM = 5B6X25244C;
+				INFOPLIST_FILE = MultipleFluttersIos/Info.plist;
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/Frameworks",
+				);
+				PRODUCT_BUNDLE_IDENTIFIER = dev.flutter.MultipleFluttersIos;
```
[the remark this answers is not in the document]
gaaclarke: My understanding is his concern is the name of the directory matching, I'll let him clarify.

add_to_app/multiple_flutters_ios/MultipleFluttersIos/AppDelegate.swift
```diff
@@ -0,0 +1,35 @@
...
+  func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+    #if DEBUG
+        let isDebug = true
+    #else
+        let isDebug = false
+    #endif
+    if (isDebug) {
+      NSLog("⚠️ WARNING: the memory and CPU costs for Flutter engine groups are significantly greater in debug builds.  See also: https://github.com/dart-lang/sdk/issues/36097")
+    } else {
+      NSLog("⚠️ WARNING: the memory and CPU costs for Flutter engine groups are significantly less here than in debug builds.  See also: https://github.com/dart-lang/sdk/issues/36097")
```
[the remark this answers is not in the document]
gaaclarke: What do you like better?  "NOTICE"?

add_to_app/README.md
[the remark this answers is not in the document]
gaaclarke: If I moved all 3 directories into a multiple_flutters directory that would be a good place for this formation:
```
add_to_app/
  multiple_flutters/
    README.md
    multiple_flutters_android/
    multiple_flutters_ios/
    multiple_flutters_module/
```

add_to_app/multiple_flutters_ios/MultipleFluttersIos/AppDelegate.swift
```diff
@@ -0,0 +1,35 @@
...
+
+import UIKit
+import Flutter
+
+/// The app's UIApplicationDelegate.
+@main
+class AppDelegate: UIResponder, UIApplicationDelegate {
+  let engines = FlutterEngineGroup(name: "multiple-flutters", project: nil)
+
+  func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
```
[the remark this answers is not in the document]
gaaclarke: I reformatted all the code with a formatter

add_to_app/multiple_flutters_ios/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,55 @@
+// Copyright 2021 The Flutter team. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Foundation
+
+/// A protocol that recieves updates when the datamodel is changed.
```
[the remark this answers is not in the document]
gaaclarke: done

add_to_app/multiple_flutters_ios/MultipleFluttersIos.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,457 @@
...
+			name = Release;
+		};
+		0D61B1F325BB8A2400A1A590 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			baseConfigurationReference = 5EE6997543EAE7F85C674231 /* Pods-MultipleFluttersIos.debug.xcconfig */;
+			buildSettings = {
+				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
+				CODE_SIGN_STYLE = Automatic;
+				DEVELOPMENT_TEAM = 5B6X25244C;
```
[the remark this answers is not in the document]
gaaclarke: done

add_to_app/multiple_flutters_ios/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,55 @@
...
+  func onCountUpdate(newCount : Int64)
+}
+
+/// A wrapper object around a weak reference to an object that implements DataModelObserver.
+///
+/// This is required since you can't directly hold weak references to protocols in data structures.
+private struct AnyDataModelObserver {
+  weak var base: DataModelObserver?
+  
+  init(_ base: DataModelObserver ) {
```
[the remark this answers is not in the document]
gaaclarke: reformatted the code automatically

add_to_app/multiple_flutters_ios/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,55 @@
...
+
+/// A protocol that recieves updates when the datamodel is changed.
+protocol DataModelObserver : AnyObject {
+  func onCountUpdate(newCount : Int64)
+}
+
+/// A wrapper object around a weak reference to an object that implements DataModelObserver.
+///
+/// This is required since you can't directly hold weak references to protocols in data structures.
+private struct AnyDataModelObserver {
```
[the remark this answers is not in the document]
gaaclarke: I believed it was standard swift nomenclature, see AnyObject.  To be more clear I named it DataModelObserverWrapper.

add_to_app/multiple_flutters_ios/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,55 @@
...
+    set {
+      self._count = newValue
+      for observer in observers {
+        if let base = observer.base {
+          base.onCountUpdate(newCount: self._count)
+        }
+      }
+    }
+  }
+  private var observers: [AnyDataModelObserver] = []
```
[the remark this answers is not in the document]
gaaclarke: No, that would require an immutable array.

add_to_app/multiple_flutters_ios/MultipleFluttersIos/DataModel.swift
```diff
@@ -0,0 +1,55 @@
...
+  static let shared = DataModel()
```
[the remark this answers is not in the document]
gaaclarke: I wouldn't say that.  Having a singleton observable datamodel is a valid design choice.

add_to_app/multiple_flutters_ios/MultipleFluttersIos/DataModel.swift
gaaclarke: done

add_to_app/multiple_flutters_ios/MultipleFluttersIos/ViewController.swift
gaaclarke: done

add_to_app/multiple_flutters_ios/MultipleFluttersIos/ViewController.swift
gaaclarke: Done, HostViewController.

add_to_app/multiple_flutters_module/README.md
gaaclarke: Ok, rewrote it.  It's now in multiple_flutters/README.md

add_to_app/multiple_flutters_module/README.md
gaaclarke: I added links to all the source code including main.dart with descriptions of the files.  I think this solves what you want and helps in other ways too.

add_to_app/multiple_flutters/README.md
```diff
@@ -0,0 +1,72 @@
+# multiple_flutters
```
gaaclarke, line 1: @xster I think this is more what you were looking for.  This consolidates most of the information without spamming the parent README.md.

add_to_app/multiple_flutters_ios/MultipleFluttersIos/AppDelegate.swift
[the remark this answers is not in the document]
gaaclarke: I switched it to NOTICE with a megaphone emoji

RedBrogdon:
> The reason I did this is because `[multiple_flutters_ios, multiple_flutters_android, multiple_flutters_module]` are all part of the same example. 

This is already the case for several projects in the directory. The *_using_plugin projects for example, or @xster's _books both follow a similar naming scheme. Unless there's a particularly strong reason, I'd prefer to avoid breaking that consistency for new projects.

add_to_app/multiple_flutters_ios/MultipleFluttersIos.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
RedBrogdon: +1. My concern was about the directory names. Because it's such a pain in the neck to rename a project after you've created it, I wasn't going to ask that of @gaaclarke.

gaaclarke:
> This is already the case for several projects in the directory. The *_using_plugin projects for example, or @xster's _books both follow a similar naming scheme. Unless there's a particularly strong reason, I'd prefer to avoid breaking that consistency for new projects.

@RedBrogdon Can you check out the structure now, it conforms to the naming structure in `add_to_app/`.  I also needed a place to put detailed information on the whole sample so adding `multiple_flutters/README.md` gave me a place to put all that information.  Previously it wasn't clear where you'd go to find out information about the samples in aggregate.  I think @xster  was putting some of that info in `add_to_app/README.md` but that doesn't scale for the amount of info this sample requires.

RedBrogdon:
> @RedBrogdon Can you check out the structure now, it conforms to the naming structure in `add_to_app/`.

I'm a bit confused -- is there a commit that hasn't gotten pushed yet? It looks like this PR would land the new projects in a folder within `add-to-app`, which isn't the convention we've used up until this point. That would leave the samples in an awkward state, since most of them would be organized one way, but a few wouldn't.

FWIW, I'm not wedded to any particular way of storing these projects. If there are enough of them now that it makes more sense to group them into subfolders ("books", "plugin", "fullscreen", and "multiple_flutters", maybe?), that's fine. My only concern is that we pick a convention and use it consistently.

@xster, please feel free to chime in here as well, since a lot of the `add-to-app` code is yours or was done in partnership with you.

add_to_app/multiple_flutters_ios/MultipleFluttersIos/AppDelegate.swift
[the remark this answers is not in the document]
xster: Sorry, I just meant that for debug mode, we should show a warning or something scary because the whole thing is about performance and performance is crappy on debug. 

For non-debug where it's working, it doesn't have to be a scary prompt or any messaging that sounds like they need to do anything about it.

add_to_app/multiple_flutters/README.md
```diff
@@ -0,0 +1,72 @@
...
+- [SingleFlutterViewController.swift](./multiple_flutters_ios/MultipleFluttersIos/SingleFlutterViewController.swift)
+  — A subclass of FlutterViewController that synchronizes with the DataModel.
+- [DoubleFlutterViewController.swift](./multiple_flutters_ios/MultipleFluttersIos/DoubleFlutterViewController.swift)
+  — A UIViewController that embeds multiple Flutter instances.
+
+### Android
+
+## Data Synchronization Description
+
+This sample code uses a somewhat novel data synchronization mechanism to share
```
xster, line 50: nit: "somewhat novel" sounds a bit tentative/risky. The mechanism itself isn't that special. It's just a bit particular to work with the communication constrains present currently.

xster:
In terms of the general directory structure, I don't have too strong of an opinion either way. I'd just like us to be consistent for all the samples. We can change the pattern but we have to make sure it works with some of the many:many groupings we have between modules and host apps like ios_fullscreen and ios_using_prebuilt_module both using flutter_module etc. 

Similarly for the readme, it could either all be in the root readme or it could all be in the subfolders. But I'd prefer not mixing patterns.

gaaclarke:
I think what is being proposed as consistent is a step backwards in terms of layout of the information and not actually helpful.  I don't think someone is going to go looking for `add_to_app/ios_multiple_flutters` and be confused when it isn't there.

People don't come to the samples and run through every single sample, they come looking for a sample that is apropos to what they want to do.  They will go to `add_to_app/README.md` and it will link them to `add_to_app/multiple_flutters` and they'll see all the information available about that sample right there.

Having [android, ios, module] all in the same directory makes their relationship more explicit, is easier to navigate and scales better if we add more platforms.  Also, having `add_to_app/README.md` act as a listing of samples instead of an aggregation of all the detailed information of all the samples is more useful.

If you both think I'm wrong here I can change it.

gaaclarke:
Alternatively, I can restructure the other samples.

xster:
Discussed offline. Aaron will make another PR to restructure the existing folders to be 2 tiered as he's proposing for this PR into theme/platform. Flutter module folders ambiguously shared between 2 themes will be duplicated.

gaaclarke:
This has now been rebased onto the restructuring commit so it matches everything now.  I've also updated the CI scripts to run the tests on the new directories.  This should be good to go now.

xster (approved):
There were a few leftover comments but otherwise LGTM!

gaaclarke:
@xster I think all comments were addressed, everything has been replied to at least.  Anything you had in mind?

gaaclarke:
@RedBrogdon I think your last concerns were the folder structure, the underlying structure was redone to make this, PTAL.

add_to_app/multiple_flutters/README.md
xster: I think just this

xster (approved):

add_to_app/multiple_flutters/README.md
gaaclarke: done

add_to_app/multiple_flutters/README.md
```diff
@@ -0,0 +1,80 @@
+# multiple_flutters
+
+This is a sample that shows how to use the Flutter Engine Groups API to embed
+multiple instances of Flutter into an existing Android or iOS project.
+
+## Getting Started
+
+For iOS instructions, see:
+[multiple_flutters_ios](../multiple_flutters_ios/README.md).
```
RedBrogdon, line 9: nit: Are the `../` paths still necessary with the new folder structure?

tool/ios_ci_script.sh
```diff
@@ -21,6 +21,11 @@ pushd add_to_app/fullscreen/flutter_module
 flutter packages get
 popd
 
+echo "Fetching dependencies and building 'multiple_flutters/multiple_flutters_module'."
+pushd add_to_app//multiple_flutters/multiple_flutters_module
```
RedBrogdon, line 25: nit: double slash.

RedBrogdon (approved):
Two small nits, but otherwise this looks great. Go ahead and land it whenever you're ready.

Thanks for taking the time to re-organize the projects in `add_to_app`, BTW. 👍

tool/ios_ci_script.sh
gaaclarke: done

add_to_app/multiple_flutters/README.md
gaaclarke: done

ducdung9295 (commented):
Trần văn Đức

Issue #674, opened 2021-01-25
Bug in Custom Tween animation sample 


priyanshu-01:
bug in samples/animations/lib/src/basics/06_custom_tween.dart

what's the bug?
A RenderFlex overflowed by 8.0 pixels on the bottom.

why it happened?
when the animation is completed, Whole Text with font size : 16.0 cannot be accommodated on the screen(mobile),

what's the apparent fix?
Wrapping the Column containing Text Widget with SingleChildScrollView so the column Size can be adjusted as per the amount of text to be accommodated

<img width="960" alt="bug" src="https://user-images.githubusercontent.com/56172661/105730986-ecd1dc00-5f54-11eb-8b95-b77b0f8533dc.png">

priyanshu-01:
I'll work on this

priyanshu-01:
please review the pull request #675

PR #675, opened 2021-01-25
bug fix in CutomTweenDemo

priyanshu-01:
fixes: #674

### DIFF
--- a/animations/lib/src/basics/06_custom_tween.dart
+++ b/animations/lib/src/basics/06_custom_tween.dart
@@ -71,27 +71,32 @@ class _CustomTweenDemoState extends State<CustomTweenDemo>
         ],
       ),
       body: SafeArea(
-        child: Column(
-          mainAxisAlignment: MainAxisAlignment.end,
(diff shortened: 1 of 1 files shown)


priyanshu-01:
@googlebot I signed it!

domesticmouse:
Hey @priyanshu-01 ,

Can you please add an explanation to this bug as to the issue and the proposed fix? Thanks!

brett

priyanshu-01:
Sure @domesticmouse , I hope this explanation will help!

The bug was in samples/animations/lib/src/basics/06_custom_tween.dart

The bug was:
A RenderFlex overflowed by 8.0 pixels on the bottom.

when it happened?
The bug happened when the 'write Essay' button was clicked.

why it happened?
when the animation is completed, Whole Text with font size : 16.0 cannot be accommodated on the screen(mobile), 

How I fixed this bug?
I Wrapped the Column containing Text Widget with SingleChildScrollView so the column Size can be adjusted as per the amount of text to be accommodated.

domesticmouse (approved):

domesticmouse:
Thank you!

PR #677, opened 2021-01-26
Multiple flutters sample android

gaaclarke:
issue: https://github.com/flutter/flutter/issues/74530


add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/App.kt
```diff
@@ -0,0 +1,13 @@
...
+import android.app.Application
+import io.flutter.embedding.engine.FlutterEngineGroup
+
+/**
+ * Application class for this app.
+ *
+ * This holds onto our engine group.
+ */
+class App : Application() {
+    val engines by lazy { FlutterEngineGroup(this) }
```
xster, line 12: don't know if we talked offline before or after this PR
 
```kotlin
lateinit val engines: FlutterEngineGroup

override fun onCreate() {
  super.onCreate()
  engines = ...
}
```

might solve the issue you had

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DataModel.kt
```diff
@@ -0,0 +1,41 @@
...
+ * A singleton/observable data model for the data shared between Flutter and the host platform.
+ *
+ * This is the definitive source of truth for all data.
+ */
+class DataModel {
+    companion object {
+        val instance = DataModel()
+    }
+
+    private val observers =  mutableListOf<WeakReference<DataModelObserver>>()
```
xster, line 22: Kotlin ftw!!

xster, line 22: Same comment as swift. Should this be a set? Since on Android it's not happening on an SDK class where the lifecycle is very clear, alternatively add some state assertions to the consuming class?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DataModel.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    public var counter = 0
+        set(value) {
```
xster, line 25: oh man, kotlin ftw!

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DataModel.kt
```diff
@@ -0,0 +1,41 @@
...
+            field = value
+            for (observer in observers) {
+                observer.get()?.onCountUpdate(value)
```
xster, line 28: you want to do the same thing as iOS here? (e.g. cull the dead references)?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+import androidx.fragment.app.FragmentManager
+import io.flutter.embedding.android.FlutterFragment
+import io.flutter.embedding.engine.FlutterEngineCache
+
+/**
+ * An activity that displays 2 FlutterFragments vertically.
+ */
+class DoubleFlutterActivity : FragmentActivity(), EngineBindingsDelegate {
+    private val topBindings : EngineBindings by lazy {
+        EngineBindings(this, this, "topMain")
```
xster, line 17: you can use named arguments to disambiguate

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+    private val bottomBindings : EngineBindings by lazy {
```
xster, line 19: I think the lateinit val pattern might be a bit clearer (since it's less ambiguous by what time they become not null). Lazy is better for things we might indeterministically (or never) access, which isn't the case here.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+            LinearLayout.LayoutParams.MATCH_PARENT
+        )
+        root.orientation = LinearLayout.VERTICAL
+        root.weightSum = numberOfFlutters.toFloat()
+
+        val fragmentManager: FragmentManager = supportFragmentManager
+
+        setContentView(root)
+
+        val app : App = applicationContext as App
```
xster, line 38: I don't think you need to specify the type twice

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DataModel.kt
xster: though for fwiw, if you use the engine cache, then things will never be gc'ed under the weak reference. I'd probably just stick to explicit destructions which make things more manageable in the JVM anyway.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+            fragmentManager
+                .beginTransaction()
+                .add(
+                    12345 + i,
+                    flutterFragment
+                )
+                .commit()
+        }
+
+        topBindings.attach()
```
xster, line 62: Could we do this for iOS too for consistency?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+        bottomBindings.attach()
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        topBindings.detach()
+        topBindings.detach()
```
xster, line 69: bottom

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
```
xster, line 70: maybe this is where we remove the engine cache?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
```diff
@@ -0,0 +1,82 @@
...
+ * This binds a FlutterEngine instance with the DataModel and a channel for communicating with that
+ * engine.
+ *
+ * Messages involving the DataModel are handled by the EngineBindings, other messages are forwarded
+ * to the EngineBindingsDelegate.
+ *
+ * @see main.dart for what messages are getting sent from Flutter.
+ */
+class EngineBindings(activity: Activity, delegate: EngineBindingsDelegate, entrypoint: String) : DataModelObserver {
+    var channel: MethodChannel
```
xster, line 33: these are all val right?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
```diff
@@ -0,0 +1,82 @@
...
+    var engine: FlutterEngine
+    val delegate: EngineBindingsDelegate
+
+    init {
+        val app: App = activity.applicationContext as App
```
xster, line 38: ditto, you may not need to type it twice

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
```diff
@@ -0,0 +1,82 @@
...
+        // This has to be lazy to avoid creation before the FlutterEngineGroup.
+        val dartEntrypoint by lazy {
```
xster, line 40: might not need this

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
```diff
@@ -0,0 +1,82 @@
...
+            DartExecutor.DartEntrypoint(FlutterInjector.instance().flutterLoader().findAppBundlePath(), entrypoint)
+        }
+        engine = app.engines.createAndRunEngine(activity, dartEntrypoint)
```
xster, line 43: ah that's where engines is lazy built first. I think this is an example of the init lifecycle not leaking outside the class and being more deterministic using lateinit vs lazy

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
```diff
@@ -0,0 +1,82 @@
...
+        this.delegate = delegate
+        channel = MethodChannel(engine.dartExecutor.binaryMessenger, "multiple-flutters")
+    }
+
+    /**
+     * This setups the messaging connections on the platform channel and the DataModel.
+     */
+    fun attach() {
+        DataModel.instance.addObserver(this)
+        channel.invokeMethod("setCount", DataModel.instance.counter)
```
xster, line 53: somewhat surprised this didn't ask you to null check it if it's a var

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * This tears down the messaging connections on the platform channel and the DataModel.
+     */
+    fun detach() {
+        DataModel.instance.removeObserver(this)
+        channel.setMethodCallHandler(null)
+    }
+
+    override fun onCountUpdate(newCount: Int) {
```
xster, line 79: doc

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.flutter.embedding.android.FlutterActivity
+
+/**
+ * The activity that uses standard platform UI APIs.
+ *
+ * This doesn't talk directly to or render Flutter content and represents a part of your app that
+ * would potentially already be written that wants to interface with Flutter.
+ */
+class MainActivity : AppCompatActivity(), DataModelObserver {
+    private var countView : TextView? = null
```
xster, line 19: private lateinit val

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+import android.widget.TextView
+import io.flutter.embedding.android.FlutterActivity
+
+/**
+ * The activity that uses standard platform UI APIs.
+ *
+ * This doesn't talk directly to or render Flutter content and represents a part of your app that
+ * would potentially already be written that wants to interface with Flutter.
+ */
+class MainActivity : AppCompatActivity(), DataModelObserver {
```
xster, line 18: I like the swift name better. It's not so much of a "main" activity since it'll cycle back over and over again.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+    private var countView : TextView? = null
+    private val mainActivityIdent : Int
```
xster, line 20: what's ident?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+    private companion object {
+        var mainActivityCount = 0
```
xster, line 22: add a comment?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+        countView?.text = DataModel.instance.counter.toString()
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        DataModel.instance.removeObserver(this)
+    }
+
+    fun onClickNext(view : View) {
+        val nextClass = if (mainActivityIdent % 2 == 0) SingleFlutterActivity::class.java else DoubleFlutterActivity::class.java
```
xster, line 44: comment

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+        countView = findViewById(R.id.count)
+        countView?.text = DataModel.instance.counter.toString()
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        DataModel.instance.removeObserver(this)
+    }
+
+    fun onClickNext(view : View) {
```
xster, line 43: add some docs to describe where this would visually come from

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/SingleFlutterActivity.kt
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        engineBindings.attach()
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+       engineBindings.detach()
```
xster, line 26: ultra-nit: indent

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/SingleFlutterActivity.kt
```diff
@@ -0,0 +1,37 @@
...
+import io.flutter.embedding.android.FlutterActivity
+import io.flutter.embedding.engine.FlutterEngine
+
+/**
+ * This is an Activity that displays one instance of Flutter.
+ *
+ * EngineBindings is used to bridge communication between the Flutter instance and the DataModel.
+ */
+class SingleFlutterActivity : FlutterActivity(), EngineBindingsDelegate {
+    private val engineBindings : EngineBindings by lazy {
```
xster, line 15: lateinit

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/res/layout/activity_main.xml
```diff
@@ -0,0 +1,52 @@
...
+    android:layout_height="match_parent"
+    tools:context=".MainActivity">
+
+    <TextView
+        android:id="@+id/count"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="1dp"
+        android:layout_marginRight="260dp"
+        android:text="Hello World!"
```
xster, line 15: does the ui say 

count:
hello world!

initially?

xster (commented):
nice! kotlin looks pretty clean!

add_to_app/multiple_flutters/multiple_flutters_android/app/build.gradle
```diff
@@ -0,0 +1,50 @@
+plugins {
+    id 'com.android.application'
+    id 'kotlin-android'
+}
+
+android {
+    signingConfigs {
+        self {
```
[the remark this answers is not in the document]
gaaclarke: No idea.  Android studio added it, maybe when I was trying to solve my signing issue for release?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/App.kt
gaaclarke: That would work, too.  I'm not sure if it is preferable to the lazy evaluation.  I googled it a bit and it isn't clear.  I can see some arguments either way.  I believe it's required just because you can't call abstract methods in the constructor, so literally any time after the constructor should be fine.  If that was not the case I think lateinit would be ideal since it would crash a bit earlier in a situation where we can't completely protect ourselves from crashing.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DataModel.kt
gaaclarke: Oh man, so much easier in this case.  Swift's protocols are weird.

gaaclarke: It isn't necessary for this example to function correctly.  I think a designer of a system for production code would have to make the semantics very clear about what happens when you double subscribe the same object.  Another option is just to throw an exception if someone double subscribes.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DataModel.kt
gaaclarke: I'm doing the same thing I'm doing on iOS.  Skip on notify, cull on removeObserver.

This should handle anybody that is being GC'd or manually deleted.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
gaaclarke: done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
gaaclarke: removed

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
gaaclarke: It already is happening, but in a slightly different way.  On iOS SingleFlutterViewController bundles the bindings with itself so it's not necessary.  I wouldn't want to add a concept that isn't necessary.  I don't think that would make the sample more clear.

A more similar approach would be to have Fragment that bundles the communication with a FlutterFragment, then have SigqleFlutterActivity and DoubleFlutterActivity embed that Fragment.  Unfortunately for me, converting an Activity to a Fragment isn't trivial.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
gaaclarke: There is no such thing as `latinit val`, there is only `lateinit var`, that's why I think lazy is better here.  Kotlin doesn't have the idea of `final` either.  From a user's perspective they should never need to think about the value ever being null.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
gaaclarke: done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
gaaclarke: done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
gaaclarke: yep, done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
gaaclarke: done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
gaaclarke: Hmm, me too.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
gaaclarke: done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
gaaclarke: done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
gaaclarke: Renamed, mainActivityIdentifier

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
gaaclarke: done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
gaaclarke: Ok, I switched engines to lateinit.  If this was code that I was writing for me I would prefer lazy evals, but having a lazy eval for a variable is a bit much for sample code.

I kept lazy for the engine bindings because they are actually used before onCreate in the case of SingleFlutterActivity, so the code is cleaner this way.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
gaaclarke: done

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/SingleFlutterActivity.kt
gaaclarke: done, reformatted all the code automatically.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/SingleFlutterActivity.kt
gaaclarke: mentioned above my these should still be lazy

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/res/layout/activity_main.xml
gaaclarke: Nope

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/App.kt
gaaclarke: Switch it, explained below.

add_to_app/multiple_flutters/multiple_flutters_android/app/build.gradle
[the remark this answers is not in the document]
xster: I think we can remove it

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DataModel.kt
xster: Either way is ok. I'd just add some comments or docs saying we're taking precautions with the engine bindings or this is a simple demo and you might run into trouble in these circumstances.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
xster: Ya, I just meant generally where the platform channel handlers sit between the 2 platforms if someone was looking for them in each of the platforms. If they need to be in different sounding files, maybe add a comment cross-referencing the difference?

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/EngineBindings.kt
xster: SG

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/MainActivity.kt
xster: There's still this bit (maybe you didn't want to do a rename in the same commit)

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/SingleFlutterActivity.kt
xster: ack

xster:
Didn't go through the complete code again yet. I think my only remaining comment was the MainActivity name. Otherwise I think this is ready to move out of draft after your folder move.

gaaclarke:
@xster Do you know how the CI works?  Looks like the version of Flutter that the tests are being run against don't have access to the new API.  It's choking on "FlutterEngineGroup".

xster:
Oh wait, right. @RedBrogdon I'm somewhat surprised the iOS PR passed the CI. 
The idea was that the samples must be buildable using the latest stable branch unless the samples are in the experimental folder. But I'm not sure why that test didn't fail with the iOS PR (which is using FlutterEngineGroup, not in 1.22). 

As a solution, I think it'd be ok if we waited until next month to click the merge button once the stable is pushed. @RedBrogdon I don't know if there's otherwise a cheaper solution than moving everything to experimental then moving back. i.e. some flag we can maybe set in the test file?

RedBrogdon:
> Looks like the intent is that things are in experimental if they use API's that haven't landed on stable.

That's correct. One of the invariants, so to speak, of the repo is that every root-level project should build/test with the stable channel, and everything that can't should be in `/experimental`.

> Oh wait, right. @RedBrogdon I'm somewhat surprised the iOS PR passed the CI.

Y'all just uncovered a bug in our CI setup. Looks like `xcpretty` is swallowing error codes returned by `xcodebuild`. We should be configuring it to return the error code instead:

https://www.rubydoc.info/gems/xcpretty/0.2.8#usage

I'll create a separate PR to fix it.

> As a solution, I think it'd be ok if we waited until next month to click the merge button once the stable is pushed. @RedBrogdon I don't know if there's otherwise a cheaper solution than moving everything to experimental then moving back. i.e. some flag we can maybe set in the test file?

You can certainly leave a PR open, since it's unlikely to run into merge conflicts. That said, the iOS version of the app has landed, and I can't fix the hole you just identified in our CI without dealing with that fact. 😄 

That leaves a few options:

* Move `multiple_flutters` to experimental and land the Android version there as well.
* Remove `multiple_flutters` entirely, add the iOS and Flutter versions to this PR (unwieldy, but doable), and leave this PR open for now.
* Leave `multiple_flutters` in place, remove it from the Flutter and iOS CI project lists, and leave this PR open for now.
* Remove `multiple_flutters`, add it to this PR, then land this PR in the `beta` branch of this repo, which is where we're getting samples for the next release ready.

WDYT? I hadn't realized y'all were targeting anything other than stable, or I would have mentioned this earlier.

gaaclarke:
I don't have a good idea about when the API is hitting stable, but a 5th option is that we pause the train until the API hits stable.  That means this doesn't land and your iOS change doesn't land.  That's the least amount of work for everyone but might not be ideal depending on the timeline for the next stable release.  Looks like the last one was 1/25/21 so it should't be long now I imagine.

xster:
Waiting to merge sounds ok to me. That would be the simplest thing to do.

domesticmouse:
I'd suggest pull from master to fix the Master Branch CI stable channel failures

RedBrogdon (approved):
Looks like Flutter tests are failing because they're pulling in older master code that's not touched by this PR. I'm going to call that a quirk of how GHA works and not a problem with this code, and then do the merge anyway.

saurabhchalke:
Is there a way to start with a Flutter screen at the beginning and bypassing any Kotlin screens at all?
I am trying to tweak this example with Flutter -> Native -> Flutter flow.

Also, the Kotlin to Flutter navigation seems to be slow on my Phone (Mi A2, Android 10), although Flutter to Kotlin seems to be quick.
Do you see any optimizations possible in here?

gaaclarke:
> I Is there a way to start with a Flutter screen at the beginning and bypassing any Kotlin screens at all?

Yep, I don't see any reason why it shouldn't work.

> Also, the Kotlin to Flutter navigation seems to be slow on my device (Mi A2, Android 10), although Flutter to Kotlin seems to be quick.
Do you see any optimizations possible in here?

It is slow on debug builds and on emulators, on release builds on real devices it is fast.  If you found a scenario where the transition isn't fast on release builds / real devices, please file an issue with the supporting evidence and reproduction steps.

PR #694, opened 2021-01-31
Adding new workflow for beta.

RedBrogdon:
This creates a second workflow that runs on the beta branch. There are a number of outstanding PRs that should be landed there, making this a needed first step.

FWIW, it's probably possibly to do some YAML-fu and get both branches into a single workflow, but I like the idea of having two completely separate ones.


RedBrogdon:
CC @johnpryan. I asked Brett for a review because he loves DevOps, but your feedback is welcome, too. 😄

RedBrogdon:
That's an interesting piece of trivia. If you force-push your branch to have no missing commits from the base, GitHub automatically closes your PR.

I'll open another.

PR #695, opened 2021-01-31
Adding new workflow for beta.

RedBrogdon:
This creates a second workflow that runs on the beta branch. There are a number of outstanding PRs that should be landed there, making this a needed first step.

FWIW, it's probably possibly to do some YAML-fu and get both branches into a single workflow, but I like the idea of having two completely separate ones.

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -2,9 +2,9 @@ name: Beta branch CI
 
 on:
   push:
-    branches: [ beta ]
+    branches: [beta]
   pull_request:
-    branches: [ beta ]
+    branches: [beta]
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * *' # Every day at midnight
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -1,4 +1,4 @@
-name: Flutter CI
+name: Master branch CI
```
domesticmouse, line 1: Probably worth changing this to "Flutter samples" as it is the string that shows up in the test status button in the README.md. Which also means the link needs to be changed, as the URL is going to change.

.github/workflows/beta.yml
```diff
@@ -0,0 +1,55 @@
+name: Beta branch CI
```
domesticmouse, line 1: What is the upside of building out a second workflow, as opposed to having one workflow with a matrix that includes stable, beta and dev?

domesticmouse:
My intent with how i designed [flutter_ci_script_stable.sh](https://github.com/flutter/samples/blob/master/tool/flutter_ci_script_stable.sh), [flutter_ci_script_beta.sh](https://github.com/flutter/samples/blob/master/tool/flutter_ci_script_beta.sh), and [flutter_ci_script_dev.sh](https://github.com/flutter/samples/blob/master/tool/flutter_ci_script_dev.sh) was that on the beta branch codelabs could be commented out in the stable CI script and added to the beta and dev CI scripts as demonstrated with how we are landing Matt Carrols codelab. 

This PR is in effect re-implementing that cut point, if i understand things correctly.

domesticmouse (approved):

RedBrogdon:
Discussed offline with @domesticmouse. This is intended to be a separate workflow that just runs for the beta branch, so that we can get new releases ready in advance. It's not intended to change the existing strategy for testing code in master, regardless of which channel that code targets.

Also, just so it gets said: our strategy for testing the beta branch will almost certainly evolve. This is just a first attempt. 😄

domesticmouse:
Sgtm!

PR #701, opened 2021-02-07
[beta branch] Migrate animation sample

RedBrogdon:
* Ups the SDK minimum for `/animations` to 2.12.0-0
* Migrates all the code
* Changes the animated list sample from using indexes to identify records to using IDs.
* Updates pubspec.(yaml|lock)

Comments especially welcome on the changes for migration.

### DIFF
--- a/animations/lib/src/basics/05_animated_builder.dart
+++ b/animations/lib/src/basics/05_animated_builder.dart
@@ -16,8 +16,8 @@ class _AnimatedBuilderDemoState extends State<AnimatedBuilderDemo>
   static const Color beginColor = Colors.deepPurple;
   static const Color endColor = Colors.deepOrange;
   Duration duration = Duration(milliseconds: 800);
-  AnimationController controller;
-  Animation<Color> animation;
+  late AnimationController controller;
+  late Animation<Color?> animation;
 
   @override
   void initState() {
--- a/animations/lib/src/basics/06_custom_tween.dart
+++ b/animations/lib/src/basics/06_custom_tween.dart
@@ -5,13 +5,13 @@
 import 'package:flutter/material.dart';
 
 class TypewriterTween extends Tween<String> {
-  TypewriterTween({String begin = '', String end})
+  TypewriterTween({String begin = '', String end = ''})
       : super(begin: begin, end: end);
 
   @override
   String lerp(double t) {
-    var cutoff = (end.length * t).round();
-    return end.substring(0, cutoff);
+    var cutoff = (end!.length * t).round();
+    return end!.substring(0, cutoff);
   }
 }
 
@@ -26,8 +26,8 @@ class _CustomTweenDemoState extends State<CustomTweenDemo>
     with SingleTickerProviderStateMixin {
   static const Duration _duration = Duration(seconds: 3);
   static const String message = loremIpsum;
-  AnimationController controller;
-  Animation<String> animation;
+  late AnimationController controller;
+  late Animation<String> animation;
 
   @override
   void initState() {
--- a/animations/lib/src/basics/06_custom_tween.dart
+++ b/animations/lib/src/basics/06_custom_tween.dart
@@ -26,8 +26,8 @@ class _CustomTweenDemoState extends State<CustomTweenDemo>
     with SingleTickerProviderStateMixin {
   static const Duration _duration = Duration(seconds: 3);
   static const String message = loremIpsum;
-  late AnimationController controller;
-  late Animation<String> animation;
+  late final AnimationController controller;
+  late final Animation<String> animation;
 
   @override
   void initState() {
--- a/animations/lib/src/basics/07_tween_sequence.dart
+++ b/animations/lib/src/basics/07_tween_sequence.dart
@@ -14,8 +14,8 @@ class TweenSequenceDemo extends StatefulWidget {
 class _TweenSequenceDemoState extends State<TweenSequenceDemo>
     with SingleTickerProviderStateMixin {
   static const Duration duration = Duration(seconds: 3);
-  AnimationController controller;
-  Animation<Color> animation;
+  late AnimationController controller;
+  late Animation<Color?> animation;
 
   static final colors = [
     Colors.red,
@@ -31,7 +31,7 @@ class _TweenSequenceDemoState extends State<TweenSequenceDemo>
   void initState() {
     super.initState();
 
-    final sequenceItems = <TweenSequenceItem<Color>>[];
+    final sequenceItems = <TweenSequenceItem<Color?>>[];
 
     for (var i = 0; i < colors.length; i++) {
       final beginColor = colors[i];
@@ -39,7 +39,7 @@ class _TweenSequenceDemoState extends State<TweenSequenceDemo>
       final weight = 1 / colors.length;
 
       sequenceItems.add(
-        TweenSequenceItem<Color>(
+        TweenSequenceItem<Color?>(
           tween: ColorTween(begin: beginColor, end: endColor),
           weight: weight,
         ),
@@ -47,7 +47,7 @@ class _TweenSequenceDemoState extends State<TweenSequenceDemo>
     }
 
     controller = AnimationController(duration: duration, vsync: this);
-    animation = TweenSequence<Color>(sequenceItems).animate(controller);
+    animation = TweenSequence<Color?>(sequenceItems).animate(controller);
   }
 
   @override
--- a/animations/lib/src/basics/07_tween_sequence.dart
+++ b/animations/lib/src/basics/07_tween_sequence.dart
@@ -14,8 +14,8 @@ class TweenSequenceDemo extends StatefulWidget {
 class _TweenSequenceDemoState extends State<TweenSequenceDemo>
     with SingleTickerProviderStateMixin {
   static const Duration duration = Duration(seconds: 3);
-  late AnimationController controller;
-  late Animation<Color?> animation;
+  late final AnimationController controller;
+  late final Animation<Color?> animation;
 
   static final colors = [
     Colors.red,
--- a/animations/lib/src/basics/08_fade_transition.dart
+++ b/animations/lib/src/basics/08_fade_transition.dart
@@ -15,9 +15,9 @@ class FadeTransitionDemo extends StatefulWidget {
 
 class _FadeTransitionDemoState extends State<FadeTransitionDemo>
     with SingleTickerProviderStateMixin {
-  AnimationController _controller;
-  Animation<double> _animation;
-  CurvedAnimation _curve;
+  late AnimationController _controller;
+  late Animation<double> _animation;
+  late CurvedAnimation _curve;
 
   @override
   void initState() {
--- a/animations/lib/src/basics/08_fade_transition.dart
+++ b/animations/lib/src/basics/08_fade_transition.dart
@@ -15,9 +15,9 @@ class FadeTransitionDemo extends StatefulWidget {
 
 class _FadeTransitionDemoState extends State<FadeTransitionDemo>
     with SingleTickerProviderStateMixin {
-  late AnimationController _controller;
-  late Animation<double> _animation;
-  late CurvedAnimation _curve;
+  late final AnimationController _controller;
+  late final Animation<double> _animation;
+  late final CurvedAnimation _curve;
 
   @override
   void initState() {
--- a/animations/lib/src/misc/animated_list.dart
+++ b/animations/lib/src/misc/animated_list.dart
@@ -20,17 +20,18 @@ class _AnimatedListDemoState extends State<AnimatedListDemo> {
     setState(() {
       var index = listData.length;
       listData.add(
-        UserModel(firstName: 'New', lastName: 'Person'),
+        UserModel(++_maxIdValue, 'New', 'Person'),
       );
       _listKey.currentState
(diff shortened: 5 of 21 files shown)


animations/lib/src/basics/05_animated_builder.dart
```diff
@@ -16,8 +16,8 @@ class _AnimatedBuilderDemoState extends State<AnimatedBuilderDemo>
   static const Color beginColor = Colors.deepPurple;
   static const Color endColor = Colors.deepOrange;
   Duration duration = Duration(milliseconds: 800);
-  AnimationController controller;
-  Animation<Color> animation;
+  late AnimationController controller;
+  late Animation<Color?> animation;
```
domesticmouse, line 20: I'm curious how nulls could be inserted this animation.

domesticmouse (approved):

animations/lib/src/basics/05_animated_builder.dart
RedBrogdon: While doing the conversion, this was a weird part for me as well. `ColorTween` is a glorified version of `Tween<Color?>`, which does allow for a null `Color` value:

https://github.com/flutter/flutter/blob/7c0300a3588d91542366d9efadda26fa2c9d9224/packages/flutter/lib/src/animation/tween.dart#L355

I thought about switching to `Tween<Color>`, but that would cause us to lose the `Color`-specific `lerp` function provided by `ColorTween`.

In the case of this example, I don't expect the tween to ever produce a null value, but it looks like `MaterialButton` is perfectly happy to accept one and just use the default color in response:

https://github.com/flutter/flutter/blob/88809aa24720705d1cd8de0bfc46faf932fe0b6e/packages/flutter/lib/src/material/material_button.dart#L180

animations/lib/src/basics/06_custom_tween.dart
```diff
@@ -26,8 +26,8 @@ class _CustomTweenDemoState extends State<CustomTweenDemo>
     with SingleTickerProviderStateMixin {
   static const Duration _duration = Duration(seconds: 3);
   static const String message = loremIpsum;
-  AnimationController controller;
-  Animation<String> animation;
+  late AnimationController controller;
+  late Animation<String> animation;
```
johnpryan, line 30: `late final`

animations/lib/src/basics/07_tween_sequence.dart
```diff
@@ -14,8 +14,8 @@ class TweenSequenceDemo extends StatefulWidget {
 class _TweenSequenceDemoState extends State<TweenSequenceDemo>
     with SingleTickerProviderStateMixin {
   static const Duration duration = Duration(seconds: 3);
-  AnimationController controller;
-  Animation<Color> animation;
+  late AnimationController controller;
+  late Animation<Color?> animation;
```
johnpryan, line 18: `late final`

animations/lib/src/basics/08_fade_transition.dart
```diff
@@ -15,9 +15,9 @@ class FadeTransitionDemo extends StatefulWidget {
 
 class _FadeTransitionDemoState extends State<FadeTransitionDemo>
     with SingleTickerProviderStateMixin {
-  AnimationController _controller;
-  Animation<double> _animation;
-  CurvedAnimation _curve;
+  late AnimationController _controller;
+  late Animation<double> _animation;
+  late CurvedAnimation _curve;
```
johnpryan, line 20: `late final`

animations/lib/src/misc/animated_list.dart
```diff
@@ -93,31 +94,23 @@ class _AnimatedListDemoState extends State<AnimatedListDemo> {
...
-    lastName: 'Cade',
-  ),
+  UserModel(0, 'Govind', 'Dixit'),
+  UserModel(1, 'Greta', 'Stoll'),
+  UserModel(2, 'Monty', 'Carlo'),
+  UserModel(3, 'Petey', 'Cruiser'),
+  UserModel(4, 'Barry', 'Cade'),
 ];
+
+int _maxIdValue = 4;
```
johnpryan, line 116: Could this be based off the length of the `listData` list instead?

animations/lib/src/misc/animated_list.dart
```diff
@@ -93,31 +94,23 @@ class _AnimatedListDemoState extends State<AnimatedListDemo> {
...
+    this.firstName,
+    this.lastName,
+  );
 
+  final int id;
   final String firstName;
   final String lastName;
 }
 
 List<UserModel> initialListData = [
```
johnpryan, line 108: nit: just inline this in the class, since it's only used once.

animations/lib/src/misc/animated_list.dart
```diff
@@ -20,17 +20,18 @@ class _AnimatedListDemoState extends State<AnimatedListDemo> {
     setState(() {
       var index = listData.length;
       listData.add(
-        UserModel(firstName: 'New', lastName: 'Person'),
+        UserModel(++_maxIdValue, 'New', 'Person'),
       );
       _listKey.currentState
-          .insertItem(index, duration: Duration(milliseconds: 300));
+          ?.insertItem(index, duration: Duration(milliseconds: 300));
```
johnpryan, line 26: I *think* it makes more sense to use `!`, since this would only be null if something is set up incorrectly:

```
  /// The current state is null if (1) there is no widget in the tree that
  /// matches this global key, (2) that widget is not a [StatefulWidget], or the
  /// associated [State] object is not a subtype of `T`.
```

animations/lib/src/misc/animated_list.dart
```diff
@@ -20,17 +20,18 @@ class _AnimatedListDemoState extends State<AnimatedListDemo> {
...
     });
   }
 
-  void deleteUser(int index) {
+  void deleteUser(int id) {
     setState(() {
+      final index = listData.indexWhere((u) => u.id == id);
       var user = listData.removeAt(index);
-      _listKey.currentState.removeItem(
+      _listKey.currentState?.removeItem(
```
johnpryan, line 34: Same here, I think this should be `!`, but I'm not 100% sure.

animations/lib/src/misc/card_swipe.dart
```diff
@@ -112,9 +112,9 @@ class SwipeableCard extends StatefulWidget {
 
 class _SwipeableCardState extends State<SwipeableCard>
     with SingleTickerProviderStateMixin {
-  AnimationController _controller;
-  Animation<Offset> _animation;
-  double _dragStartX;
+  late AnimationController _controller;
+  late Animation<Offset> _animation;
+  double? _dragStartX;
```
johnpryan, line 117: I think this can be `late double` since _dragStart is called before `_dragUpdate`

animations/lib/src/misc/carousel.dart
```diff
@@ -46,15 +46,15 @@ typedef OnCurrentItemChangedCallback = void Function(int currentItem);
...
+  const Carousel({Key? key, required this.itemBuilder}) : super(key: key);
 
   @override
   _CarouselState createState() => _CarouselState();
 }
 
 class _CarouselState extends State<Carousel> {
-  PageController _controller;
-  int _currentPage;
+  late PageController _controller;
```
johnpryan, line 56: `_controller` can be `late final`

animations/lib/src/misc/curved_animation.dart
```diff
@@ -15,14 +15,15 @@ class CurvedAnimationDemo extends StatefulWidget {
...
 }
 
 class _CurvedAnimationDemoState extends State<CurvedAnimationDemo>
     with SingleTickerProviderStateMixin {
-  AnimationController controller;
-  Animation<double> animationRotation;
-  Animation<Offset> animationTranslation;
+  late AnimationController controller;
+  late Animation<double> animationRotation;
+  late Animation<Offset> animationTranslation;
```
johnpryan, line 26: These can be `late final`

animations/lib/src/misc/curved_animation.dart
```diff
@@ -38,8 +39,8 @@ class _CurvedAnimationDemoState extends State<CurvedAnimationDemo>
     CurveChoice(curve: Curves.easeInCirc, name: 'Ease In Circle'),
     CurveChoice(curve: Curves.easeOutCirc, name: 'Ease Out Circle'),
   ];
-  CurveChoice selectedForwardCurve, selectedReverseCurve;
-  CurvedAnimation curvedAnimation;
+  late CurveChoice selectedForwardCurve, selectedReverseCurve;
+  late CurvedAnimation curvedAnimation;
```
johnpryan, line 43: `curvedAnimation` can be `late final`

animations/lib/src/misc/hero_animation.dart
```diff
@@ -18,7 +18,7 @@ class HeroAnimationDemo extends StatelessWidget {
           tag: 'hero-page-child',
           child: _createHeroContainer(
             size: 50.0,
-            color: Colors.grey[300],
+            color: Colors.grey[300]!,
```
johnpryan, line 21: Use `Colors.grey.shade300` instead

animations/lib/src/misc/physics_card_drag.dart
```diff
@@ -27,15 +27,15 @@ class PhysicsCardDragDemo extends StatelessWidget {
...
+  DraggableCard({required this.child});
 
   @override
   _DraggableCardState createState() => _DraggableCardState();
 }
 
 class _DraggableCardState extends State<DraggableCard>
     with SingleTickerProviderStateMixin {
-  AnimationController _controller;
+  late AnimationController _controller;
```
johnpryan, line 38: `late final`

animations/lib/src/misc/physics_card_drag.dart
```diff
@@ -44,7 +44,7 @@ class _DraggableCardState extends State<DraggableCard>
   /// this value is set to the value of the [_animation].
   var _dragAlignment = Alignment.center;
 
-  Animation<Alignment> _animation;
+  Animation<Alignment>? _animation;
```
johnpryan, line 47: could this be `late`?

animations/lib/src/misc/repeating_animation.dart
```diff
@@ -13,8 +13,8 @@ class RepeatingAnimationDemo extends StatefulWidget {
 
 class RepeatingAnimationDemoState extends State<RepeatingAnimationDemo>
     with SingleTickerProviderStateMixin {
-  AnimationController _controller;
-  Animation<BorderRadius> _borderRadius;
+  late AnimationController _controller;
+  late Animation<BorderRadius> _borderRadius;
```
johnpryan, line 17: these can both be `late final`

johnpryan (changes requested):

animations/lib/src/misc/animated_list.dart
RedBrogdon: I believe that would cause problems if items were repeatedly added and removed. For example:

1) Add a new UserModel, which would get an ID of 5 if the code's using the length of the list
2) Remove the first UserModel in the list.
3) Add another UserModel, which would also get an ID of 5.

animations/lib/src/misc/animated_list.dart
RedBrogdon: Done.

animations/lib/src/misc/card_swipe.dart
RedBrogdon: Nice catch. You're right, it's never assigned a nullable value.

animations/lib/src/misc/hero_animation.dart
RedBrogdon: TIL.

animations/lib/src/misc/physics_card_drag.dart
RedBrogdon: Now that I'm looking at this again, I think you're right.

animations/lib/src/basics/06_custom_tween.dart
```diff
@@ -5,13 +5,13 @@
 import 'package:flutter/material.dart';
 
 class TypewriterTween extends Tween<String> {
-  TypewriterTween({String begin = '', String end})
+  TypewriterTween({String begin = '', String end = ''})
```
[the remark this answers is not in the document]
RedBrogdon: Yeah, this threw me a bit, as you probably saw. Tween<T> takes a single type param, and then defines `begin` and `end` with the nullable version of that type parameter.

RedBrogdon:
@johnpryan Thanks, and PTAL! I believe I've addressed all the actionable comments.

animations/lib/src/basics/06_custom_tween.dart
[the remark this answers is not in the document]
johnpryan: Yeah in theory T? is a subtype of T so it's not clear to me why this isn't working

animations/lib/src/misc/animated_list.dart
johnpryan: ah, right.

johnpryan (approved):

animations/lib/src/basics/06_custom_tween.dart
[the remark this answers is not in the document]
RedBrogdon: I thought the subtype relationship was the reverse (`String` is a subtype of `String?`). I'm going to go read the [null safety guide](https://dart.dev/null-safety/understanding-null-safety#using-nullable-types) again. 😄

PR #702, opened 2021-02-07
[flutter_maps_firestore] Update sample

AyushBherwani1998:
## Description 
- Update samples to use new versions.


AyushBherwani1998:
@domesticmouse @RedBrogdon This can land on the master because, I have removed the generated_plugin_registrant and added it to git ignore. [Here's the related PR](https://github.com/flutter/flutter/pull/39781), the sample was created before that, hence it doesn't have it in the git ignore

flutter_maps_firestore/ios/Podfile
```diff
@@ -1,5 +1,5 @@
 # Uncomment this line to define a global platform for your project
-# platform :ios, '9.0'
+platform :ios, '10.0'
```
AyushBherwani1998, line 2: After cloud_firestore 0.15.0 the min support has been 10 or greater

flutter_maps_firestore/lib/main.dart
```diff
@@ -16,7 +17,11 @@ const _pinkHue = 350.0;
 // Places API client used for Place Photos
 final _placesApiClient = GoogleMapsPlaces(apiKey: googleMapsApiKey);
 
-void main() => runApp(App());
+Future<void> main() async {
+  WidgetsFlutterBinding.ensureInitialized();
+  await Firebase.initializeApp();
+  runApp(App());
+}
```
AyushBherwani1998, line 24: If this is not following the coding standards, then I can also use a `FutureBuilder` to initialize. Let me know your thoughts.

RedBrogdon: If you're referring to the `ensureInitialized` call, it's fine to do that in a sample. If this were a production app and we were focused on keeping launch time to the bare minimum it might be an issue, but code-readability is more important here.

RedBrogdon (approved):
LGTM

flutter_maps_firestore/lib/main.dart
domesticmouse: I've struggled with this one myself. I tend to put firebase initialisation closer to where it is used, eg in a provider state object or similar.

No hard guidelines

PR #707, opened 2021-02-08
Added Focusnodes in form sample

Alabhya268:
Added FocusNodes to improve user experience in form_app.

### DIFF
--- a/form_app/lib/src/autofill.dart
+++ b/form_app/lib/src/autofill.dart
@@ -14,12 +14,13 @@ class AutofillDemo extends StatefulWidget {
 class _AutofillDemoState extends State<AutofillDemo> {
   final _formKey = GlobalKey<FormState>();
 
-  FocusNode _lastname, _email, _telephone, _streetAddresss, _postalCode, _country, _countrycode;
+  FocusNode _firstName, _lastName, _email, _telephone, _streetAddresss, _postalCode, _country, _countrycode;
 
   @override
   void initState() {
     super.initState();
-    _lastname = FocusNode();
+    _firstName =FocusNode();
+    _lastName = FocusNode();
     _email = FocusNode();
     _telephone = FocusNode();
     _streetAddresss = FocusNode();
@@ -30,13 +31,14 @@ class _AutofillDemoState extends State<AutofillDemo> {
 
   @override
   void dispose() {
-    _lastname = FocusNode();
-    _email = FocusNode();
-    _telephone = FocusNode();
-    _streetAddresss = FocusNode();
-    _postalCode = FocusNode();
-    _country = FocusNode();
-    _countrycode = FocusNode();
+    _firstName.dispose();
+    _lastName.dispose();
+    _email.dispose();
+    _telephone.dispose();
+    _streetAddresss.dispose();
+    _postalCode.dispose();
+    _country.dispose();
+    _countrycode.dispose();
     super.dispose();
   }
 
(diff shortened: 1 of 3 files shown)


Alabhya268:
@RedBrogdon  @johnpryan , Please review this PR , Let me if anything needs to be changed.

form_app/lib/src/autofill.dart
```diff
@@ -14,6 +14,32 @@ class AutofillDemo extends StatefulWidget {
 class _AutofillDemoState extends State<AutofillDemo> {
   final _formKey = GlobalKey<FormState>();
 
+  FocusNode _lastname, _email, _telephone, _streetAddresss, _postalCode, _country, _countrycode;
```
RedBrogdon, line 17: nit: We generally use a max line length of 80 in this repo, and typically break fields out onto their own lines, even if the types are the same:

```dart
FocusNode _lastName;
FocusNode _mail;
... etc. ...
```

form_app/lib/src/autofill.dart
```diff
@@ -31,13 +57,21 @@ class _AutofillDemoState extends State<AutofillDemo> {
                   ...[
                     Text('This sample demonstrates autofill. '),
                     TextFormField(
+                      onFieldSubmitted: (vaue) {
+                        _lastname.requestFocus();
```
RedBrogdon, line 61: If we're adding focus nodes to the other fields, it's probably a good idea to add one to this field as well, in order to complete the set. I wouldn't be surprised if it mattered for tab-order when building for the web, for example.

form_app/lib/src/sign_in_http.dart
```diff
@@ -54,16 +68,19 @@ class _SignInHttpDemoState extends State<SignInHttpDemo> {
               children: [
                 ...[
                   TextFormField(
+                    textInputAction: TextInputAction.next,
```
RedBrogdon, line 71: Similar to the above, I'd recommend adding focus nodes to all the fields in this and the other forms.

RedBrogdon (changes requested):
I added a few nits, then realized @johnpryan is probably better suited to review this. 😄

Alabhya268:
@RedBrogdon , I made the necessary changes  : )

RedBrogdon (approved):
LGTM

form_app/lib/src/autofill.dart
```diff
@@ -31,13 +66,23 @@ class _AutofillDemoState extends State<AutofillDemo> {
...
                         hintText: 'Jane',
                         labelText: 'First Name',
                       ),
                       autofillHints: [AutofillHints.givenName],
                     ),
                     TextFormField(
+                      focusNode: _lastName,
+                      textInputAction: TextInputAction.next,
+                      onFieldSubmitted: (value) {
+                        _email.requestFocus();
```
johnpryan, line 84: Manually focusing the next field is not required when using `TextInputAction.next`, so the FocusNodes can be removed.

form_app/lib/src/autofill.dart
```diff
@@ -31,13 +66,23 @@ class _AutofillDemoState extends State<AutofillDemo> {
                   ...[
                     Text('This sample demonstrates autofill. '),
                     TextFormField(
+                      autofocus: true,
+                      focusNode: _firstName,
```
johnpryan, line 70: Add `textInputAction: TextInputAction.next` to this field too.

johnpryan (changes requested):

form_app/lib/src/autofill.dart
Alabhya268: I didn't knew it, I've now removed unnecessary focusnodes

Alabhya268:
@johnpryan ,Please review this PR again.

johnpryan (approved):

PR #710, opened 2021-02-09
Unmasking CI error codes from iOS

RedBrogdon:
The `xcpretty` gem used to prettyprint the output of `xcodebuild` has been masking error codes and causing tests to pass even when builds fail. This PR sets `pipefail` in the iOS CI script, fixing the issue.


RedBrogdon:
Q.V. https://github.com/flutter/samples/pull/677#issuecomment-775608148

xster (approved):

PR #712, opened 2021-02-09
Rename Navigator 2.0 to Router

johnpryan:
This aligns better with the documentation for this new API, and doesn't imply that apps need to be upgraded.

cc: @chunhtai


navigation_and_routing/README.md
```diff
@@ -1,38 +1,38 @@
 # Navigation and Routing sample code
 
 Sample code for [Learning Flutter's New Navigation and Routing System][article],
-an article explaining Navigator 2.0.
+an article explaining the Router widget and it's associated classes.
 
 ## Samples
 
-**Navigator 1.0 samples**
+**Introductory samples**
```
chunhtai, line 8: Introductory samples sounds a bit weird, Maybe just `Navigator samples`?

chunhtai (approved):
I left a comment on the wording. the rest LGTM

navigation_and_routing/README.md
johnpryan: I think that makes sense, I was concerned that it might be misleading to call these Navigator samples, since `Router` works alongside `Navigator` too

chunhtai: I have no strong opinion on this maybe you can add something like` existing navigator sample` or `imperative navigator example` but then you will need to explain what imperative is....

johnpryan: Me neither, I'll change this to `navigator`

RedBrogdon (approved):

Issue #729, opened 2021-02-22
[provider_shopper] Migrate to null safety


filiph:
The sample at https://github.com/flutter/samples/tree/master/provider_shopper needs to be migrated to sound null safety. I'm going to do it by Wednesday EOD unless someone else wants to take a stab at it soon. If so, please send a PR against the `beta` branch ([example](https://github.com/flutter/samples/pull/716)).

filiph:
(Possible to combine with #728.)

arteevraina:
@filiph I will send a patch for this by today.

arteevraina:
Hey @filiph , I will migrate this one by today.

arteevraina:
Since `provider_shopper` has been already migrated to null safety. I think this can be closed now. WDYT @filiph ?

PR #736, opened 2021-02-23
[provider_shopper] migrate to null safe

arteevraina:
fixes: #729 

* updated dependencies.
* migrated using "dart migrate".

Have a look @filiph @RedBrogdon @domesticmouse

### DIFF
--- a/provider_shopper/lib/main.dart
+++ b/provider_shopper/lib/main.dart
@@ -30,7 +30,7 @@ class MyApp extends StatelessWidget {
         ChangeNotifierProxyProvider<CatalogModel, CartModel>(
           create: (context) => CartModel(),
           update: (context, catalog, cart) {
-            cart.catalog = catalog;
+            cart!.catalog = catalog;
             return cart;
           },
         ),
--- a/provider_shopper/lib/main.dart
+++ b/provider_shopper/lib/main.dart
@@ -30,6 +30,7 @@ class MyApp extends StatelessWidget {
         ChangeNotifierProxyProvider<CatalogModel, CartModel>(
           create: (context) => CartModel(),
           update: (context, catalog, cart) {
+            if (cart == null) throw ArgumentError.notNull('cart');
             cart.catalog = catalog;
             return cart;
           },
--- a/provider_shopper/lib/main.dart
+++ b/provider_shopper/lib/main.dart
@@ -30,7 +30,7 @@ class MyApp extends StatelessWidget {
         ChangeNotifierProxyProvider<CatalogModel, CartModel>(
           create: (context) => CartModel(),
           update: (context, catalog, cart) {
-            cart!.catalog = catalog;
+            cart.catalog = catalog;
             return cart;
           },
         ),
--- a/provider_shopper/lib/models/cart.dart
+++ b/provider_shopper/lib/models/cart.dart
@@ -7,7 +7,7 @@ import 'package:provider_shopper/models/catalog.dart';
 
 class CartModel extends ChangeNotifier {
   /// The private field backing [catalog].
-  CatalogModel _catalog;
+  late CatalogModel _catalog;
 
   /// Internal, private state of the cart. Stores the ids of each item.
   final List<int> _itemIds = [];
@@ -16,9 +16,6 @@ class CartModel extends ChangeNotifier {
   CatalogModel get catalog => _catalog;
 
   set catalog(CatalogModel newCatalog) {
-    assert(newCatalog != null);
-    assert(_itemIds.every((id) => newCatalog.getById(id) != null),
-        'The catalog $newCatalog does not have one of $_itemIds in it.');
     _catalog = newCatalog;
     // Notify listeners, in case the new catalog provides information
     // different from the previous one. For example, availability of an item
--- a/provider_shopper/lib/models/cart.dart
+++ b/provider_shopper/lib/models/cart.dart
@@ -7,7 +7,7 @@ import 'package:provider_shopper/models/catalog.dart';
 
 class CartModel extends ChangeNotifier {
   /// The private field backing [catalog].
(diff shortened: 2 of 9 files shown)


provider_shopper/lib/main.dart
```diff
@@ -30,7 +30,7 @@ class MyApp extends StatelessWidget {
         ChangeNotifierProxyProvider<CatalogModel, CartModel>(
           create: (context) => CartModel(),
           update: (context, catalog, cart) {
-            cart.catalog = catalog;
+            cart!.catalog = catalog;
```
filiph, line 33: This seems unsafe. How do we _know_ we can assert non-null here when the `Provider` itself can't guarantee it? See `update`'s signature [here](https://pub.dev/documentation/provider/5.0.0-nullsafety.5/provider/ProxyProviderBuilder.html).

provider_shopper/lib/models/cart.dart
```diff
@@ -7,7 +7,7 @@ import 'package:provider_shopper/models/catalog.dart';
 
 class CartModel extends ChangeNotifier {
   /// The private field backing [catalog].
-  CatalogModel _catalog;
+  late CatalogModel _catalog;
```
filiph, line 10: Consider using `CatalogModel?` instead of `late CatalogModel` here. It makes the code more verbose but we can be more sure we're not unwittingly letting some null error slip through.

If we were _sure_ that `_catalog` will be set before it's used (e.g. it's assigned in the Constructor), then `late` is a great option.

provider_shopper/lib/models/cart.dart
```diff
@@ -16,9 +16,9 @@ class CartModel extends ChangeNotifier {
   CatalogModel get catalog => _catalog;
 
   set catalog(CatalogModel newCatalog) {
-    assert(newCatalog != null);
-    assert(_itemIds.every((id) => newCatalog.getById(id) != null),
-        'The catalog $newCatalog does not have one of $_itemIds in it.');
+    // // assert(newCatalog != null);
+    // assert(_itemIds.every((id) => newCatalog.getById(id) != null),
+    // 'The catalog $newCatalog does not have one of $_itemIds in it.');
```
filiph, line 21: Either delete or update. Don't leave commented code in the repo.

provider_shopper/lib/screens/cart.dart
```diff
@@ -65,7 +65,7 @@ class _CartTotal extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     var hugeStyle =
-        Theme.of(context).textTheme.headline1.copyWith(fontSize: 48);
+        Theme.of(context).textTheme.headline1!.copyWith(fontSize: 48);
```
filiph, line 68: I think this is unsafe, although probably fine for the purposes of this particular sample. We don't want to spend too much time on minutiae when the star of the show should be state management.

filiph (approved):
LGTM with suggestions

provider_shopper/lib/screens/cart.dart
filiph: Let's keep this as is.

arteevraina: Sure thing!

provider_shopper/lib/main.dart
arteevraina: Exactly, I think instead of `!` operator `?.` can be used ?

provider_shopper/lib/models/cart.dart
arteevraina: Yes, making it `CatelogModel?` makes complete sense. But, since I am changing this, I think I also need to change its getters and setters. For example :
`CatalogModel? get catalog => _catalog;`

provider_shopper/lib/models/cart.dart
arteevraina: Yes, I will keep in mind this in my upcoming patches. Probably, for this case these lines of code will be required again as we are making `CatalogModel` as `nullable`.

filiph: Ah, ok, I didn't realize this is still a draft PR. You can mark it as such on GitHub.

provider_shopper/lib/models/cart.dart
filiph: Yes, that change will reverberate through the code.

provider_shopper/lib/main.dart
filiph: You'll need to try. Maybe it'll make more sense to have a guard clause, such as `if (cart == null) return;`.

filiph:
@googlebot I consent.

filiph:
Hi @arteevraina, we need to land this today, so I'm taking over the PR. Hopefully, I'll be able to land this in the context of this PR, so you'll still get the credit.

provider_shopper/lib/main.dart
```diff
@@ -30,6 +30,7 @@ class MyApp extends StatelessWidget {
         ChangeNotifierProxyProvider<CatalogModel, CartModel>(
           create: (context) => CartModel(),
           update: (context, catalog, cart) {
+            if (cart == null) throw ArgumentError.notNull('cart');
```
filiph, line 33: I know this could have been done by doing `cart!.catalog` below, but this way it's more explicit. This really shouldn't ever happen in practice: the `update` method is called after the `create` method.

afitz0 (approved):
LGTM as well. Thanks for submitting this update to the sample!!!

arteevraina:
Thanks a bunch, @filiph ! :smile:  I am really sorry as I was slow to respond to this PR as this was my first encounter with null safety. I hope to make more impactful contributions in the future :confetti_ball:

filiph:
You did great, thanks!

Issue #737, opened 2021-02-23
NDK missing error in flutter/samples CI


gaaclarke:
An unrelated PR started showing an error in the samples CI:  https://github.com/flutter/samples/pull/677/checks?check_run_id=1923183048

`No version of NDK matched the requested version 21.3.6528147.`

It sounds related to https://github.com/flutter/flutter/issues/53657.

jmagman:
PR to upgrade that version to the one available on CI: https://github.com/flutter/samples/pull/738.  I'm not sure what a more permanent solution would be since this seems to be a moving target.

PR #738, opened 2021-02-23
Remove ndkVersion and buildToolsVersion version from add_to_app/android_view

jmagman:
> No version of NDK matched the requested version 21.3.6528147. Versions available locally: 21.4.7075529, 21.4.7075529, 21.4.7075529

https://github.com/flutter/samples/pull/677/checks?check_run_id=1923183048#step:5:430

Looks like the migration to GitHub actions made hardcoding these unnecessary in CI.

Fixes https://github.com/flutter/samples/issues/737

### DIFF
--- a/add_to_app/android_view/app/build.gradle
+++ b/add_to_app/android_view/app/build.gradle
@@ -4,7 +4,7 @@ plugins {
 }
 
 android {
-    ndkVersion "21.3.6528147"
+    ndkVersion "21.4.7075529"
     compileSdkVersion 30
     buildToolsVersion "30.0.3"
 
--- a/add_to_app/android_view/app/build.gradle
+++ b/add_to_app/android_view/app/build.gradle
@@ -4,9 +4,7 @@ plugins {
 }
 
 android {
-    ndkVersion "21.4.7075529"
     compileSdkVersion 30
-    buildToolsVersion "30.0.3"
 
     defaultConfig {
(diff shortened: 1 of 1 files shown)


RedBrogdon:
Is it possible to just yank that version number entirely? That's what we did to solve this problem for another sample in #717.

jmagman:
> Is it possible to just yank that version number entirely? That's what we did to solve this problem for another sample in #717.

That seems like the right thing to do, but it looks like @xster left it off https://github.com/flutter/samples/pull/632, then had to re-add it to make CI happy
https://github.com/flutter/samples/pull/632/commits/d3aaf72018f24a69001a5ad5b6bc9e50f3f48b44
https://github.com/flutter/samples/pull/632/commits/319f40d718eb7479a130b8533cda382d57bd1613

The failure was https://github.com/flutter/samples/pull/632/checks?check_run_id=1764739542#step:5:1343

RedBrogdon:
> That seems like the right thing to do, but it looks like @xster left it off #632, then had to re-add it to make CI happy

We've moved to GitHub actions in between then and now, and it's possible the hardcoded version was only required while we were using Travis. If you haven't already tried, it might be worth removing it just to see for certain if GHA needs it or not.

jmagman:
Removing it worked, thanks @RedBrogdon

gaaclarke (approved):

xster:
Thanks for re-slapping the tv

PR #740, opened 2021-02-24
Re-orged android_view to match the rest of the projects.

gaaclarke:


add_to_app/android_view/android_view/build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:4.2.0-beta02"
+        classpath 'com.android.tools.build:gradle:4.0.0'
```
gaaclarke, line 9: This was requiring people to use a beta version of android studio to build the sample.

tool/flutter_ci_script_stable.sh
```diff
@@ -3,6 +3,7 @@
 set -e
 
 declare -ar PROJECT_NAMES=(
+    "add_to_app/android_view/flutter_module_using_plugin" \
```
gaaclarke, line 6: Looks like the android part was never submitted to CI.  We should do that in another PR in case it's hairy and not conflate issues.

RedBrogdon (commented):
I imagine there's going to be an iOS version of this sample at some point as well? If that's the case, how about calling the root folder `partial_view` rather than `android_view`, so you could eventually end up with `/add_to_app/partial_view/android_view` and `add_to_app/partial_view/ios_view`?

gaaclarke:
> I imagine there's going to be an iOS version of this sample at some point as well? If that's the case, how about calling the root folder `partial_view` rather than `android_view`, so you could eventually end up with `/add_to_app/partial_view/android_view` and `add_to_app/partial_view/ios_view`?

Nope, there won't be an iOS variant.

RedBrogdon:
> Nope, there won't be an iOS variant.

SGTM!

RedBrogdon (approved):

PR #742, opened 2021-02-26
Restores `add_to_app/multiple_flutters/multiple_flutters_ios` to the iOS CI list

RedBrogdon:
This should not be merged until the next stable release.

### DIFF
--- a/tool/ios_ci_script.sh
+++ b/tool/ios_ci_script.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+set -o pipefail # Necessary so xcpretty won't mask xcodebuild failures later.
 
 echo "Pre-caching ios artifacts, such as the Flutter.framework"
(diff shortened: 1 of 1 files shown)


gaaclarke (approved):
LGTM, land after FlutterEngineGroup hits stable.

gaaclarke:
I rebased this for you to kickoff CI now that 2.0 is on stable.

RedBrogdon:
I'm not sure why these tests aren't passing. It looks like the GHA bots are pulling in an old version of the master branch code.

If the iOS CI script works, I'm inclined to call it a flake and land anyway. Thoughts?

gaaclarke:
@RedBrogdon It looks like there are new deprecations in 2.0 that are causing problems with some of the tests, unrelated to this change.  I can look into addressing them.  I'll land this in the meantime to make things easier.

gaaclarke:
Oh, looks like you already addressed the issues this CI was choking on, weird, OK.

PR #748, opened 2021-03-03
Upgrade null safety example

mit-mit:
* Move null safety example out of experimental
* Roll SDK constraints for null safety example
* Fix a deprecation warning in the isolate example


mit-mit:
CI is all green now. @RedBrogdon can you take a look?

RedBrogdon (approved):
LGTM. We talked at one point about finding another format for this content (codelab, guide, that sort of thing), since in comparison with the other samples it's a bit of an outlier in terms of structure. Is that still a conversation we can have?

mit-mit:
Sure thing!

PR #756, opened 2021-03-05
[testing_app] Fix failing integration tests

abd99:
After the release of Flutter 2, there were some integration tests that were failing in the sample.
This PR fixes those issues.

Please note that this sample does not use null-safety yet.

/cc @johnpryan

### DIFF
--- a/testing_app/integration_test/app_test.dart
+++ b/testing_app/integration_test/app_test.dart
@@ -80,7 +80,7 @@ void main() {
 
       // Tap on the remove icon.
       await tester.tap(find.byKey(ValueKey('remove_icon_5')));
-      await tester.pumpAndSettle();
+      await tester.pumpAndSettle(Duration(seconds: 1));
 
       // Verify if it disappears.
       expect(find.text('Item 5'), findsNothing);
--- a/testing_app/integration_test/driver.dart
+++ b/testing_app/integration_test/driver.dart
@@ -4,20 +4,4 @@
 
 import 'package:integration_test/integration_test_driver.dart';
 
-Future<void> main() {
-  return integrationDriver(
-    responseDataCallback: (data) async {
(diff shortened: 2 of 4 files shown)


testing_app/integration_test/driver.dart
```diff
@@ -4,20 +4,4 @@
...
-          await writeResponseData(
-            entry.value as Map<String, dynamic>,
-            testOutputFilename: entry.key,
-          );
-        }
-      }
-    },
-  );
-}
+Future<void> main() => integrationDriver();
```
abd99, line 7: Since the use of `traceAction` is removed, there are no complete timelines generated, just their summaries. Hence, the custom implementation of `responseDataCallback` isn't required.

testing_app/integration_test/app_test.dart
```diff
@@ -80,7 +80,7 @@ void main() {
 
       // Tap on the remove icon.
       await tester.tap(find.byKey(ValueKey('remove_icon_5')));
-      await tester.pumpAndSettle();
+      await tester.pumpAndSettle(Duration(seconds: 1));
```
abd99, line 83: Adding `Duration` to `pumpAndSettle` after performing every action was required because ScaffoldMessenger keeps Snackbars from previous screens.

abd99:
@johnpryan, I'm sharing my thoughts on the issue with Chrome's tracing here.

As per my understanding, chrome can only load timelines, it cannot load summaries. So, in a JSON file, when there is a combination of timelines and summaries, chrome can't load the data from it. Maybe that's why the docs on Flutter Driver tests always wrote timelines and summaries in different files.

I'm can't say anything about multiple timelines in a single file since I don't have a lot of experience with chrome's tracing tools.

Again, I might be wrong here, but I thought that I should share what I feel about this topic.

domesticmouse (approved):

Issue #762, opened 2021-03-09
Android multiple flutters sample can crash when navigating back through stack


xster:
Navigate forward until the single or double activity loops through twice, then start popping activities. When it gets to the first occurrence of the single/double activity, 

```
Caused by: java.lang.IllegalStateException: The requested cached FlutterEngine did not exist in the FlutterEngineCache: '0'
        at io.flutter.embedding.android.FlutterActivityAndFragmentDelegate.setupFlutterEngine(FlutterActivityAndFragmentDelegate.java:218)
        at io.flutter.embedding.android.FlutterActivityAndFragmentDelegate.onAttach(FlutterActivityAndFragmentDelegate.java:155)
        at io.flutter.embedding.android.FlutterFragment.onAttach(FlutterFragment.java:608)
        at androidx.fragment.app.Fragment.performAttach(Fragment.java:2574)
        at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:828)
        at androidx.fragment.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManagerImpl.java:1238)
        at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:1303)
        at androidx.fragment.app.FragmentManagerImpl.dispatchStateChange(FragmentManagerImpl.java:2659)
        at androidx.fragment.app.FragmentManagerImpl.dispatchCreate(FragmentManagerImpl.java:2607)
        at androidx.fragment.app.FragmentController.dispatchCreate(FragmentController.java:235)
        at androidx.fragment.app.FragmentActivity.onCreate(FragmentActivity.java:316)
        at dev.flutter.multipleflutters.DoubleFlutterActivity.onCreate(DoubleFlutterActivity.kt:25)
        at android.app.Activity.performCreate(Activity.java:8000)
        at android.app.Activity.performCreate(Activity.java:7984)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1309)
```

and the counter number resets.

gaaclarke:
I usually don't get that far because the emulator is so slow, haha.

PR #763, opened 2021-03-09
fixed crash if you back out of 2 double engine activities

gaaclarke:
fixes https://github.com/flutter/samples/issues/762

### DIFF
--- a/add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
+++ b/add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
@@ -20,9 +20,14 @@ class DoubleFlutterActivity : FragmentActivity(), EngineBindingsDelegate {
         EngineBindings(activity = this, delegate = this, entrypoint = "bottomMain")
     }
     private val numberOfFlutters = 2
-    private var engineCountStart = 0
+    private val engineCountStart : Int
     private companion object {
-        var engineCount = 0
+        var engineCounter = 0
+    }
+
+    init {
+        engineCountStart = engineCounter
+        engineCounter += numberOfFlutters
     }
 
     override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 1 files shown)


add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
```diff
@@ -20,6 +20,10 @@ class DoubleFlutterActivity : FragmentActivity(), EngineBindingsDelegate {
         EngineBindings(activity = this, delegate = this, entrypoint = "bottomMain")
     }
     private val numberOfFlutters = 2
+    private var engineCountStart = 0
+    private companion object {
+        var engineCount = 0
```
xster, line 25: `totalEngineCount` for readability? Since you have a lot of `engine*` names in this file.

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
```diff
@@ -20,6 +20,10 @@ class DoubleFlutterActivity : FragmentActivity(), EngineBindingsDelegate {
         EngineBindings(activity = this, delegate = this, entrypoint = "bottomMain")
     }
     private val numberOfFlutters = 2
+    private var engineCountStart = 0
```
xster, line 23: Could this be `private val engineCountStart = engineCount`? We need to make sure the value never changes through future refactors so the onDestroy start number is right and not leak.

xster (approved):

add_to_app/multiple_flutters/multiple_flutters_android/app/src/main/java/dev/flutter/multipleflutters/DoubleFlutterActivity.kt
gaaclarke: IMO that is a bit misleading because that would suggest it was the number of engines alive, not the number of engines ever made.

xster: you're right. Let's leave it as is.

PR #787, opened 2021-04-30
Begins readying new beta branch

RedBrogdon:
* Updates pubspec.locks with beta channel SDK.
* Reworks the dependencies for Veggieseasons to update them to more modern versions. Previously this was causing an odd compile error in the Win32 version of path_provider.
* Tweaks test scripts:
  - Removes `animations` and `isolate_example` from dev channel testing
    * There's a null safety-related API change to one of the animation classes that borks them. We can fix once we're past the current beta.
  - Sorts the project lists in the CI scripts and adds `multiple_flutters_module` back to the list.
  - Corrects typo in the shared CI script.


tool/flutter_ci_script_dev.sh
```diff
@@ -6,30 +6,30 @@ DIR="${BASH_SOURCE%/*}"
...
+    "add_to_app/prebuilt_module/flutter_module"
+    # "animations"
+    "experimental/desktop_photo_search"
+    "experimental/federated_plugin/federated_plugin"
+    "experimental/web_dashboard"
     "flutter_maps_firestore"
     "infinite_list"
     "ios_app_clip"
-    "isolate_example"
+    # "isolate_example"
```
filiph, line 21: Can we add a comment saying why we're (temporarily) skipping this? Otherwise, people will be afraid to uncomment it AND remove it completely, because they won't know why it's skipped.

filiph (approved):
LGTM with one comment. I'm pleasantly surprised there was no Dart code change in `veggieseasons` after so many upgrades to dependencies.

tool/flutter_ci_script_dev.sh
filiph: (Alternately, just remove the line altogether.)

RedBrogdon: Good point. I'll add comments -- we've done that before, and it's how I knew to add `multiple_flutters_module` back to the list.

PR #796, opened 2021-05-18
Change router examples to use SynchronousFuture

goderbauer:
Some methods of the Router API support asynchronous processing. However, if the information is available synchronously, it's best to just return a SynchronousFuture to indicate that (also documented, e.g. here: https://master-api.flutter.dev/flutter/widgets/RouterDelegate/setNewRoutePath.html). 

What's the difference? Let's say we link directly to `/book/1`. When `parseRouteInformation` is async, the Router will render for one frame the home page `/` because the Future returned by `parseRouteInformation` with the correct route information will not complete until the next frame. When the future then complete in the next frame, `/book/1` is rendered as expected.

When `parseRouteInformation` returns a SynchronousFuture, the information that `/book/1` should be rendered is available in the first frame and the Router can render `/book/1` correctly even for the first frame.

### DIFF
--- a/navigation_and_routing/lib/router/router.dart
+++ b/navigation_and_routing/lib/router/router.dart
@@ -5,6 +5,7 @@
 /// Full sample that shows a custom RouteInformationParser and RouterDelegate
 /// parsing named routes and declaratively building the stack of pages for the
 /// [Navigator].
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
 void main() {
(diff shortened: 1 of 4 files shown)


johnpryan:
Thanks for the explanation - I didn't realize this.  I probably thought this was okay because functions marked async run synchronously until the first await statement (https://github.com/dart-lang/sdk/issues/30345). Here's a [DartPad example](http://dartpad.dev/0999eac181958e74196f814e9566c6ef) to show what I mean.

johnpryan (approved):

johnpryan:
@goderbauer @chunhtai what if we used FutureOr instead? https://github.com/flutter/flutter/issues/82827

PR #797, opened 2021-05-18
Use a newer version of gradle to avoid an AGP bug around NDK

xster:
Update the gradle and AGP version used in some samples to avoid the AGP bug described in https://github.com/flutter/flutter/issues/53657.


RedBrogdon (approved):
LGTM.

Issue #731, opened 2021-02-23
[testing_app] Migrate to null-safety


abd99:
The `integration_test` package was moved to the `flutter/flutter` repo a few months back. However, its version on the beta channel doesn't have null-safety implemented.
```
➜  testing_app git:(testing_app_null_safety) ✗ dart pub outdated --mode=null-safety
Showing dependencies that are currently not opted in to null-safety.
[✗] indicates versions without null safety support.
[✓] indicates versions opting in to null safety.

Package Name      Current  Upgradable  Resolvable           Latest               

direct dependencies:
cupertino_icons   ✗1.0.0   ✓1.0.2      ✓1.0.2               ✓1.0.2               
provider          ✗4.3.3   ✗4.3.3      ✓5.0.0-nullsafety.5  ✓5.0.0-nullsafety.5  

dev_dependencies:
integration_test  ✗(sdk)   ✗(sdk)      ✗(sdk)               ✗(sdk)               

1 upgradable dependency is locked (in pubspec.lock) to an older version.
To update it, use `dart pub upgrade`.

1 dependency is constrained to a version that is older than a resolvable version.
To update it, edit pubspec.yaml, or run `dart pub upgrade --null-safety`.
```

Creating this issue as a reminder to migrate the testing sample once the integration_test package is null-safe on the stable/beta channels.

Swayam221:
Can I be guided to work on this?

abd99:
Hello @Swayam221,

Thanks for showing interest in contributing to the repo.

As mentioned in the issue description, there's a dependency for this sample that isn't migrated yet. We'll have to wait for it migrate first. Then, we can migrate this sample.

PR #800, opened 2021-05-18
[testing_app] Migrate to null safety

abd99:
- Upgraded dependencies
- Migrated sample using dart migrate command
- Changed a few things to better leverage null safety

closes: #731 

/cc @domesticmouse @RedBrogdon


domesticmouse:
Thanks again @abd99!

PR #803, opened 2021-05-29
[platform_channels] Migrate to sound null safety

AyushBherwani1998:
## Description
Migrates platform_channels sample to sound null safety


AyushBherwani1998:
I don't know why tests are failing on returning `null` as message reply. It does take `null` as a return value because it's has a return type of `Future<String?>` and Future<ByteData?>.

RedBrogdon (changes requested):
The parameter type for the handled is `String?`, but if you check the SDK source, I believe you'll find it strips the nullability for the return type of the handler:

https://github.com/flutter/flutter/blob/c99ed373b3d254f9b10789ddf3771c3aa39e48ca/packages/flutter/lib/src/services/platform_channel.dart#L68

It looks to me like the BasicMessageChannel itself is getting its type inferred differently than you'd like it. If you assign the `BasicMessageChannel` on line 25 to a variable and then mouse over that variable in your IDE (no need to actually run the code), I believe you'll find it typed as `BasicMessageChannel<String>` rather than the `BasicMessageChannel<String?>` you're expecting.

If you manually add the generic type param:

```dart
    BasicMessageChannel<String?>('stringCodecDemo', StringCodec())
        .setMessageHandler((message) async {
      // ...
    });
```

That fixes the issue, at least on my machine. There's a couple more instances in the test file that need the same treatment (one with `String?` and one with `ByteData?`).

RedBrogdon:
Thanks for doing this migration, by the way. I hope you and your family are doing well.

AyushBherwani1998:
> I believe you'll find it typed as `BasicMessageChannel<String>` rather than the `BasicMessageChannel<String?>` you're expecting.

I did try this, and weirdly it shows me `BasicMessageChannel<String?>`.
Screenshot 2021-06-05 at 11 45 19 AM

AyushBherwani1998:
> Thanks for doing this migration, by the way. I hope you and your family are doing well.
Thank you, dad and aunt tested positive last week. Currently they are in isolation and recovering well.

RedBrogdon:
Well that's just weird: 

image

I'd love to know why we're seeing different info in the popups.

RedBrogdon (approved):

AyushBherwani1998:
@RedBrogdon I'm using VS Code. If it shows `BasicMessageChannel<String?>` in VS Code then there might be some bug? I guess @DanTup works on the Dart and Flutter extensions for VS Code. Maybe he can help a little to figure out.

DanTup:
@AyushBherwani1998 can you confirm which version of Flutter you're using? Can you reproduce this (perhaps after restarting VS Code)? I tried with both latest stable and master channels of Flutter, and always see `String`:

Screenshot 2021-06-08 at 09 41 07

I can't think of a reason why it should show `String?`. Since `StringCodec` is `MessageCodec<String>` I would expect it to match that. A long shot - is it possible during the migration you inadvertently modified your local source for `StringCodec` and added a `?`? If not, let me know which SDK version you're on and I'll try that. Thanks!

DanTup:
I'm also curious about the first line of your tooltip only showing "basicMessageChannel" whereas mine shows the full definition. I wonder if you're on an older version of Dart-Code - although I wouldn't expect it to affect this.

AyushBherwani1998:
@DanTup I'm not able to reproduce this now. While working on this PR I was on the stable v2.0.3.

PR #811, opened 2021-06-06
Add `flutter_lints` to `add_to_app/multiple_flutters/multiple_flutters_module`

domesticmouse:
And another

### DIFF
--- a/add_to_app/multiple_flutters/multiple_flutters_module/analysis_options.yaml
+++ b/add_to_app/multiple_flutters/multiple_flutters_module/analysis_options.yaml
@@ -1,0 +1,19 @@
+include: package:flutter_lints/flutter.yaml
+
+analyzer:
+  strong-mode:
+    implicit-casts: false
+    implicit-dynamic: false
+
+linter:
+  rules:
+    avoid_types_on_closure_parameters: true
+    avoid_void_async: true
+    cancel_subscriptions: true
+    close_sinks: true
+    directives_ordering: true
+    package_api_docs: true
+    package_prefixed_library_names: true
+    test_types_in_equals: true
+    throw_in_finally: true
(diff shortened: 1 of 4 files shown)


add_to_app/multiple_flutters/multiple_flutters_module/analysis_options.yaml
```diff
@@ -0,0 +1,19 @@
...
+    avoid_types_on_closure_parameters: true
+    avoid_void_async: true
+    cancel_subscriptions: true
+    close_sinks: true
+    directives_ordering: true
+    package_api_docs: true
+    package_prefixed_library_names: true
+    test_types_in_equals: true
+    throw_in_finally: true
+    unnecessary_statements: true
```
gaaclarke, line 19: Should we pull this into a shared file somewheres?  Looks like you are adding these same rules in a few places.

gaaclarke (approved):

add_to_app/multiple_flutters/multiple_flutters_module/analysis_options.yaml
domesticmouse: What are our options here? Can we create a template file in the top level of the samples repo and include it in all the samples? It would simplify maintenance going forward, but it might be a tad confusing to new users.

/cc @RedBrogdon

RedBrogdon: My two cents is to use individual files in the projects. @gaaclarke's right that we'd be repeating ourselves, but we'd also be keeping each sample as a discrete entity and modeling the correct way to place an analysis options file in a project.

In other words, I'd prefer to be more beginner-friendly at the cost of duplicating the analysis options files.

PR #827, opened 2021-06-07
Add `flutter_lints` to `navigation_and_routing`

domesticmouse:
And rebuild runners for bonus points.

PR #832, opened 2021-06-09
Replace navigation_and_routing with a new sample

johnpryan:
This replaces the samples in `navigation_and_routing/` with a new sample, and moves the old snippets into `navigation_and_routing/old_snippets`.

This sample implements the four routing [scenarios](https://github.com/flutter/uxr/tree/master/nav2-usability/storyboards) studied by the [API usability research](https://github.com/flutter/uxr/wiki/Navigator-2.0-API-Usability-Research) being conducted in the [flutter/uxr](github.com/flutter/uxr) repo.

- **Deep (dynamic) linking** - Navigate to `/book/:bookId` or `author/:authorId`. The `TemplateRouteParser` class uses [package:path_to_regexp](https://pub.dev/packages/path_to_regexp) to parse the route into the `ParsedRoute` class, which is used throughout the app.
- **Skipping stacks** - Go to the settings page and click on "Go directly to /book/0"
- **Validation** - attempt to change the url at the Sign in screen. 
- **Nested routing** - The `/books`, `/authors`, and `/settings` pages are controlled via an `AdaptiveScaffold` from [package:adaptive_navigation](https://pub.dev/packages/adaptive_navigation). The body of the adaptive scaffold builds a nested Navigator that use a `FadeTransitionPage` to fade between the corresponding pages.

This also uses the [Link](https://pub.dev/documentation/url_launcher/latest/link/Link-class.html) widget from the [url_launcher](https://pub.dev/packages/url_launcher) package in some places

This is staged at https://flutter-navigation-and-routing.web.app/

To do:
- [x] Set up [deep linking](https://flutter.dev/docs/development/ui/navigation/deep-linking) on iOS and Android
- [x] Clean up UI
- [x] Configure the URL strategy using [package:url_strategy](https://pub.dev/packages/url_strategy)
- [x] apply https://github.com/flutter/samples/pull/827 to the old_snippets/ directory

Might do:
- [ ] State restoration

FYI @filiph @goderbauer @jackkim9 @mariamhas @csells


johnpryan:
Here's a link to  try out the deployed version: https://flutter-navigation-and-routing.web.app/

domesticmouse (approved):

navigation_and_routing/lib/src/auth/auth.dart
```diff
@@ -0,0 +1,59 @@
...
+  @override
+  bool operator ==(Object other) {
+    return other is BookstoreAuth && other._signedIn == _signedIn;
+  }
+
+  @override
+  int get hashCode => _signedIn.hashCode;
+}
+
+class BookstoreAuthScope extends InheritedWidget {
```
chunhtai, line 47: InheritedNotifier maybe?

navigation_and_routing/lib/src/routing/auth_guard.dart
```diff
@@ -0,0 +1,33 @@
...
+class BookstoreRouteGuard implements RouteGuard<ParsedRoute> {
+  BookstoreAuth auth;
+
+  BookstoreRouteGuard({
+    required this.auth,
+  });
+
+  /// Redirect to /signin if the user isn't signed in.
+  @override
+  Future<ParsedRoute> redirect(ParsedRoute from) async {
```
chunhtai, line 19: It feels weird to me that  RouteGuard is applied after the route information parser

navigation_and_routing/lib/src/app.dart
```diff
@@ -0,0 +1,76 @@
...
+    ..addBook('Too Like the Lightning', 'Ada Palmer', false, true)
+    ..addBook('Kindred', 'Octavia E. Butler', true, false)
+    ..addBook('The Lathe of Heaven', 'Ursula K. Le Guin', false, false);
+
+  @override
+  void initState() {
+    guard = BookstoreRouteGuard(auth: auth);
+    routeState = RouteState(routeParser);
+    routerDelegate = BookstoreRouterDelegate(routeState, auth, guard);
+    auth.addListener(() async {
```
chunhtai, line 52: both routerDelegate and auth need to be disposed

navigation_and_routing/lib/src/routing/delegate.dart
```diff
@@ -0,0 +1,144 @@
...
+    auth.addListener(notifyListeners);
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final pathTemplate = routeState.route.pathTemplate;
+    final library = LibraryScope.of(context);
+
+    Book? book;
+    if (pathTemplate == '/books/:bookId') {
```
chunhtai, line 50: I wonder whether we should use ParsedRoute subclass to differentiate the route.

chunhtai: directly comparing the pattern feel weird.

navigation_and_routing/lib/src/routing/route_state.dart
```diff
@@ -0,0 +1,53 @@
...
+        await parser.parseRouteInformation(RouteInformation(location: route));
+  }
+
+  static RouteState of(BuildContext context) {
+    return context.dependOnInheritedWidgetOfExactType<RouteStateScope>()!.state;
+  }
+}
+
+/// Provides the current [RouteState] to descendent widgets in the tree.
+class RouteStateScope extends InheritedWidget {
```
chunhtai, line 41: ditto

navigation_and_routing/lib/src/app.dart
chunhtai: same for routeState

navigation_and_routing/lib/src/screens/scaffold_body.dart
```diff
@@ -0,0 +1,54 @@
...
+            key: ValueKey('settings'),
+            child: SettingsScreen(),
+          )
+        else if (currentRoute.pathTemplate.startsWith('/books') ||
+            currentRoute.pathTemplate == '/')
+          FadeTransitionPage<void>(
+            key: const ValueKey('books'),
+            child: BooksScreen(currentRoute: currentRoute),
+          )
+        //  TODO: determine why the Navigator is built with empty pages when the user is signed out...
```
chunhtai, line 45: We should figure this out before merging this pr

navigation_and_routing/lib/src/routing/delegate.dart
johnpryan: My thinking was that `TemplateRouteParser` is a general-purpose class (not app-specific), so it returns a `ParsedRoute`, which is also general-purpose.

navigation_and_routing/lib/src/routing/guard.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright 2021, the Flutter project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+abstract class RouteGuard<T> {
```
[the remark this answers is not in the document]
johnpryan: nope - I moved it into `parser.dart` since it's part of the TemplateParser API.

navigation_and_routing/lib/src/screens/settings.dart
```diff
@@ -0,0 +1,95 @@
...
+                          uri: Uri.parse('/books/0'),
+                          builder: (context, followLink) {
+                            return TextButton(
+                              child: const Text('Go directly to /books/0'),
+                              onPressed: followLink,
+                            );
+                          },
+                        ),
+                        Link(
+                          uri: Uri.parse('/books/0'),
```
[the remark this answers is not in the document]
johnpryan: good catch! fixed.

navigation_and_routing/macos/Runner/DebugProfile.entitlements
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.app-sandbox</key>
+	<true/>
+	<key>com.apple.security.cs.allow-jit</key>
+	<true/>
+	<key>com.apple.security.network.server</key>
```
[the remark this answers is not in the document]
johnpryan: It's not using networking yet, so I'll leave this for now.

navigation_and_routing/lib/src/app.dart
johnpryan: good catch, fixed.

navigation_and_routing/lib/src/auth/auth.dart
johnpryan: Nice - that's the API I was looking for.

navigation_and_routing/lib/src/screens/scaffold_body.dart
johnpryan: I don't think the framework is doing what I expect here. I'm replacing one page with another, but the `child` for both pages is building during the same build.

Calling `signOut()` changes RouteState from `/settings` to `/signin`, which replaces the Page that builds the BookstoreScaffoldBody (whose parent widget is BookstoreScaffold) with the page that builds the SignInScreen. But I can see the `build()` methods for both of these widgets happening in the same build when I set a breakpoint.

I don't expect the `build()` method for `BookstoreScaffoldBody` to happen after `signOut()`, since it's Page was removed from the the top-level Navigator's pages list, if that makes sense.

navigation_and_routing/lib/src/routing/delegate.dart
johnpryan: After thinking about this some more, I do think it makes sense to look at the path template here. In this case, the same widget gets built for every path matching that pattern. Comparing the ParsedRoute would be for cases where you want to build a different widget depending on the exact path (`/book/1` and `/book/2` would build two different widgets)

chunhtai: Maybe define the string as a const variable somewhere?

navigation_and_routing/lib/src/screens/scaffold_body.dart
chunhtai: It is probably due to page transition, before the page transition ends, both pages will be built because they are on the screen at the same time

johnpryan: It looks like using InheritedNotifier for RouteStateScope causes the build method for BookstoreScaffoldBody to happen when the route changes, so it will handle whenever the global route changes.  I think this is a scenario where more targeted nested routing is needed. Commit 4138241 adds a comment to explain why this is necessary here.

johnpryan:
@chunhtai could you take another look?

navigation_and_routing/lib/src/screens/scaffold_body.dart
```diff
@@ -0,0 +1,62 @@
...
+            currentRoute.pathTemplate == '/')
+          FadeTransitionPage<void>(
+            key: const ValueKey('books'),
+            child: BooksScreen(currentRoute: currentRoute),
+          )
+
+          // Avoid building a Navigator with an empty `pages` list when the
+          // RouteState is set to an unexpected path, such as /signin.
+          //
+          // Since RouteStateScope is an InheritedNotifier, any change to the
```
chunhtai, line 51: Can you delay the page transition duration to a large number like 1 sec, and hit sign out to see what will happen? I am worrying that the page will suddenly become empty before the page transition finishes

johnpryan: Here's a [video](https://drive.google.com/file/d/1i8oBmnFxySWgC7ExuaxWEeJDtj5sg4kq/view?usp=sharing) Yeah this is an issue. 

I think what we need here is  a way to listen for changes to specific routes, rather than just listen to the global `RouteStateScope. This widget only wants to re-build when the route is  to `/authors`, `/settings`, or `/books`.

navigation_and_routing/lib/src/app.dart
```diff
@@ -0,0 +1,101 @@
...
+          child: MaterialApp.router(
+            routerDelegate: routerDelegate,
+            routeInformationParser: routeParser,
+          ),
+        ),
+      ),
+    );
+  }
+
+  Future<void> _handleAuthStateChanged() async {
```
goderbauer, line 88: Wondering: why is this `async` and returning a Future?

navigation_and_routing/lib/src/auth.dart
```diff
@@ -0,0 +1,2 @@
+export 'auth/auth.dart';
```
goderbauer, line 1: nit: license header?

navigation_and_routing/lib/src/app.dart
```diff
@@ -0,0 +1,101 @@
...
+    if (!auth.signedIn) {
+      routeState.go('/signin');
+    }
+  }
+
+  @override
+  void dispose() {
+    auth.removeListener(_handleAuthStateChanged);
```
goderbauer, line 96: dispose auth as well? It's a ChangeNotifier that has a dispose method.

navigation_and_routing/lib/src/auth/auth.dart
```diff
@@ -0,0 +1,53 @@
...
+    await Future<void>.delayed(const Duration(milliseconds: 200));
+
+    // Sign in. Allow any password.
+    _signedIn = true;
+    notifyListeners();
+    return _signedIn;
+  }
+
+  @override
+  bool operator ==(Object other) {
```
goderbauer, line 33: Having custom `operator==` and `hashCode` on a non-immutable object is strange and can yield undesired/unpredictable behavior. 

See also: https://dart.dev/guides/language/effective-dart/design#avoid-defining-custom-equality-for-mutable-classes

Do you actually need this?

navigation_and_routing/lib/src/auth/auth.dart
```diff
@@ -0,0 +1,53 @@
...
+}
+
+class BookstoreAuthScope extends InheritedNotifier<BookstoreAuth> {
+  const BookstoreAuthScope({
+    required BookstoreAuth notifier,
+    required Widget child,
+    Key? key,
+  }) : super(key: key, notifier: notifier, child: child);
+
+  static BookstoreAuth? of(BuildContext context) {
```
goderbauer, line 48: I would consider making the return type here non-nullable and make this throw a meaningful error if no BookstoreAuth can be found. This avoids the `!` when you call `BookstoreAuthScope.of` and also avoids the mental overhead there of wondering if this can ever be null.

navigation_and_routing/lib/src/data.dart
```diff
@@ -0,0 +1,3 @@
+export 'data/author.dart';
```
goderbauer, line 1: nit: licence

navigation_and_routing/lib/src/data/book.dart
```diff
@@ -0,0 +1,15 @@
...
+import 'author.dart';
+
+class Book {
+  final int id;
+  final String title;
+  late final Author author;
+  final bool isPopular;
+  final bool isNew;
+
+  Book(this.id, this.title, this.isPopular, this.isNew);
```
goderbauer, line 14: nit: consider making the boolean arguments named required for better readability on the call side.

navigation_and_routing/lib/src/data/library.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'package:collection/collection.dart';
+
+import 'author.dart';
+import 'book.dart';
+
+class Library {
+  final List<Book> allBooks = [];
+  final List<Author> allAuthors = [];
+
+  void addBook(String title, String authorName, bool isPopular, bool isNew) {
```
goderbauer, line 14: nit: named required arguments for the booleans?

navigation_and_routing/lib/src/routing.dart
```diff
@@ -0,0 +1,4 @@
+export 'routing/delegate.dart';
```
goderbauer, line 1: nit: license

navigation_and_routing/lib/src/routing/delegate.dart
```diff
@@ -0,0 +1,52 @@
...
+  final WidgetBuilder builder;
+
+  @override
+  final GlobalKey<NavigatorState> navigatorKey;
+
+  SimpleRouterDelegate({
+    required this.routeState,
+    required this.builder,
+    required this.navigatorKey,
+    // ignore: prefer_initializing_formals
```
goderbauer, line 25: Is this actually ignoring an error here?

navigation_and_routing/lib/src/routing/parser.dart
```diff
@@ -0,0 +1,68 @@
...
+      }
+    }
+
+    // Redirect if a guard is present
+    var guard = this.guard;
+    if (guard != null) {
+      return guard.redirect(parsedRoute);
+    }
+
+    return parsedRoute;
```
goderbauer, line 61: This could be wrapped in a SynchronousFuture, I think, since there appears to be no asynchrony in this method if guard is null.

navigation_and_routing/lib/src/routing/route_state.dart
```diff
@@ -0,0 +1,50 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+
+import 'parsed_route.dart';
+import 'parser.dart';
+
+/// The current route state. To change the current route, call obtain the state using
```
goderbauer, line 11: nit: extra word "call" in this line?

navigation_and_routing/lib/src/routing/route_state.dart
```diff
@@ -0,0 +1,50 @@
...
+/// `RouteState.of(context)` and call `go()`:
```
goderbauer, line 12: RouteStateScope instead of RouteState, here and below? (or move the implementation of `of`)?

navigation_and_routing/lib/src/routing/route_state.dart
```diff
@@ -0,0 +1,50 @@
...
+  TemplateRouteParser parser;
+  ParsedRoute _route;
+
+  RouteState(this.parser)
+      : _route = parser.initialRoute;
+
+  ParsedRoute get route => _route;
+
+  set route(ParsedRoute route) {
+    _route = route;
```
goderbauer, line 27: not sure, but is is worthwhile to check whether route is different from _route and only in that case notifying listeners?

navigation_and_routing/lib/src/routing/route_state.dart
```diff
@@ -0,0 +1,50 @@
...
+
+/// Provides the current [RouteState] to descendent widgets in the tree.
+class RouteStateScope extends InheritedNotifier<RouteState> {
+  const RouteStateScope({
+    required RouteState notifier,
+    required Widget child,
+    Key? key,
+  }) : super(key: key, notifier: notifier, child: child);
+
+  static RouteState? of(BuildContext context) {
```
goderbauer, line 45: Same as with the other `of` method: I would consider making the return here non-nullable.

navigation_and_routing/lib/src/screens/author_details.dart
```diff
@@ -0,0 +1,49 @@
...
+        title: Text(author.name),
+      ),
+      body: Center(
+        child: Column(
+          children: [
+            Expanded(
+              child: BookList(
+                books: author.books,
+                onTap: (book) {
+                  Navigator.of(context).push<dynamic>(
```
goderbauer, line 32: Is it on purpose that this uses the page-less imperative API of the navigator? Why is it not navigating to `/books/:bookId`?

navigation_and_routing/lib/src/screens/book_details.dart
```diff
@@ -0,0 +1,75 @@
...
+    Key? key,
+    this.book,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    if (book == null) {
+      return const Scaffold(
+        body: Center(
+          child: Text('No book with found.'),
```
goderbauer, line 24: nit: is there something missing after "with"?

navigation_and_routing/lib/src/screens/sign_in.dart
```diff
@@ -0,0 +1,76 @@
...
+            padding: const EdgeInsets.all(8),
+            child: Column(
+              mainAxisAlignment: MainAxisAlignment.center,
+              mainAxisSize: MainAxisSize.min,
+              children: [
+                Text('Sign in', style: Theme.of(context).textTheme.headline4),
+                TextField(
+                  decoration: const InputDecoration(labelText: 'Username'),
+                  onChanged: (v) {
+                    setState(() {
```
goderbauer, line 46: Instead of rebuilding this widget on every keypress, maybe use two TextEditingController and grab their value only when the user hits sign-in?

navigation_and_routing/lib/src/screens/navigator.dart
```diff
@@ -0,0 +1,111 @@
+import 'package:collection/collection.dart';
```
goderbauer, line 1: nit: license header

navigation_and_routing/lib/src/screens/books.dart
```diff
@@ -0,0 +1,130 @@
...
+        break;
+      case 0:
+      default:
+        routeState.go('/books/popular');
+        break;
+    }
+  }
+
+  @override
+  void didUpdateWidget(BooksScreen oldWidget) {
```
goderbauer, line 119: I don't understand why this logic is in `didUpdateWidget`. Instead of putting the logic here, shouldn't this widget listen to the routeState and execute this logic whenever `routeState` notifies its listeners?

goderbauer (commented):
Pretty cool and comprehensive example. Thanks for putting this together!

navigation_and_routing/lib/src/app.dart
chunhtai: I think I missed this in my previous review
The auth.signedIn should be an async getter as we are mimicking the real world authentication services.  If the auth.signedIn is async, the below check should be
```
if (!await auth.signedIn)
```
Thus this function need to be async as well

goderbauer: Wondering, why would `auth.signedIn` be async? When can you not synchronously determine whether you're signed in or not?

PR #834, opened 2021-06-09
Null Safety migrate `add_to_app/books/flutter_module_books`

domesticmouse:


domesticmouse:
Hey Aaron, I'm attempting (and failing) to null safe migrate this codebase. Any chance of giving it a quick once over to see if you can spot how I'm holding it wrong?

brett

gaaclarke:
Here's the error:
```
══╡ EXCEPTION CAUGHT BY GESTURE ╞═══════════════════════════════════════════════════════════════════
The following _TypeError was thrown while handling a gesture:
type 'Null' is not a subtype of type 'Future<void>'

When the exception was thrown, this was the stack:
#0      MockHostBookApi.cancel (package:flutter_module_books/api.dart:72:16)
#1      _BookDetailState.build.<anonymous closure> (package:flutter_module_books/main.dart:115:22)
#2      _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:989:21)
#3      GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:182:24)
#4      TapGestureRecognizer.handleTapUp (package:flutter/src/gestures/tap.dart:607:11)
```

Here is the line of code it is complaining about: https://github.com/flutter/samples/pull/834/files#diff-c0a7338531a5fb82951a8faea02e56991b7b258f8d59014a434e7f9e0cd548faR72

Based off what is said [here](https://stackoverflow.com/a/67683803) it sounds like we are calling a method on a mock object that doesn't have an implementation, so the default handler returns `null` which violates the signature.

You can see we don't have an implementation for the mock [here](https://github.com/flutter/samples/pull/834/files#diff-c4e1e371673d5c0e75708b5f2dd59a319af61d153dc9add4e9a1601040d1374fR43).

You might have to upgrade mockito and switch to the codegen mocks, I'm not sure.

PR #840, opened 2021-06-30
Add Linting tool starter project

abd99:
I haven't added Android or iOS runners as this sample is dedicated to desktop platforms only. Not sure about web though. I think we should keep it until we hit any blocker that requires it to be removed. WDYT @domesticmouse?

Let me know if I have missed something.

/cc @RedBrogdon


domesticmouse (approved):
Approving this as a starting point

PR #841, opened 2021-07-03
Adding new flag for dev channel format

RedBrogdon:
Changes the `ci_projects` function in `flutter_ci_script_shared.sh` to use the `-o none` option in place of the deprecated `-n` when testing with the dev channel.

### DIFF
--- a/tool/flutter_ci_script_shared.sh
+++ b/tool/flutter_ci_script_shared.sh
@@ -1,5 +1,14 @@
 function ci_projects () {
     local channel="$1"
+
+    # The `-n` option has been replaced with a more flexible `-o` for "output"
+    # in dev.
+    if [ $channel == "dev" ]; then
+        format_option="-o none"
+    else
+        format_option="-n"
+    fi
+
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):
I'm curious what is driving this change. And I'm guessing we are going to have to replicate it everywhere.

RedBrogdon:
@jonahwilliams is, I believe, the author of the change and may have more context.

jonahwilliams:
Sorry, i missed this while I was OOO.

I would strongly, strongly recommend not using `flutter format` at all and just using `dart format`. We don't really control how the formatter behaves or its flags.

RedBrogdon:
Hrm. That could be a problem, since part of this repo's purpose is to model good CI behavior that other devs can copy.

Is flutter format something we shouldn't recommend in general?

domesticmouse:
https://github.com/flutter/codelabs/pull/146 and https://github.com/flutter/samples/pull/846

PR #846, opened 2021-07-09
`flutter format` -> `dart format`

domesticmouse:


RedBrogdon (approved):
LGTM. While you're in here, @jonahwilliams also recommended using `dart analyze` instead of `flutter analyze`.

jonahwilliams:
I'm not certain that the functionality between the two is 100% yet, since there is no watch mode (I dunno if you are using that). If not, I would switch to the dart CLI

PR #851, opened 2021-07-12
Update Navigation and Routing sample

johnpryan:
This adds a README and addresses code review comments that were missed in the [first PR](https://github.com/flutter/samples/pull/832).

@chunhtai the only change I didn't make was to make the `BookstoreAuth.signedIn` getter async. Firebase for example has a [synchronous `user` getter](https://firebase.google.com/docs/auth/web/manage-users) to check if a user is signed in.

### DIFF
--- a/navigation_and_routing/lib/src/routing/parser.dart
+++ b/navigation_and_routing/lib/src/routing/parser.dart
@@ -7,7 +7,7 @@ import 'package:path_to_regexp/path_to_regexp.dart';
 
 import 'parsed_route.dart';
 
-/// Used by `TemplateRouteParser` to guard access to routes.
+/// Used by [TemplateRouteParser] to guard access to routes.
 ///
 /// Override this class to change the route that is returned by
 /// [TemplateRouteParser.parseRouteInformation] if a condition is not met, for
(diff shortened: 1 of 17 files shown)


navigation_and_routing/lib/src/routing/parser.dart
```diff
@@ -7,6 +7,11 @@ import 'package:path_to_regexp/path_to_regexp.dart';
 
 import 'parsed_route.dart';
 
+/// Used by `TemplateRouteParser` to guard access to routes.
```
goderbauer, line 10: to get linking:

```suggestion
/// Used by [TemplateRouteParser] to guard access to routes.
```

navigation_and_routing/lib/src/routing/route_state.dart
```diff
@@ -8,8 +8,8 @@ import 'package:flutter/widgets.dart';
 import 'parsed_route.dart';
 import 'parser.dart';
 
-/// The current route state. To change the current route, call obtain the state using
-/// `RouteState.of(context)` and call `go()`:
+/// The current route state. To change the current route, call obtain the state
+/// using `RouteState.of(context)` and call `go()`:
```
goderbauer, line 12: here and below: Isn't the `of` method on RouteStateScope instead of RouteState?

goderbauer (approved):
LGTM

PR #857, opened 2021-07-23
Did you mean to use .go here?

idkq:


RedBrogdon:
@johnpryan, this is your sample, so I'm sending the review your way.

johnpryan:
Indeed I was, thanks @idkq!

PR #858, opened 2021-07-28
Update README.md

domesticmouse:
Hey @gaaclarke, using a flutter hash as the required version is a tad baroque. Can we just say Flutter v2 is the minimum?


RedBrogdon (approved):
I think we can probably just run with this.

gaaclarke:
Sorry, just seeing this.  I think that's correct, v2.  Thanks.

PR #867, opened 2021-08-06
Fix <base href="/"> tags in hosted samples, if required.

johnpryan:
Each sample's `<base>` tag needs to point to the path the app is being hosted at. This creates a tool that runs after each sample is built to fix these tags.

This is currently causing the [navigation_and_routing](https://flutter.github.io/samples/navigation_and_routing.html) sample to break

### DIFF
--- a/web/_tool/build_ci.dart
+++ b/web/_tool/build_ci.dart
@@ -5,6 +5,7 @@
 import 'dart:io';
 import 'package:path/path.dart' as p;
 import 'common.dart';
+import 'fix_base_tags.dart';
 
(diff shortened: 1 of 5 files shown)


domesticmouse (approved):

domesticmouse:
Quick heads up, this is going to have to be updated for Flutter 2.5 (currently on the dev channel). The base changed from a hard coded `/` to a interpolated variable.

johnpryan:
Ah, that's right... I think this will still work but probably won't be
necessary if we regenerate the project files for each sample

PR #869, opened 2021-08-09
[linting_tool] Implement exporting profiles

abd99:
/cc @domesticmouse

### DIFF
--- a/experimental/linting_tool/lib/pages/saved_lints_page.dart
+++ b/experimental/linting_tool/lib/pages/saved_lints_page.dart
@@ -45,7 +45,7 @@ class SavedLintsPage extends StatelessWidget {
               ),
               itemCount: profilesStore.savedProfiles.length,
               cacheExtent: 5,
-              itemBuilder: (context, index) {
+              itemBuilder: (itemBuilderContext, index) {
                 var profile = profilesStore.savedProfiles[index];
                 return ListTile(
                   title: Text(
@@ -74,10 +74,19 @@ class SavedLintsPage extends StatelessWidget {
                       ),
                       PopupMenuButton<String>(
                         icon: const Icon(Icons.more_vert),
-                        onSelected: (value) {
+                        onSelected: (value) async {
                           switch (value) {
                             case 'Export file':
-                              // TODO(abd99): Implement exporting files.
+                              var saved = await profilesStore
+                                  .exportProfileFile(profile);
+
+                              if (!saved) {
+                                _showSnackBar(
+                                  context,
+                                  profilesStore.error ?? 'Failed to save file.',
+                                );
+                              }
+
                               break;
                             case 'Delete':
                               profilesStore.deleteProfile(profile);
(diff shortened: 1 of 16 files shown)


experimental/linting_tool/lib/pages/saved_lints_page.dart
```diff
@@ -45,7 +45,7 @@ class SavedLintsPage extends StatelessWidget {
               ),
               itemCount: profilesStore.savedProfiles.length,
               cacheExtent: 5,
-              itemBuilder: (context, index) {
+              itemBuilder: (itemBuilderContext, index) {
```
domesticmouse, line 48: I'm curious why the `context` rename here

experimental/linting_tool/lib/repository/api_provider.dart
```diff
@@ -5,14 +5,15 @@
...
+import 'package:yaml/yaml.dart';
 
 class APIProvider {
   final _baseURL = 'https://dart-lang.github.io/linter';
   final http.Client httpClient;
   APIProvider(this.httpClient);
 
   Future<List<Rule>> getRulesList() async {
-    http.Response response =
+    var response =
```
domesticmouse, line 16: `final`? =)

domesticmouse (approved):

experimental/linting_tool/lib/pages/saved_lints_page.dart
abd99: I'm passing the `context` to a `ScaffoldMessenger`(to show a SnackBar) later in the code. If it isn't renamed, this context gets passed which causes the following error:
```
Unhandled Exception: Looking up a deactivated widget's ancestor is unsafe.
At this point the state of the widget's element tree is no longer stable.
To safely refer to a widget's ancestor in its dispose() method, save a reference to the ancestor by calling dependOnInheritedWidgetOfExactType() in the widget's didChangeDependencies() method.
```

experimental/linting_tool/lib/repository/api_provider.dart
abd99: Sure.

abd99: I wonder why we don't have lint rules enabled to pick this up? Should I create a new PR for this?

experimental/linting_tool/lib/pages/saved_lints_page.dart
domesticmouse: I'll look more closely at this after it is landed.

experimental/linting_tool/lib/repository/api_provider.dart
domesticmouse: I suspect we are missing a lint rule. What lints are setup for this project?

abd99: I have used the default lints that are used in most of the projects in the repo.
[Example](https://github.com/flutter/samples/blob/master/veggieseasons/analysis_options.yaml)

PR #870, opened 2021-08-09
[linting_tool] Stop duplication of rules in the same profile

abd99:
/cc @domesticmouse

### DIFF
--- a/experimental/linting_tool/lib/model/profiles_store.dart
+++ b/experimental/linting_tool/lib/model/profiles_store.dart
@@ -49,6 +49,9 @@ class ProfilesStore extends ChangeNotifier {
   }
 
   Future<void> addToExistingProfile(RulesProfile profile, Rule rule) async {
+    // TODO(abd99): Consider refactoring to LinkedHashSet/SplayTreeSet to avoid
+    // duplication automatically.
+    // ref: https://github.com/flutter/samples/pull/870#discussion_r685666792
     var rules = profile.rules;
     if (!rules.contains(rule)) {
       rules.add(rule);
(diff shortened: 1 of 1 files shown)


experimental/linting_tool/lib/model/profiles_store.dart
```diff
@@ -49,8 +49,12 @@ class ProfilesStore extends ChangeNotifier {
   }
 
   Future<void> addToExistingProfile(RulesProfile profile, Rule rule) async {
-    RulesProfile newProfile =
-        RulesProfile(name: profile.name, rules: profile.rules..add(rule));
+    var rules = profile.rules;
```
domesticmouse, line 52: Should rules be a `set` (assuming `rule` implements equality and hash and all that jazz)?

domesticmouse (approved):

experimental/linting_tool/lib/model/profiles_store.dart
abd99: The `Rule` class extends `Equatable`, which handles this for us(at least for now). Do you think that it might be a good idea to refactor this to a `Set` anyway?

domesticmouse: I'd use https://api.dartlang.org/stable/dart-collection/LinkedHashSet-class.html to get set semantics while retaining ordering.

Actually, it's probably worth thinking about using https://api.dartlang.org/stable/dart-collection/SplayTreeSet-class.html and order rules lexicographically

abd99: Thanks a lot for this!
I have added a TODO for the time being to remind me of the refactor later on.

domesticmouse (approved):

PR #872, opened 2021-08-14
Updates samples for latest beta release

RedBrogdon:
* Uncomments any commented-out apps in the test scripts for beta and dev
* Updates pubspecs
* Fixes a few minor errors caused by deprecations
* Sorts a bunch of imports, which is suddenly a thing the analyzer cares about
* Removes mockito from flutter_books, which removes a version incompatibility
* Updates the charts_flutter version in web_dashboard to be HEAD directly from its repo
  - The most recent code there hasn't been published. We'll need to return and fix that, ideally prior to the next stable.


experimental/web_dashboard/lib/src/auth/firebase.dart
```diff
@@ -2,9 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:firebase_auth/firebase_auth.dart' hide FirebaseUser;
```
domesticmouse, line 5: i'm curious why `FirebaseUser` isn't triggering a lint warning

domesticmouse (approved):

experimental/web_dashboard/lib/src/auth/firebase.dart
RedBrogdon: Because it's not actually referenced in the file, you mean?

Issue #877, opened 2021-08-19
add_to_app books sample cannot resolve mockito dependency


bryanoltman:
When running `flutter pub get` from the `add_to_app/books` directory, it fails with the following error:

```
Because no versions of mockito match >4.1.1+1 <4.1.2 and mockito >=4.0.0 <=4.1.1 depends on test_api ^0.2.1, mockito >=4.0.0 <4.1.1+1 or >4.1.1+1 <4.1.2-∞ requires test_api ^0.2.1.
And because mockito >=4.1.2 <=5.0.0-nullsafety.7 depends on test_api ^0.2.19-nullsafety, mockito >=4.0.0 <4.1.1+1 or >4.1.1+1 <5.0.0-∞ requires test_api ^0.2.1.
And because mockito 4.1.1+1 depends on test_api >=0.2.1 <0.4.0 and every version of flutter_test from sdk depends on test_api 0.4.2, flutter_test from sdk is incompatible with mockito >=4.0.0 <=5.0.0-nullsafety.7.
So, because flutter_module_books depends on both mockito ^4.1.1 and flutter_test any from sdk, version solving failed.
Running "flutter pub get" in flutter_module_books...
pub get failed (1; So, because flutter_module_books depends on both mockito ^4.1.1 and flutter_test any from sdk, version solving failed.)
```

jmagman:
It seems to be working in CI https://github.com/flutter/samples/runs/3366765137?check_suite_focus=true#step:5:123
Can we get the output of `flutter doctor -v`?

bryanoltman:
Sure thing!

```
[✓] Flutter (Channel master, 2.4.0-5.0.pre.59, on macOS 11.5.1 20G80 darwin-x64, locale en-US)
    • Flutter version 2.4.0-5.0.pre.59 at /Users/bryanoltman/Documents/flutter
    • Upstream repository git@github.com:bryanoltman/flutter.git
    • Framework revision f71ceb77fe (5 weeks ago), 2021-07-15 10:18:49 -0700
    • Engine revision 093c4ed1a3
    • Dart version 2.14.0 (build 2.14.0-312.0.dev)

[!] Android toolchain - develop for Android devices (Android SDK version 31.0.0-rc1)
    ✗ cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
[...]

PR #794, opened 2021-05-08
[Formapp] migration to null safety

arteevraina:
Fixes: #771 

* Updated dependencies.
* Ran dart migrate.

cc @RedBrogdon @domesticmouse .

PR #878, opened 2021-08-19
Upgrade mockito package and dependencies

jmagman:
I think this version of mockito works without null safety migration.  It will require tests to be rewritten once null safety is adopted: https://github.com/flutter/samples/pull/794#issuecomment-855861088

Stop pinning `json_annotation` and `json_serializable` since `flutter_test` also has version constraints for these, and it was hard to get a range that worked for both master and stable.

Fixes https://github.com/flutter/samples/issues/877.


form_app/pubspec.yaml
```diff
@@ -12,14 +12,12 @@ dependencies:
   cupertino_icons: ^1.0.0
   intl: any
   http: ^0.12.0
-  mockito: ^4.1.1
-  json_annotation: ^3.0.0
```
jmagman, line 16: > Stop pinning `json_annotation` and `json_serializable` since `flutter_test` also has version constraints for these, and it was hard to get a range that worked for both master and stable.

@RedBrogdon Is this a reasonable thing to do?  Or do you have a better suggestion?

You can see the problem if you check out flutter on master and then flutter on stable.  They want different `json_` version ranges.

xster:
LGTM

form_app/pubspec.yaml
johnpryan: This seems like normal maintenance to me, is there something about `flutter_test` that makes this difficult to maintain?

jmagman: Yes, check out https://github.com/flutter/flutter/issues/82280.

jmagman:
> LGTM

Thanks @xster, need a review approval though, assuming @justinmc is also okay with this.

jmagman:
Will be made obsolete by https://github.com/flutter/samples/pull/895.

PR #883, opened 2021-08-26
Update AndroidManifest.xml

MainTobias:
To fix it opening the app in the same window (fe in whatsapp and not going to desired screen but instead only to login screen and then home screen instead of book/0 or book/2).

### DIFF
--- a/navigation_and_routing/android/app/src/main/AndroidManifest.xml
+++ b/navigation_and_routing/android/app/src/main/AndroidManifest.xml
@@ -5,7 +5,7 @@
         android:icon="@mipmap/ic_launcher">
         <activity
             android:name=".MainActivity"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
             android:hardwareAccelerated="true"


johnpryan:
@MainTobias can you help me understand the logic behind this change? The [docs](https://developer.android.com/guide/topics/manifest/activity-element) say:

> An instruction on how the activity should be launched. There are four modes that work in conjunction with activity flags (FLAG_ACTIVITY_* constants) in Intent objects to determine what should happen when the activity is called upon to handle an intent. They are:
"standard"
"singleTop"
"singleTask"
"singleInstance"
> 
> The default mode is "standard".
> As shown in the table below, the modes fall into two main groups, with "standard" and "singleTop" activities on one side, and "singleTask" and "singleInstance" activities on the other. An activity with the "standard" or "singleTop" launch mode can be instantiated multiple times. The instances can belong to any task and can be located anywhere in the activity stack. Typically, they're launched into the task that called startActivity() (unless the Intent object contains a FLAG_ACTIVITY_NEW_TASK instruction, in which case a different task is chosen — see the taskAffinity attribute).
> 
> In contrast, "singleTask" and "singleInstance" activities can only begin a task. They are always at the root of the activity stack. Moreover, the device can hold only one instance of the activity at a time — only one such task.

I'm thinking that there are reasons Flutter defaults to `singleTop`, but I'm not sure what they are right now.

cc: @chunhtai

chunhtai:
cc @blasten

MainTobias:
Yeah well about the default I would change it to but there is probably an argument to be made there. However because this example tries to demonstrate deep linking you cannot go with singleTop. WHY? Because when you open a link with f. e. WhatsApp you have the problem that it will show you the login screen and then go to the home screen instead of the link because of route guard. BUT if you change it to singleTask you can open app/link then login go back to WhatsApp and open the link again. It will then actually open the link without having to relogin which is the desired behaviour.

domesticmouse (approved):

PR #890, opened 2021-08-27
[testing_app] Add separate driver file for performance tests

abd99:
@johnpryan, should I do something similar for the codelab too?


domesticmouse:
@abd99 yes please =)

PR #895, opened 2021-09-08
Lands beta branch into master for 2.5

RedBrogdon:


RedBrogdon:
Looks like we're having an issue with form_app and the web demo deployment. This should have been caught earlier -- is there a reason form_app isn't included in the CI scripts?

place_tracker/ios/.gitignore
```diff
@@ -25,6 +25,7 @@ Flutter/flutter_assets/
 Flutter/flutter_export_environment.sh
 ServiceDefinitions.json
 Runner/GeneratedPluginRegistrant.*
+build/
```
domesticmouse, line 28: I've been putting this in as `/build/` to make sure it's only the generated directory

domesticmouse (approved):

PR #899, opened 2021-09-21
Drop author from pubspecs, organise imports

domesticmouse:


domesticmouse:
@RedBrogdon there is one outstanding failure in the verify Web demos CI run - `form_app`, which needs a `mockito` upgrade. Do you have time to have a look?

johnpryan (approved):

domesticmouse:
Landing this as is. The `mockito` update required in `form_app` can be fixed in a followup PR.

PR #952, opened 2021-11-05
fix the issue of generating more than [catalogLength] items when [cat…

qfgaohao:
…alogLength] is not divisible by [itemsPerPage]

It only checks if [startingIndex] is greater than [catalogLength] in the existing implementation. It doesn't check the index of an item in a page. So it may generate more items than the number specified by [catalogLength]. i.e if [catalogLength] is 113, the existing implementation will still generate 120 items. This PR solves the issue by adding the check for the indices of items in a page.

### DIFF
--- a/infinite_list/lib/src/api/fetch.dart
+++ b/infinite_list/lib/src/api/fetch.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:math';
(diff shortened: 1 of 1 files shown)


johnpryan (approved):

johnpryan:
@qfgaohao We need you to sign the CLA before we can accept this pull request. Thanks!

qfgaohao:
@googlebot I signed it!

PR #997, opened 2022-01-19
Photo search: minor menu item text cleanup

cbracken:
Switches to using an elipsis character instead of three periods.

Eliminates the elipsis after 'About'. This appears to be consistent
across operating systems:

From Apple's Human Interface guidelines:
Use an ellipsis whenever choosing a menu item requires additional input
from the user. The ellipsis character (…) means a dialog or separate
window will open and prompt the user for additional information or to
make a choice.

From Microsoft's UI guidelines:
An ellipsis should be used "only when additional information is required
to perform the action. For example, the commands About, Advanced, Help,
Options, Properties, and Settings must display another window when
clicked, but don't require additional information from the user.
Therefore they don't need ellipses."

Ref: https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/
Ref: https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#using-ellipses

## Pre-launch Checklist

- [X] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [X] I signed the [CLA].
- [X] I read the [Contributors Guide].
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] All existing and new tests are passing.

[Contributors Guide]: https://github.com/flutter/samples/blob/master/CONTRIBUTING.md

### DIFF
--- a/desktop_photo_search/fluent_ui/lib/main.dart
+++ b/desktop_photo_search/fluent_ui/lib/main.dart
@@ -62,7 +62,7 @@ class UnsplashHomePage extends StatelessWidget {
     menubar.setApplicationMenu([
       menubar.Submenu(label: 'Search', children: [
         menubar.MenuItem(
-          label: 'Search...',
+          label: 'Search…',
           onClicked: () {
             showDialog<void>(
               context: context,
@@ -74,7 +74,7 @@ class UnsplashHomePage extends StatelessWidget {
       ]),
       menubar.Submenu(label: 'About', children: [
         menubar.MenuItem(
-          label: 'About...',
+          label: 'About',
           onClicked: () {
             showDialog<void>(
               context: context,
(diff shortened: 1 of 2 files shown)


domesticmouse (commented):
Holding for incoming refactoring

domesticmouse:
I introduced the ellipsis here because it is (was?) used to indicate that a menu item lead to a dialog box. See https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#using-ellipses for details.

cbracken:
Ah yep - makes sense, but oddly this rule doesn't seem to apply to "About Foo" menu items. I have no idea why, but it seems to apply on Linux, Windows, and macOS (see "About this Mac" under the Apple menu, or "About Foo" under the appplication's menu).

In cases where it does apply, we should be using the elipsis character rather than three dots, and shouldn't have a space between it and the menu item.

Chrome:
image

Firefox:
image

gEdit:
image

cbracken:
@domesticmouse ptal. 

Rebased and re-patched in the new code locations and updated the commit comment with the rationale from Apple's HIG and Microsoft's docs.

Sorry for the delay in getting this one updated!

domesticmouse (approved):

domesticmouse:
Thanks for the update!

PR #1006, opened 2022-02-03
Lands beta changes.

RedBrogdon:
Readies us for the next release.

* Updates pubspecs as necessary.
* Bumps Android SDK versions and Kotlin versions.
* Removes the dev channel tests from beta workflow.
* Also removes superfluous "beta-on-beta" tests.

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -12,23 +12,6 @@ defaults:
     shell: bash
 
 jobs:
-  flutter-tests:
-    name: Test Flutter ${{ matrix.flutter_version }} on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        flutter_version: [dev, beta]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-java@v1
-        with:
-          java-version: '12.x'
-      - uses: subosito/flutter-action@4389e6cbc6cb8a4b18c628ff96ff90be0e926aa8
-        with:
-          channel: ${{ matrix.flutter_version }}
-      - run: ./tool/flutter_ci_script_${{ matrix.flutter_version }}.sh
   # Run the stable test script on the beta channel. Since this branch will soon
   # be merged into master as our stable-targeting code, this is the key thing we
   # need to test.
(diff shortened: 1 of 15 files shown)


.github/workflows/beta.yml
```diff
@@ -12,23 +12,6 @@ defaults:
...
-        os: [ubuntu-latest, macos-latest, windows-latest]
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-java@v1
-        with:
-          java-version: '12.x'
-      - uses: subosito/flutter-action@4389e6cbc6cb8a4b18c628ff96ff90be0e926aa8
-        with:
-          channel: ${{ matrix.flutter_version }}
-      - run: ./tool/flutter_ci_script_${{ matrix.flutter_version }}.sh
```
johnpryan, line 31: Why is this being removed?

RedBrogdon: Good callout. I should have included more detail in my PR description.

The beta branch is used to ready changes for upcoming stable releases, so the important thing is to run the "stable" tests with the beta branch code and the beta channel SDK (which is about to become the new stable). The goal is to give us confidence that when the SDK release is made, the code in the beta branch can be merged into master and pass the stable tests with that new release.

We've already got that setup in the next section of the workflow. The bit that I'm cutting out contains dev channel tests (no longer needed) and the "beta" tests that are meant to be run against the beta SDK channel generally, which are good to have in the master workflow but irrelevant here.

johnpryan: 👍  to removing redundancy. 

If I understand our CI correctly, we test the main branch against Flutter [stable and beta](https://github.com/flutter/samples/blob/master/.github/workflows/main.yml#L23), which guarantees that any of the samples in [flutter_ci_script_stable.sh](https://github.com/flutter/samples/blob/master/tool/flutter_ci_script_stable.sh), [flutter_ci_script_stable.sh](https://github.com/flutter/samples/blob/master/tool/flutter_ci_script_beta.sh) work on their respective channels.

Why are there two scripts,`flutter_ci_script_stable.sh` and `flutter_ci_script_beta.sh`? They test the same set of samples same right now. Is the idea that when there's a breaking change on the beta channel we should remove the failing projects from `flutter_ci_script_beta.sh`?

The naming makes it a little harder to understand what's going on. [beta.yaml](https://github.com/flutter/samples/blob/master/.github/workflows/beta.yml) runs [the tests in flutter_ci_script_stable.sh](https://github.com/flutter/samples/blob/master/.github/workflows/beta.yml#L32). But I don't know how to clear this up other than to rename these scripts to `test_all.sh` and `test_only_the_projects_that_we_expect_to_work_on_the_beta_channel.sh`.

PR #1028, opened 2022-02-25
Setup to handle app window on desktop

miquelbeltran:
This PR is a draft to showcase how we could handle the desktop window in the samples repo. In some cases, apps are not designed to be displayed as desktop apps but as mobile apps, in other cases, resizing the windows smaller than a certain size would case rendering issues to appear. For both cases it would be good to have a way to handle that.

The main idea is to use the `window_size` package from the Flutter team. It is currently not part of the Flutter framework (that may change soon™️) so needs to be imported using a git reference.

The code takes care of resizing the app window and centering on the screen, as well as changing the app name in the window frame (from `provider_shopper` to `Provider Demo`).

The default window size could be changed, the current 400x800 looks like this on a 1440p desktop:
Screenshot 2022-02-25 114712

(this has been tested on Linux and Windows)

Any thoughts? how could we improve it?

@RedBrogdon @domesticmouse


miquelbeltran:
the PR seems large because it adds the windows and linux folders, but the imporant part is here: provider_shopper/lib/main.dart

provider_shopper/pubspec.yaml
```diff
@@ -13,6 +13,13 @@ dependencies:
   # Import the provider package.
   provider: ^5.0.0
 
+  # plugin is not yet part of the flutter framework
+  window_size:
+    git:
+      url: https://github.com/google/flutter-desktop-embedding.git
+      path: plugins/window_size
+      ref: 03d957e8b5c99fc83cd4a781031b154ab3de8753
```
domesticmouse, line 21: Any particular reason to pin a git ref?

provider_shopper/lib/main.dart
```diff
@@ -10,11 +13,32 @@ import 'package:provider_shopper/models/catalog.dart';
...
 void main() {
+  setupWindow();
   runApp(const MyApp());
 }
 
+const double kWindowWidth = 400;
+const double kWindowHeight = 800;
+
+void setupWindow() {
+  if (!kIsWeb && (Platform.isWindows || Platform.isLinux || Platform.isMacOS)) {
```
domesticmouse, line 27: Nice!

domesticmouse (approved):

domesticmouse:
Looking at the CI logs I'm guessing you need to amend the `pubspec.yaml` to make the bots happy. https://github.com/flutter/samples/runs/5332720556?check_suite_focus=true#step:5:842

provider_shopper/pubspec.yaml
miquelbeltran: Yes, there's a reason. This package is yet not part of the flutter framework nor it is available on pub.dev. The install instructions are here: https://github.com/google/flutter-desktop-embedding/blob/master/plugins/README.md

domesticmouse: Yes, I understand the need for git dependency. I'm questioning pinning to a specific git hash. As long as the `pubspec.lock` is checked in, that should be enough to replicate builds.

provider_shopper/pubspec.yaml
```diff
@@ -1,5 +1,6 @@
 name: provider_shopper
 description: A shopping app sample that uses Provider for state management.
+publis_to: none
```
domesticmouse, line 3: Typo. Once this is fixed I'm guessing you'll have a green build

domesticmouse (approved):

provider_shopper/pubspec.yaml
miquelbeltran: Argh my bad, I shouldn't be checking PRs Friday evening :)

PR #1036, opened 2022-03-01
Add min window size to fluent_ui desktop app

miquelbeltran:

Here, I have only added the min window size. The window title seems to be from the native code, and since this is a desktop app, the window size is not limited by a maximum size.

From my tests, a minimum 1024 x 800 is necessary, so the UI is not cut off.

Interestingly, this project also contained another desktop only plugin as git reference (the menubar)


domesticmouse (approved):

domesticmouse:
> Interestingly, this project also contained another desktop only plugin as git reference (the menubar)

Yup, menubar is still in flux

PR #1039, opened 2022-03-03
Add desktop window setup to form_app

miquelbeltran:

In this case, the window is set to a fixed size with the aspect ratio of a mobile device, since the app is designed as a mobile app.

Also added Linux and Windows project files.


form_app/lib/main.dart
```diff
@@ -11,9 +15,29 @@ import 'src/sign_in_http.dart';
 import 'src/validation.dart';
 
 void main() {
+  setupWindow();
   runApp(const FormApp());
 }
 
+const double kWindowWidth = 480;
+const double kWindowHeight = 854;
```
domesticmouse, line 23: Rename: https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters

PR #1043, opened 2022-03-04
Add window desktop setup to navigation_and_routing

miquelbeltran:

In this case the app is setup to maintain a mobile-like window size.

### DIFF
--- a/navigation_and_routing/lib/main.dart
+++ b/navigation_and_routing/lib/main.dart
@@ -2,8 +2,12 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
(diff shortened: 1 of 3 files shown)


navigation_and_routing/lib/main.dart
```diff
@@ -19,5 +23,25 @@ void main() {
   setHashUrlStrategy();
   // setPathUrlStrategy();
 
+  setupWindow();
   runApp(const Bookstore());
 }
+
+const double kWindowWidth = 480;
+const double kWindowHeight = 854;
```
domesticmouse, line 31: Rename: https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters

domesticmouse (approved):

PR #1044, opened 2022-03-04
Add window setup to null_safe_app

miquelbeltran:

The min window size is set to 400 x 400 as the sample doesn't really have any real UI, it just shows a string.

Also added the macos and windows folders.


domesticmouse (approved):

null_safety/null_safe_app/lib/main.dart
```diff
@@ -2,10 +2,26 @@
...
 
 import 'goodapp.dart';
 
 void main() {
+  setupWindow();
   runApp(const GoodMyApp());
 }
+
+const double kWindowWidth = 400;
+const double kWindowHeight = 400;
```
domesticmouse, line 19: Rename: https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters

PR #1045, opened 2022-03-04
Add window setup to null_unsafe_app

miquelbeltran:
same as https://github.com/flutter/samples/pull/1044 but for the other project inside the null_safety folder

The min window size is set to 400 x 400 as the sample doesn't really have any real UI, it just shows a string.

Also added the macos and windows folders.


domesticmouse (approved):

null_safety/null_unsafe_app/lib/main.dart
```diff
@@ -2,12 +2,28 @@
...
 import 'package:flutter/material.dart';
+import 'package:window_size/window_size.dart';
 
 void main() {
+  setupWindow();
   runApp(const MyApp());
 }
 
+const double kWindowWidth = 400;
+const double kWindowHeight = 400;
```
domesticmouse, line 17: Rename: https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters

domesticmouse (changes requested):

domesticmouse (approved):

PR #1046, opened 2022-03-04
Add window setup to provider_counter

miquelbeltran:

The app is setup to maintain a mobile-like window size, relatively small since it is just the counter default example. Also adds window title.

### DIFF
--- a/provider_counter/lib/main.dart
+++ b/provider_counter/lib/main.dart
@@ -2,10 +2,15 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:io';
+
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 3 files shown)


provider_counter/lib/main.dart
```diff
@@ -23,6 +28,25 @@ void main() {
   );
 }
 
+const double kWindowWidth = 360;
+const double kWindowHeight = 640;
```
domesticmouse, line 32: Actually I just realised this is an issue: https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters

can you rip these out from your other updates as well?

domesticmouse (changes requested):

provider_counter/lib/main.dart
miquelbeltran: yes, will do that

domesticmouse (approved):

PR #1182, opened 2022-04-29
[beta updates] fullscreen

craiglabenz:
Updates `fullscreen` app for Dart 2.17

### DIFF
--- a/add_to_app/fullscreen/flutter_module/pubspec.lock
+++ b/add_to_app/fullscreen/flutter_module/pubspec.lock
@@ -49,7 +49,7 @@ packages:
       name: collection
       url: "https://pub.dartlang.org"
     source: hosted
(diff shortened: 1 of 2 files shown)


domesticmouse:
I'm curious why this is passing on stable. Did you bump the min dart version to 2.17.0-0?

domesticmouse (approved):
Set the min Dart version to 2.17.0-0

craiglabenz:
Yep, I did update that in this PR. It's way more difficult to tell than would be ideal, though, because I somehow seem to have a few straggler commits from Andrew in all of my beta PRs. Let's get that ironed out soon so these are a lot less noisy, @domesticmouse.

domesticmouse:
That looks waaay betterer

PR #1186, opened 2022-05-01
[beta-updates] animations

craiglabenz:
Updates `animations` for Dart 2.17


animations/lib/main.dart
```diff
@@ -6,6 +6,7 @@ import 'dart:io';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+// ignore: depend_on_referenced_packages
```
craiglabenz, line 9: Ideally, we wouldn't need to continue using this override; but I've never encountered this lint before and am not sure what to do for plugins that come directly from Flutter

domesticmouse: Which package is failing `depend_on_referenced_packages`? By my reading, it's `package:window_size` which definitely should be directly referenced in `pubspec.yaml`, or am I missing something?

domesticmouse: And it is referenced. I'm confused as to why this lint is firing here. @bwilkerson halp?

domesticmouse (approved):

animations/lib/main.dart
craiglabenz: Seems to be a potential bug around git-based dependencies?

PR #1192, opened 2022-05-03
[beta-updates] web_dashboard

craiglabenz:
Updated `web_dashboard` for Dart 2.17


experimental/web_dashboard/lib/src/app.dart
```diff
@@ -51,10 +51,10 @@ class DashboardApp extends StatefulWidget {
         super(key: key);
 
   @override
-  _DashboardAppState createState() => _DashboardAppState();
+  DashboardAppState createState() => DashboardAppState();
```
domesticmouse, line 54: Instead of making the `_DashboardAppState ` objects public, I'd change the return value of the `createState ()` accessor from `_DashboardAppState` to `State<Dashboard>`

domesticmouse (changes requested):
Change the return value of `createState()` to the appropriate `State<Thingey>` instead of making the `State` objects public. They are implementation detail after all.

experimental/web_dashboard/lib/src/app.dart
craiglabenz: Updated all instances in this PR

domesticmouse (approved):

PR #1197, opened 2022-05-03
[beta-updates] ios_app_clip

craiglabenz:
Updated `ios_app_clip` for Dart 2.17

### DIFF
--- a/ios_app_clip/lib/main.dart
+++ b/ios_app_clip/lib/main.dart
@@ -16,10 +16,10 @@ class Demo extends StatefulWidget {
   const Demo({Key? key}) : super(key: key);
 
   @override
-  State<Demo> createState() => DemoState();
+  State<Demo> createState() => _DemoState();
 }
 
-class DemoState extends State<Demo> {
+class _DemoState extends State<Demo> {
   String deviceInfo = '';
 
   @override
--- a/ios_app_clip/lib/main.dart
+++ b/ios_app_clip/lib/main.dart
@@ -13,13 +13,13 @@ void main() {
 // The same content is shown for both the main app target and in the App
 // Clip.
 class Demo extends StatefulWidget {
-  const Demo({Key key}) : super(key: key);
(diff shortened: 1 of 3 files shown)


ios_app_clip/lib/main.dart
```diff
@@ -13,13 +13,13 @@ void main() {
 // The same content is shown for both the main app target and in the App
 // Clip.
 class Demo extends StatefulWidget {
-  const Demo({Key key}) : super(key: key);
+  const Demo({Key? key}) : super(key: key);
 
   @override
-  State<StatefulWidget> createState() => _DemoState();
+  State<Demo> createState() => DemoState();
```
domesticmouse, line 19: `State<Demo> createState() => _DemoState();`

domesticmouse (changes requested):

ios_app_clip/lib/main.dart
craiglabenz: Which change are you looking for here - the return of the leading underscore? If I'm reading this right, that's now frowned upon by the updated set of lints.

domesticmouse: The way I've been writing the code is keeping the `State` object private, with it's leading underscore, and make the return type of the `createState()` method be `State<WidgetName>`. You can see an example of this [on DartPad](https://dartpad.dev/?id=85e77d36533b16647bf9b6eb8c03296d&channel=beta) complying with Flutter beta's linter. Specifically

```dart
class VariousDiscs extends StatefulWidget {
  final int numberOfDiscs;

  const VariousDiscs(this.numberOfDiscs);

  @override
  State<VariousDiscs> createState() => _VariousDiscsState();
}

class _VariousDiscsState extends State<VariousDiscs> {
  // redacted
}
```

craiglabenz: Updated

domesticmouse (approved):

PR #1204, opened 2022-05-03
[beta-updates] platform_channels

craiglabenz:
Updated `platform_channels` for Dart 2.17


platform_channels/lib/src/add_pet_details.dart
```diff
@@ -13,10 +13,10 @@ class AddPetDetails extends StatefulWidget {
   const AddPetDetails({Key? key}) : super(key: key);
 
   @override
-  _AddPetDetailsState createState() => _AddPetDetailsState();
```
domesticmouse, line 16: `State<AddPetDetails> createState() => _AddPetDetailsState();`

domesticmouse (changes requested):

PR #1292, opened 2022-05-30
Try fixing setup-java flake

parlough:
Trying to see if this fixes the setup-java flakes on macOS since the Zulu builds have been reported as flaking recently. temurin builds are also open but unfortunately don't have a JDK 12 build. We should update from JDK 12 anyway as it was not a LTS release and reached end of life 2+ years ago. We'll see if it updating works with this PR. 

## Pre-launch Checklist

- [x] I signed the [CLA].
- [] All existing and new tests are passing.

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -29,8 +29,8 @@ jobs:
       - uses: actions/checkout@f25a3a9f25bd5f4c5d77189cab02ff357b5aedeb
       - uses: actions/setup-java@860f60056505705214d223b91ed7a30f173f6142
         with:
-          distribution: 'zulu'
-          java-version: '12.x'
+          distribution: 'temurin'
+          java-version: '17.x'
       - uses: subosito/flutter-action@d8687e6979e8ef66d2b2970e2c92c1d8e801d7bf
         with:
           channel: beta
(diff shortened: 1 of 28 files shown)


domesticmouse (approved):

parlough:
Seems it is caching old Gradle versions. Not sure how to avoid the cache, but I'll look tomorrow :)

domesticmouse:
> Seems it is caching old Gradle versions. Not sure how to avoid the cache, but I'll look tomorrow :)

Waiting for tomorrow will hopefully timeout the cache 😎

parlough:
It seems the zulu builds are fixed, perhaps they added a Cloudflare exception for GitHub actions, but upgrading past JDK 12 would still be good: https://github.com/flutter/samples/issues/1293

domesticmouse:
> It seems the zulu builds are fixed, perhaps they added a Cloudflare exception for GitHub actions, but upgrading past JDK 12 would still be good.
> 
> 
> 
> I can (or someone else) can look into what they would take later on, so I created an issue to track that work in https://github.com/flutter/samples/issues/1293

Is there any reason not to land this PR as-is? It seems like an improvement over the status quo.

parlough:
This PR includes some non-desired changes from when I was trying to get Gradle to work. I'll need to figure out how to reliably not use the older Gradle version or its cache. It seems some Gradle scripts (or their results) are cached and I'm not yet sure how to refresh these, as using our own actions/cache step with a custom cache-key didn't seem to work.

> Could not open cp_init generic class cache for initialization script '/opt/hostedtoolcache/flutter/packages/flutter_tools/gradle/aar_init_script.gradle' (/home/runner/.gradle/caches/6.7/scripts/8yq6uh13upgpmx5fnozgpb2bk).

domesticmouse:
Ok, I’ll run up a PR that just bumps the Java version to `17.x`

PR #1305, opened 2022-06-14
Drop redundant `ToggleButtonsStateManager` instances

domesticmouse:
If I understand the code correctly, there appears to be redundant `ToggleButtonsStateManager` instances in the tree. There are three instances, all with the same configuration, with two instances underneath a single higher instance. I'm guessing all three wind up with identical state. I hope.

Or am I misunderstanding the intent of the structure?

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/simplistic_editor/lib/main.dart
+++ b/simplistic_editor/lib/main.dart
@@ -310,58 +310,44 @@ class _MyHomePageState extends State<MyHomePage> {
                   child: Row(
                     mainAxisAlignment: MainAxisAlignment.center,
                     children: [
-                      ToggleButtonsStateManager(
-                        isToggleButtonsSelected: _isSelected,
-                        updateToggleButtonsStateOnButtonPressed:
-                            _updateToggleButtonsStateOnButtonPressed,
-                        updateToggleButtonStateOnSelectionChanged:
-                            _updateToggleButtonsStateOnSelectionChanged,
-                        child: Builder(builder: (innerContext) {
-                          final ToggleButtonsStateManager manager =
-                              ToggleButtonsStateManager.of(innerContext);
+                      Builder(builder: (innerContext) {
+                        final ToggleButtonsStateManager manager =
+                            ToggleButtonsStateManager.of(innerContext);
 
-                          return ToggleButtons(
-                            borderRadius:
-                                const BorderRadius.all(Radius.circular(4.0)),
-                            isSelected: [
-                              manager.toggleButtonsState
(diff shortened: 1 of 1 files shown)


domesticmouse:
> Yeah they are all winding up the same state. I had a misunderstanding on how inherited widgets cause the Widgets below them to rebuild, so I wrapped individual widgets where I could. Though correct me if i'm wrong, after some reading I think the Widgets that have not changed won't be rebuilt.

I can honestly say I don't understand how the state is updated in this app. Which makes it a little convoluted to explain =)

Renzo-Olivares:
> > Yeah they are all winding up the same state. I had a misunderstanding on how inherited widgets cause the Widgets below them to rebuild, so I wrapped individual widgets where I could. Though correct me if i'm wrong, after some reading I think the Widgets that have not changed won't be rebuilt.
> 
> I can honestly say I don't understand how the state is updated in this app. Which makes it a little convoluted to explain =)

This is how I meant it to work.

The state and methods to mutate the state of the ToggleButtons is held within `ToggleButtonsStateManager` inherited widget.
The state and methods to mutate the state of the TextEditingDeltaHistory is held within `TextEditingDeltaHistoryManager` inherited widget.

The `ToggleButtonsState` is updated when the toggle buttons are pressed.
The `BasicTextInputClient` updates the `ToggleButtonsState` when the selection has changed. 

The `BasicTextInputClient` updates the `TextEditingDeltaHistory` when a delta comes in from the engine through `updateEditingValueWithDeltas` or a delta is created by the framework (through `Shortcuts/Actions`) and goes through `_userUpdateTextEditingValueWithDelta`. There is also another case when the `TextEditingValue` is set by `RenderEditable` during a gesture https://github.com/flutter/samples/blob/e3040c9e36fef6bf7facf935eb1be627f34b2246/simplistic_editor/lib/basic_text_field.dart#L118, which calls `UserUpdateTextEditingValue` and in there the `TextEditingDeltaHistory` is also updated.

This change makes sense to me though. Let me know if my explanation helps or if it needs more clarification.

Renzo-Olivares (approved):

domesticmouse:
> This is how I meant it to work.

Thanks for this write up, I'll have a closer look. The reason for my confusion is that the change methods housed in the managers are methods on the `_MyHomePageState` object that end in `setState((){})` calls, which is going to regenerate the whole tree.

I'll probably wind up with something a bit simpler, trading efficiency for explainability. This is the way when it comes to teaching materials =)

PR #1444, opened 2022-09-26
Add code sharing sample

craiglabenz:
Begins introducing a sample for sharing code between the client and server. Functionally complete, but marked as `[WIP]` for now as it needs thorough documentation.


code_sharing/Makefile
```diff
@@ -0,0 +1,22 @@
...
+run-server:
+	# Depends on `make build`
+	docker-compose up -d
+
+stop-server:
+	docker-compose stop
+
+# Flutter
+flutter-local:
+	cd client && flutter run -d macos
```
domesticmouse, line 11: I feel like we should set this up for linux as well. 

And then there is Windows...

code_sharing/client/test/widget_test.dart
```diff
@@ -0,0 +1 @@
+
```
domesticmouse, line 1: Not even a smoke test to confirm it compiles? Sad face

domesticmouse (approved):

code_sharing/client/test/widget_test.dart
craiglabenz: Oh right, that's fair!

code_sharing/Makefile
craiglabenz: The `Makefile` may actually just be a bad idea 🤔

tool/flutter_ci_script_stable.sh
```diff
@@ -15,6 +15,7 @@ declare -ar PROJECT_NAMES=(
     "add_to_app/prebuilt_module/flutter_module"
     "android_splash_screen"
     "animations"
+    "code_sharing"
```
craiglabenz, line 18: @domesticmouse - After this commit, I was expecting to see my new sample audited in the CI pass, yet I'm not seeing it. Do you know if I forgot to make a change when adding this?

domesticmouse: This probably should be `"code_sharing/client"` and `"code_sharing/server"`

domesticmouse (approved):

tool/flutter_ci_script_stable.sh
craiglabenz: omg, duh

tool/flutter_ci_script_beta.sh
```diff
@@ -17,6 +17,7 @@ declare -ar PROJECT_NAMES=(
     "animations"
     "code_sharing/client"
     "code_sharing/server"
+    "code_sharing/server/shared"
```
domesticmouse, line 20: Do you need to add it to `master` as well?

domesticmouse (approved):

tool/flutter_ci_script_beta.sh
domesticmouse: Probably want to pull all three into `stable` as well.

domesticmouse:
Triple approved!

PR #1463, opened 2022-10-06
Custom context menu example

justinmc:
This a sample meant to demonstrate custom context menus and the various ways they can be used.  This means the unified ability to customize the text selection toolbar in text fields and on selectable text and the right-click/long-press menu anywhere else in the app.

| Screenshot from 2022-10-06 16-44-23 | Screenshot from 2022-10-06 16-39-30 |
| --- | --- |

Must wait for https://github.com/flutter/flutter/pull/107193.


domesticmouse:
Hey @justinmc, I'm curious what versions of Flutter this sample should compile and run with. I'm happy with `master` only or `master` + `beta` in `experimental`, but I'd like an idea of when it is intended to migrate to top level with `stable` availability of the used APIs.

Secondly, this sample should be added to the CI pipelines. I'm happy to either point you in the right direction, or submit a PR to your repo for the required changes...

justinmc:
I've made this PR a draft for now because it requires https://github.com/flutter/flutter/pull/107193 to work, which hasn't been merged yet.  Once it's merged I'll make sure this works on master and open it for review.  Should I document this anywhere in the PR?

About CI, I'm happy to do the work to add it if you point me in the right direction.  Thank you for commenting and helping me with this!

domesticmouse:
Please add it to https://github.com/flutter/samples/blob/main/tool/flutter_ci_script_master.sh

justinmc:
@domesticmouse This PR is now ready for review.  The PR that it depended on in flutter/flutter has been merged, so it works on Flutter's master channel.

Are the CI failures anything that I should worry about?

tool/flutter_ci_script_master.sh
```diff
@@ -19,6 +19,7 @@ declare -ar PROJECT_NAMES=(
     "code_sharing/server"
     "desktop_photo_search/fluent_ui"
     "desktop_photo_search/material"
+    "experimental/context_menus"
```
domesticmouse, line 22: When are we expecting this to be usable from `beta` channel? I'd like to add it as a commented out entry to the beta and stable CI scripts with TODO comments as to when it should be enabled.

domesticmouse (approved):

domesticmouse:
From the CI runs, it looks like you stripped the `+x` permission from the master CI script. We need the CI passing before I can land this PR.

tool/flutter_ci_script_master.sh
justinmc: Will do.  It was merged 10 days ago (https://github.com/flutter/flutter/commit/0b451b6dfd6de73ff89d89081c33d0f971db1872).  I think it will make it into the next beta release?

justinmc:
Maybe my editor automatically stripped that `+x` permission somehow?  To fix it I just tried running this:

```
$ git update-index --chmod=+x tool/flutter_ci_script_master.sh   
```

And committed and pushed it.  Hopefully that fixes it.

justinmc:
Looks like there are still failures.  What were you seeing that made you think it was a file permissions problem, is that still showing up in the test output?  I can't tell what's wrong but I'm not familiar with this repo.

tool/flutter_ci_script_beta.sh
```diff
@@ -20,6 +20,9 @@ declare -ar PROJECT_NAMES=(
     "code_sharing/shared"
     "desktop_photo_search/fluent_ui"
     "desktop_photo_search/material"
+    // TODO(justinmc): To be enabled when its PR hits the master channel.
```
domesticmouse, line 23: // is not a comment in a bash script. Please change to #

domesticmouse (changes requested):

tool/flutter_ci_script_beta.sh
justinmc: Big fail, thanks.

domesticmouse:
Also, looking at https://github.com/flutter/samples/actions/runs/3412894076/jobs/5678974213#step:5:1 it looks like there are files in need of `dart format`

justinmc:
Current failure: I never even thought to add tests to this, so the default "flutter create" test is still there and failing.  I'll add some real tests and deleted that one.

domesticmouse (approved):

tool/flutter_ci_script_master.sh
domesticmouse: sgtm!

PR #1515, opened 2022-12-03
Fix go_router navigation

### LINKED ISSUE #1514: game_template navigation error
When `GoRouter.of(context).pop()` method is used on router to get back to previous screen it breaks navigation flow. 

https://github.com/flutter/samples/blob/56fa9f93e3d723f00e4b1c54f2be68859d7407f0/game_template/lib/src/settings/settings_screen.dart#L106

Steps to reproduce:

- Run Flutter Game Template app
- Goto Settings screen [Settings]
- Go back to Home screen [Back]
- Once again try to go to Settings screen [Settings] (this time nothing will happen) 

Consider to use `GoRouter.of(context).go('/')` instead `pop()` or use `push(...)` instead of `go(...)` method to jump to new screen.

filiph:
Mixing declarative (`.go()`) and imperative (`.pop()`) navigation no longer works.

This CL changes all navigation to the declarative `go()` wherever possible. The only imperative navigation still standing is the `/settings` route, which needs to be accessed (pushed) from both the main screen and the play session screen, and therefore there is no single URL to go back to. So we’re keeping `push()` / `pop()` there.

I’m sure it’s possible to implement even the `/settings` route with declarative navigation, but it seems way beyond the scope of this sample.

Fixes #1514.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/game_template/pubspec.yaml
+++ b/game_template/pubspec.yaml
@@ -15,7 +15,7 @@ dependencies:
 
   audioplayers: ^1.1.0
   cupertino_icons: ^1.0.2
-  go_router: ^5.0.5
+  go_router: ^5.2.1
   logging: ^1.1.0
   provider: ^6.0.2
   shared_preferences: ^2.0.13
--- a/game_template/lib/src/level_selection/level_selection_screen.dart
+++ b/game_template/lib/src/level_selection/level_selection_screen.dart
@@ -61,7 +61,7 @@ class LevelSelectionScreen extends StatelessWidget {
         ),
         rectangularMenuArea: ElevatedButton(
           onPressed: () {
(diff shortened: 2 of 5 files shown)


game_template/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
 
   audioplayers: ^1.1.0
   cupertino_icons: ^1.0.2
-  go_router: ^5.0.5
+  go_router: ^5.2.1
```
filiph, line 18: Raising dependency version lower bound because I'm not 100% sure that the new code works as expected with the previous versions.

domesticmouse (approved):

Issue #1510, opened 2022-11-30
[M3] Group components by taxonomy group 


guidezpl:
We should use the taxonomy groups listed at https://m3.material.io/components to group the components page on the m3 sample app. cc @QuncCccccc

domesticmouse:
@guidezpl I'm assigning this issue to you and @QuncCccccc, but feel free to push back if there should be a different owner for this issue.

QuncCccccc:
Thanks for letting me know. Will reorganize them:)

guidezpl:
Totally down to contribute. I haven't been able to figure out where the latest code is. Using the `master`/`main` branch, I get the stable layout. Which branch is the [latest work](https://flutter-experimental-m3-demo.web.app/#/) on?

domesticmouse:
It's not on a branch, it's in the `experimental` directory.

PR #1530, opened 2022-12-15
Update Material 3 Demo App

QuncCccccc:
This PR is to update the Material 3 Demo app in the experimental folder. The layout of the demo app is adaptive based on the screen size and more newly updated components are added. 

https://user-images.githubusercontent.com/36861262/207951232-be4b0b77-ba4f-41b0-ab91-bd0433bfff37.mov

Fix: 
#1510, 
[#112000](https://github.com/flutter/flutter/issues/112000), 
[#111999](https://github.com/flutter/flutter/issues/111999)

## Pre-launch Checklist

- [x] I signed the [CLA].


domesticmouse (approved):

experimental/material_3_demo/lib/component_screen.dart
```diff
@@ -4,67 +4,52 @@
...
-                        isExampleBar: true,
-                      )
-                    : Container(),
-              ],
-            ),
-          ),
-        ),
-      ),
+    return ListView(
+      shrinkWrap: true,
```
darrenaustin, line 22: This is probably fine for now, but we should really look into finding a way around using `shrinkWrap` here as it can lead to performance issues and we don't really want to encourage its use in an example. You might talk this over with @Piinks and see if she has any suggestions here.

darrenaustin (approved):
LGTM. Nice work. Looks so much more polished!

Just had a minor concern about using shrinkWrap below, but it is fine for now.

Couple of things I noticed that we might want to consider for future improvements:

- I think it might look better to use a grid for the entire expanded view rather than two independently scrolling lists. That way everything would scroll together and we could potentially let it expand to three columns if the window is wide enough.
- Do we need the gaps between each group block? It looks a little odd to me to have the white showing between the card colors.

These aren't critical, just something to think about going forward. 

It really looks nice. Thx!

domesticmouse:
@QuncCccccc feel free to land when you are happy. Thanks!

experimental/material_3_demo/lib/component_screen.dart
Piinks: Actually, I'd love to take a look at this first. We are trying to not recommend shrinkwrap unless absolutely necessary and have been removing it from sample code. It is very costly to performance and should only be used if there isn't an option to use slivers. I'll be able to take a look today and see if we can remove this.

Piinks (changes requested):
Just noting so we can look at this shrinkwrap issue

experimental/material_3_demo/lib/component_screen.dart
```diff
@@ -4,67 +4,52 @@
...
+  }
+}
+
+class SecondComponentList extends StatelessWidget {
+  const SecondComponentList({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return ListView(
+      shrinkWrap: true,
```
Piinks, line 40: It looks like it is being used here too. @QuncCccccc are these used inside of other lists, or are they being placed in Columns? These should probably be SliverLists instead

experimental/material_3_demo/lib/main.dart
```diff
@@ -176,40 +251,393 @@ class _Material3DemoState extends State<Material3Demo> {
...
+                onDestinationSelected: (index) {
+                  setState(() {
+                    screenIndex = index;
+                    handleScreenChanged(screenIndex);
+                  });
+                },
+                trailing: Expanded(
+                    child: Padding(
+                  padding: const EdgeInsets.only(bottom: 20),
+                  child: showLargeSizeLayout
```
Piinks, line 277: For readability, can you wrap these subtrees in methods that return a Widget?

As in

```dart
child: showLargeSizeLayout ? _largeLayout() : _smallLayout()
```

Piinks:
I checked this out and removed shrinkwrap, there were 5 cases of it and from what I can tell they weren't all necessary, so just removing it was an easy fix for some. :)

See changes in this commit: https://github.com/flutter/samples/pull/1534/commits/fb5084fdc8d90a2935c3f5681aa6cdbb4e040cf5

experimental/material_3_demo/lib/main.dart
QuncCccccc: Fixed:)

QuncCccccc:
Not sure why some tests in web_dashboard failed🙂, any suggestions? @domesticmouse 

Screenshot 2022-12-19 at 11 00 24 AM

domesticmouse:
> Not sure why some tests in web_dashboard failed🙂, any suggestions? @domesticmouse 
> 
> 
> 
> Screenshot 2022-12-19 at 11 00 24 AM
> 
> 

Web dashboard needs to be either updated or dropped from master CI. I'll get a PR together for this today.

QuncCccccc:
> > Not sure why some tests in web_dashboard failed🙂, any suggestions? @domesticmouse
> > Screenshot 2022-12-19 at 11 00 24 AM
> 
> Web dashboard needs to be either updated or dropped from master CI. I'll get a PR together for this today.

Got it. Thanks!

domesticmouse:
Please pull latest from `main`. That should hopefully clean up the CI failures

Piinks (approved):
LGTM! Thank you!

PR #1558, opened 2023-01-12
Update M3 demo app layout on small screen

QuncCccccc:
This PR is to update the layout of the M3 demo app when it is running on a narrow screen (phone devices).

On iPhone 8 (screen width: 375):

https://user-images.githubusercontent.com/36861262/211948833-3dbed76e-9e8d-4a45-9ae1-e42c97137782.mov

On a 16-inch laptop:

Screenshot 2023-01-11 at 4 47 40 PM

On a 32-inch monitor (future improvement: create a 3-column grid view so that more components can show up and the list can scroll together. And welcome for any other suggestions:) )

Screenshot 2023-01-11 at 4 49 22 PM

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/experimental/material_3_demo/lib/component_screen.dart
+++ b/experimental/material_3_demo/lib/component_screen.dart
@@ -281,32 +281,61 @@ class FloatingActionButtons extends StatelessWidget {
       label: 'Floating action buttons',
       tooltipMessage:
           'Floating action buttons include: \nFloatingActionButton.small, FloatingActionButton, FloatingActionButton.extended, and FloatingActionButton.large',
-      child: Wrap(
-        alignment: WrapAlignment.spaceEvenly,
-        crossAxisAlignment: WrapCrossAlignment.center,
-        children: [
-          FloatingActionButton.small(
-            onPressed: () {},
-            child: const Icon(Icons.add),
-          ),
-          rowDivider,
-          FloatingActionButton(
-            onPressed: () {},
-            child: const Icon(Icons.add),
-          ),
-          rowDivider,
-          FloatingActionButton.extended(
-            onPressed: () {},
-            icon: const Icon(Icons.add),
-            label: const Text('Create'),
-          ),
-          rowDivider,
-          FloatingActionButton.large(
-            onPressed: () {},
-            child: const Icon(Icons.add),
-          ),
-        ],
-      ),
+      child: LayoutBuilder(builder: (context, constraints) {
+        double screenWidth = MediaQuery.of(context).size.width;
+        if (screenWidth < 375.0) {
+          return Column(
+            children: [
+              Row(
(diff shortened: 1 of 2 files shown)


domesticmouse (approved):
LGTM if @darrenaustin agrees =)

experimental/material_3_demo/lib/component_screen.dart
```diff
@@ -281,32 +281,61 @@ class FloatingActionButtons extends StatelessWidget {
...
-          FloatingActionButton.large(
-            onPressed: () {},
-            child: const Icon(Icons.add),
-          ),
-        ],
-      ),
+      child: LayoutBuilder(builder: (context, constraints) {
+        double screenWidth = MediaQuery.of(context).size.width;
+        if (screenWidth < 375.0) {
+          return Column(
```
darrenaustin, line 287: This works, but seems like a fair amount of duplication. That said, I don't have a cleaner suggestion that wouldn't make it harder to understand. I wonder if instead of just having the extended on a row by itself, would it make sense to just have two rows of two with the small & extended and the medium & large? I am probably overthinking this. Just leave it as it is 😚.

experimental/material_3_demo/lib/component_screen.dart
```diff
@@ -455,9 +484,12 @@ class TextFields extends StatelessWidget {
                     ),
                   ),
                 ),
+                SizedBox(
+                  width: 10,
+                ),
```
darrenaustin, line 489: Should be on a single line and `const`.

experimental/material_3_demo/lib/component_screen.dart
```diff
@@ -509,9 +541,10 @@ class TextFields extends StatelessWidget {
                         ),
                       ),
                     ),
+                    SizedBox(width: 10),
```
darrenaustin, line 544:
```suggestion
                    const SizedBox(width: 10),
```

darrenaustin (approved):
LGTM. Looks really nice @QuncCccccc.

Just a couple of comments/formatting issues below. Thx.

experimental/material_3_demo/lib/component_screen.dart
QuncCccccc: Two rows of small & extended and medium & large sounds good to me. Just fixed:)

experimental/material_3_demo/lib/component_screen.dart
QuncCccccc: Fixed!

experimental/material_3_demo/lib/component_screen.dart
QuncCccccc: The `const` is added before the bracket for the `children` property, so I think we don't need to add const here:)

PR #1585, opened 2023-01-22
Fix and reenable context_menus email button test

parlough:
Closes #1584

PR #1588, opened 2023-01-23
Context Menu sample missing pumpAndSettle

justinmc:
I was getting some weird flaky behavior from these tests that used dragUntilVisible.  As fixed by @parlough in https://github.com/flutter/samples/pull/1585, this was because of a missing pumpAndSettle.  I've added them to other tests here to avoid problems in the future.


parlough (approved):
Looks good to me, thanks!

Issue #1616, opened 2023-01-31
`simplistic_editor` needs to implement `TextInputClient.insertContent`


domesticmouse:
Currently `simplistic_editor` is broken on `master` due to missing `TextInputClient.insertContent` implementation.

```console
error - lib/basic_text_input_client.dart:42:7 - Missing concrete implementation of 'TextInputClient.insertContent'. Try implementing the missing method, or make the class abstract. - non_abstract_class_inherits_abstract_member
```

PR #1618, opened 2023-02-01
Mixin `TextInputClient` to include `insertContent` implementation

parlough:
The new `insertContent` method is for rich content such as images from Android, so it doesn't seem like something a "simplistic" editor needs to handle for now. Including `TextInputClient` as a mixin allows us to incorporate its empty implementation of `insertContent`. This has the benefit of the sample working on both `stable` and `master`.

Fixes https://github.com/flutter/samples/issues/1616

### DIFF
--- a/simplistic_editor/lib/basic_text_input_client.dart
+++ b/simplistic_editor/lib/basic_text_input_client.dart
@@ -40,8 +40,7 @@ class BasicTextInputClient extends StatefulWidget {
 }
 
 class BasicTextInputClientState extends State<BasicTextInputClient>
-    with TextSelectionDelegate, TextInputClient
-    implements DeltaTextInputClient {
+    with TextSelectionDelegate, TextInputClient, DeltaTextInputClient {
   final GlobalKey _textKey = GlobalKey();
   late AppStateWidgetState manager;
   final ClipboardStatusNotifier? _clipboardStatus =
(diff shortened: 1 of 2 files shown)


domesticmouse:
PTAL @Renzo-Olivares

simplistic_editor/lib/basic_text_input_client.dart
```diff
@@ -40,7 +40,7 @@ class BasicTextInputClient extends StatefulWidget {
 }
 
 class BasicTextInputClientState extends State<BasicTextInputClient>
-    with TextSelectionDelegate
+    with TextSelectionDelegate, TextInputClient
```
Renzo-Olivares, line 43: I think we should also mixin the `DeltaTextInputClient` instead of implementing it. This  helps the sample in that every addition to `TextInputClient` will not break the sample. More context here https://github.com/flutter/flutter/issues/114380#issuecomment-1299137406 .

Renzo-Olivares (approved):
Other than my comment LGTM

simplistic_editor/lib/basic_text_input_client.dart
parlough: Makes sense to me, I'll make that change. Thanks!

domesticmouse (approved):

Issue #1685, opened 2023-02-28
Use `BrowserContextMenu` in the context menu sample


navaronbracke:
The `context_menu` sample at https://github.com/flutter/samples/tree/main/experimental/context_menus
still states that the samples do not work on web.

However, there is now a new API that enables using the `contextMenuBuilder` on the web, namely `BrowserContextMenu`.
We should update the samples to use that API and remove the note at the top of the sample.

We also had to update the examples in the SDK itself [here](https://github.com/flutter/flutter/issues/119184).

Side note: The context menu API's are not experimental anymore, since they are now available on stable.

domesticmouse:
Hey @justinmc, I'm sticking this on your TODO list =)

justinmc:
Thanks for pointing this out, on it!

PR #1687, opened 2023-02-28
Enable the context menu examples on web

justinmc:
Support for context menus on the web was added in https://github.com/flutter/flutter/issues/119184.  This PR enables support in these examples.

Closes https://github.com/flutter/samples/issues/1685

### DIFF
--- a/experimental/context_menus/README.md
+++ b/experimental/context_menus/README.md
@@ -10,9 +10,6 @@ such as the text selection toolbar on mobile or the right click menu on desktop.
 
 Just run `flutter run` in the same directory as this README file.
 
-Currently, most of the examples in this demo do not support web, because Flutter
-uses the browser's built-in context menu instead.
-
 ## The examples
 
 ### [Anywhere](https://github.com/flutter/samples/blob/main/experimental/context_menus/lib/anywhere_page.dart)
--- a/experimental/context_menus/lib/main.dart
+++ b/experimental/context_menus/lib/main.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 
 import 'anywhere_page.dart';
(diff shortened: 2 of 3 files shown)


justinmc:
@domesticmouse Should I add this demo to flutter.github.io/samples now?  If so, how do I do that?

domesticmouse:
There is a whole bunch of red on the CI front. How much of this can be fixed on stable?

justinmc:
It looks like it's complaining that BrowserContextMenu doesn't exist, which is true for stable.  Does CI here always use stable?

domesticmouse:
> It looks like it's complaining that BrowserContextMenu doesn't exist, which is true for stable. Does CI here always use stable?

Yup, the samples in the main directory are on stable, the samples in the `experimental` directory are on a mix of `beta` and `master`. 

I see two possible paths forward, first we can duplicate the sample in the main directory for `stable` channel, and a second version in `experimental` which is for `beta` channel. Second approach is to leave this PR active until the version of `stable` increments to include this new functionality.

Any preferences?

justinmc:
Thanks for the explanation.  I think I should leave this for now until it's available on `stable`, and worry about moving this into the main samples in a separate PR then.

Let me know if I should convert this to a draft or anything, otherwise I'll just write myself a reminder to come back to this after the next stable release.

domesticmouse:
I'm happy to take ownership of landing this PR. I'll construct an uber-PR with all the fixes and changes required for the next stable major version increment and I'll roll this PR into it. Thanks!

domesticmouse:
Justin, facepalm time. I just realised your sample *is* in the experimental directory. I'm going to remove the stable CI from this PR and land as is. We can worry about copying across to the top level once you are happy that the API has stabilised.

domesticmouse (approved):

justinmc:
Ah ok, I was wondering why I didn't ever have this problem before.  Thanks!

PR #1703, opened 2023-03-19
Add image-based color selection to M3 demo

esouthren:
Functionality added in https://github.com/flutter/flutter/pull/122919 allows developers to use an `ImageProvider` to set the color scheme of the theme, by using Material Color Utilities to extract the dominant colors of the image. 

This PR updates the M3 demo to add this functionality, adding 6 images to choose from. 

Video of the new demo:

https://user-images.githubusercontent.com/15033982/226174405-53846d6b-8401-45b2-b322-58d0a02a49db.mov

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/experimental/material_3_demo/lib/home.dart
+++ b/experimental/material_3_demo/lib/home.dart
@@ -165,7 +165,7 @@ class _HomeState extends State<Home> with SingleTickerProviderStateMixin {
     );
   }
 
-  Widget _expandedTrailingActions(ColorScheme colorScheme) {
+  Widget _expandedTrailingActions() {
     return Container(
       constraints: const BoxConstraints.tightFor(width: 250),
       padding: const EdgeInsets.symmetric(horizontal: 30),
@@ -201,7 +201,7 @@ class _HomeState extends State<Home> with SingleTickerProviderStateMixin {
             const Divider(),
             _expandedColorSeedAction(),
             const Divider(),
-            _expandedImageColorAction(Theme.of(context).colorScheme),
+            _expandedImageColorAction(),
           ],
         ),
       ),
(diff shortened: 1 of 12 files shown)


domesticmouse:
For changes that only work on `master` branch, please make the change to `experimental/material_3_demo` instead.

experimental/material_3_demo/lib/home.dart
```diff
@@ -146,15 +153,23 @@ class _HomeState extends State<Home> with SingleTickerProviderStateMixin {
...
+              )
             ]
           : [Container()],
     );
   }
 
-  Widget _expandedTrailingActions() => Container(
-        constraints: const BoxConstraints.tightFor(width: 250),
-        padding: const EdgeInsets.symmetric(horizontal: 30),
+  Widget _expandedTrailingActions() {
```
HansMuller, line 168: Functions like this could be private Widget classes instead. In many cases that means less work to do when this widget rebuilds; about the same otherwise.

experimental/material_3_demo/lib/main.dart
```diff
@@ -52,18 +55,34 @@ class _AppState extends State<App> {
...
     });
   }
 
+  void handleImageSelect(int value) {
+    setState(() {
+      colorSelectionMethod = ColorSelectionMethod.image;
+      imageSelected = ColorImageProvider.values[value];
+      ColorScheme.fromImageProvider(
+              provider: NetworkImage(ColorImageProvider.values[value].url))
+          .then((newScheme) => imageColorScheme = newScheme);
```
HansMuller, line 69: Should we defer changing the app's state until the image loads and the new color scheme has been computed?
```dart
void handleImageSelect(int value) {
  final String url = ColorImageProvider.values[value].url;
  ColorScheme.fromImageProvider(
    provider: NetworkImage(url).then((newScheme) {
      setState(() {
        colorSelectionMethod = ColorSelectionMethod.image;
        imageSelected = ColorImageProvider.values[value];
        imageColorScheme = newScheme;
      }),
    }),
  );
}
```

HansMuller (approved):
LGTM

QuncCccccc:
This looks great! Thanks a lot for helping update the demo! Looks like the tailing actions(switches, color seed grid and image grid) are just located below the Navigation rail labels now. Would it be better if we can put them on the bottom left? so that there's some space between labels and trailing actions.
Like: 
Screenshot 2023-03-21 at 3 19 47 PM

experimental/material_3_demo/lib/home.dart
esouthren: Thanks - changed.

experimental/material_3_demo/lib/main.dart
esouthren: Good shout, thanks

esouthren:
> This looks great! Thanks a lot for helping update the demo! Looks like the tailing actions(switches, color seed grid and image grid) are just located below the Navigation rail labels now. Would it be better if we can put them on the bottom left? so that there's some space between labels and trailing actions. Like: Screenshot 2023-03-21 at 3 19 47 PM

Thanks Qun, fixed - I've added a size-dependent scroll view so if the screen is wide but short, the user can still scroll through the trailing actions. But if there is screen space, they'll be anchored to the bottom.

QuncCccccc:
Deployed the web with the new change. Just to update and here's the web link: https://flutter-experimental-m3-demo.web.app/#/ :)

PR #1795, opened 2023-05-09
Update material_3_demo to Dart 3

johnpryan:


domesticmouse:
Have you patched both copies of `material_3_demo`? There is a copy in `experimental` that is being pasted over the copy in `/` in this PR.

domesticmouse:
> Is [experimental/material_3_demo](https://github.com/flutter/samples/tree/main/experimental/material_3_demo) the source of truth?

Yup. The intent is that `/experimental/material_3_demo` tracks `master` channel, and gets copied over `/material_3_demo` at each stable increment. This is to allow the Material team to show case new capabilities ahead of stable release.

johnpryan:
Do I need to update that one? Both?

domesticmouse:
> Do I need to update that one? Both?

Right now the best approach is to update `/experimental/material_3_demo` and then ping me. I'll deal with the copy.

PR #1802, opened 2023-05-10
Use SliverList and cache content height

### LINKED ISSUE #1694: Consider refactoring component lists into SliverLists
From investigating https://github.com/flutter/flutter/pull/121786

Each section of the component list is itself a long list that would benefit from being split into SliverLists. These sections are all very large and there are only a few of them. This has two effects:
- Way more content is being built than necessary since each large component section is only one child of the component ListView
- The scrollbar jumps around as the scroll metrics report very different metrics as these large chunks are being built on demand.

In the 2 list view, there are only 3 children in the lists on either side. Instead, each side should be a CustomScrollView, and each component section should be a SliverList. 

Screenshot 2023-03-08 at 6 22 48 PM

QuncCccccc:
Fixes #1694
This PR is to fix the shaking scrollbar problem. We replace the `ListView`s with `CustomScrollView` and cache the content height to provide a better estimation for the max scroll offset.

Same changes are applied to both Material 3 demo in the root directory and the M3 demo in the `/experimental/`

https://github.com/flutter/samples/assets/36861262/24041a84-64e7-480e-8ea3-0d0811038818

https://github.com/flutter/samples/assets/36861262/9f274109-f698-4522-8316-7c75b2596c8b

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/experimental/material_3_demo/lib/component_screen.dart
+++ b/experimental/material_3_demo/lib/component_screen.dart
@@ -49,19 +49,22 @@ class FirstComponentList extends StatelessWidget {
       child: CustomScrollView(
         slivers: [
           SliverPadding(
-              padding: showSecondList
-                  ? const EdgeInsetsDirectional.only(end: smallSpacing)
-                  : EdgeInsets.zero,
-              sliver: SliverList(
-                  delegate: BuildSlivers(
-                      heights: heights,
-                      builder: (context, index) {
-                        return _CacheHeight(
-                          heights: heights,
-                          index: index,
-                          child: children[index],
-                        );
-                      }))),
+            padding: showSecondList
+                ? const EdgeInsetsDirectional.only(end: smallSpacing)
+                : EdgeInsets.zero,
+            sliver: SliverList(
+              delegate: BuildSlivers(
+                heights: heights,
+                builder: (context, index) {
+                  return _CacheHeight(
+                    heights: heights,
+                    index: index,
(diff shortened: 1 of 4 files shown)


domesticmouse:
You'll need to pull from `main` to have a chance at getting a (mostly) green CI.

experimental/material_3_demo/lib/component_screen.dart
```diff
@@ -62,22 +81,123 @@ class SecondComponentList extends StatelessWidget {
...
         ],
       ),
     );
   }
 }
 
+// Based on the fact that the list content is fixed, we cache
+// the height of the content during the first-time scrolling.
+// The cached height can be used to override
+// `SliverChildDelegate.estimateMaxScrollOffset`, to avoid a shaking scrollbar.
```
HansMuller, line 121:
```suggestion
// If the content of a CustomScrollView does not change, then it's
// safe to cache the heights of each item as they are laid out. The
// sum of the cached heights are returned by an override of
// `SliverChildDelegate.estimateMaxScrollOffset`. The default version
// of this method bases its estimate on the average height of the
// visible items. The override ensures that the scrollbar thumb's
// size, which depends on the max scroll offset, will shrink smoothly
// as the contents of the list are exposed for the first time, and
// then remain fixed.
```

HansMuller (approved):
LGTM

PR #1815, opened 2023-05-12
Update ng-flutter dependencies for Angular 16.0.1

parlough:
Tested by building and running full app locally.


ditman:
You shouldn't have to do any more of these, if we merge the PR that it's pending ;)

domesticmouse (approved):

PR #1857, opened 2023-05-31
`game_template`: Make `_handleAppLifecycle` switch exahustive

domesticmouse:
## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/game_template/lib/src/audio/audio_controller.dart
+++ b/game_template/lib/src/audio/audio_controller.dart
@@ -164,7 +164,8 @@ class AudioController {
         // No need to react to this state change.
         break;
       default:
-        _log.warning('Unhandled AppLifecycleState: ${_lifecycleNotifier!.value}');
+        _log.warning(
+            'Unhandled AppLifecycleState: ${_lifecycleNotifier!.value}');
     }
   }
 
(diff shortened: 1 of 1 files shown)


domesticmouse:
> Wait, I don't understand. https://api.flutter.dev/flutter/dart-ui/AppLifecycleState.html is already exhaustive in the switch statement. There are 4 possible values.
> 
> I always thought that the cool part about not having a `default` case is that if `AppLifecycleState` ever gains another value, the code will immediately get a compile-time error. By adding a `default`, we might miss an important breaking change. Is there a new lint that forces `default`? If so, I'd argue it's ill-advised.

Wanna explain why I'm seeing https://github.com/flutter/samples/actions/runs/5129921533/jobs/9228068909?pr=1856 ?

domesticmouse:
I'm guessing `AppLifecycleState` gained a `hidden` state, but only on master. So, I supposed the correct path forward is to progressively elide `game_template` from each level of CI until the new state hits master and we can add the required code for `hidden` and turn all the CI checks back on. 

Which leads to the fun question, what is the correct code for `AppLifecycleState` of `hidden`? =)

PR #1909, opened 2023-06-23
adding more named and explicit imports

miquelbeltran:
Following the proposal in https://github.com/flutter/website/issues/7798 I have added more explicit and named imports on those places I thought it would make sense.

Let me know what you think.

## Pre-launch Checklist

- [ ] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [ ] I signed the [CLA].
- [ ] I read the [Contributors Guide].

### DIFF
--- a/animations/lib/src/misc/carousel.dart
+++ b/animations/lib/src/misc/carousel.dart
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:ui';
+import 'dart:ui' show PointerDeviceKind;
 import 'package:flutter/material.dart';
 
 class CarouselDemo extends StatelessWidget {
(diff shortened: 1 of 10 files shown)


domesticmouse:
I'm wondering if we should just adopt the default of `import 'dart:ui' as ui;` as opposed to having a mix of `as ui` and `show Name`

domesticmouse:
Also, please update this PR to turn the CI green. =)

miquelbeltran:
> I'm wondering if we should just adopt the default of `import 'dart:ui' as ui;` as opposed to having a mix of `as ui` and `show Name`

makes sense, I will do the change to see how it looks like

domesticmouse (approved):
Is `import 'dart:math' as math;` something we do elsewhere? It almost feels like overkill?

miquelbeltran:
> Is `import 'dart:math' as math;` something we do elsewhere? It almost feels like overkill?

As with the other changes we have done related to https://github.com/flutter/website/issues/7798 none of them are strictly necessary.

I like the idea of having the math dependency as a named import because it allows for easier discoverability, and because the math import has a lot of top level functions and constants, so it doesn't mess with autocomplete. Although, effective Dart doesn't have a recommendation regarding this, and I don't have a strong opinion, so I'm fine to revert the change.

PR #1945, opened 2023-07-18
add React embedding example

p-mazhnik:
Part of https://github.com/flutter/flutter/issues/128980

Todo list: 
- [ ] Currently `web_embedding/cra-flutter/flutter` is copied from `web_embedding/ng-flutter/flutter` without any changes. Maybe it will be better to move `web_embedding/ng-flutter/flutter` directory to the `web_embedding` and share between different examples?
- [ ] remove generated unused images (react logo)
- ~~add js tests~~
- [ ] add details to the `web_embedding/cra-flutter/README.md`
- [ ] display spinner when Flutter is not ready

## Pre-launch Checklist

- [x] I signed the [CLA].


ditman:
> Maybe it will be better to move web_embedding/ng-flutter/flutter directory to the web_embedding and share between different examples?

This is a good idea, especially to unify the code we use in the Flutter side to communicate into JS (`@JSExports`, signaling with a custom event from the root of the app, etc...) however I'm concerned about having flutter *outside* of the Angular/ReactJS project directory; will the build system copy code from the outside? Will it make setup more weird than what it is right now?

For this examples, I'd rather have a duplicated Flutter app than a weird build setup for the app (IMO).

> add JS tests

I held off on these. If there's interest, we can extract the relevant bits (the Flutter component or whatever) to their own NPM package so it's reusable, and we can add testing to those.

> display spinner when Flutter is not ready

In the angular implementation, I just rendered a bunch of angular code inside of the tree to render the spinner when flutter is not loaded, and then the flutter app, I'd expect this to be somewhat "easy"?

web_embedding/cra-flutter/flutter/README.md
```diff
@@ -0,0 +1,16 @@
+# ng_companion
```
ditman, line 1: //TODO update this

web_embedding/cra-flutter/src/App/App.tsx
```diff
@@ -0,0 +1,217 @@
...
+            edge="start"
+            color="inherit"
+            aria-label="menu"
+            onClick={handleDrawer}
+            sx={{ mr: 2 }}
+          >
+            <MenuIcon/>
+          </IconButton>
+          <Typography variant="h6" component="div" sx={{ flexGrow: 1 }}>
+            React 🤝 Flutter
```
ditman, line 134: 200% love this haha, thanks for keeping the handshake! :P

web_embedding/cra-flutter/src/App/FlutterView/FlutterView.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const initFlutterApp = async () => {
+      if (!engineInitializerPromise) {
+        console.log('create Flutter engine initializer promise...')
+        engineInitializerPromise = new Promise<any>((resolve) => {
+          console.log('setup Flutter engine initializer...')
+          _flutter.loader.loadEntrypoint({
+            entrypointUrl: src,
+            onEntrypointLoaded: async (engineInitializer: any) => {
+              resolve(engineInitializer)
+            }
```
ditman, line 35: Maybe you can simplify this as:

```suggestion
            onEntrypointLoaded: resolve
```

ditman: (Also, why do you need to "leak" the engineInitializer out of this callback, you could potentially just have a `Promise<bool>` that resolves to 'true' once the whole async initialization procedure is done, right?)

ditman (commented):
This is great stuff, thanks for the contribution!

web_embedding/cra-flutter/src/App/App.tsx
```diff
@@ -0,0 +1,217 @@
...
+  const flutterState = useRef<any>(null)
+  const handleDrawer = () => {
+    setDrawerOpened(!drawerOpened)
+  }
+  const onFlutterAppLoaded = (state: any) => {
+    flutterState.current = state
+    state.onClicksChanged(() => { setClicks(flutterState.current?.getClicks()) })
+    state.onTextChanged(() => { setText(flutterState.current?.getText()) })
+    state.onScreenChanged(() => { setScreen(flutterState.current?.getScreen()) })
+  }
```
ditman, line 96: @p-mazhnik from a ReactJS standpoint, how weird is this interaction? Is there anything we could change in the state object coming from flutter so this is more natural for a ReactJS site?

web_embedding/cra-flutter/src/App/FlutterView/FlutterView.tsx
p-mazhnik: Component can be mounted and unmounted multiple times, but we can't call `loadEntrypoint` function multiple times as well -- it will result in an error. So we need a way to call `engineInitializer.initializeEngine` multiple times without re-creating `engineInitializer`. This means that the `engineInitializer` must be shared between different component re-renders

ditman: Ohh, interesting, that might be a bug in the Angular version then, I don't remember attempting to preserve any of that :)

p-mazhnik: It only happens if `loadEntrypoint` function is called a second time before the completion of the first call. So it is hard to detect. In this sample, special [StrictMode](https://react.dev/reference/react/StrictMode) component is used (exists by default for the newly created project), that is designed to find bugs in the components. For instance, it specifically re-renders components to find accidentally impure code.

web_embedding/cra-flutter/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "cra-flutter",
+  "version": "0.0.0",
+  "scripts": {
+    "prestart": "rimraf node_modules/.cache/",
+    "start": "react-scripts start",
+    "prebuild": "pushd flutter && flutter clean && flutter build web --output=../public/flutter && popd",
```
p-mazhnik, line 7: > > Maybe it will be better to move web_embedding/ng-flutter/flutter directory to the web_embedding and share between different examples?

> This is a good idea, especially to unify the code we use in the Flutter side to communicate into JS (@JSExports, signaling with a custom event from the root of the app, etc...) however I'm concerned about having flutter outside of the Angular/ReactJS project directory; will the build system copy code from the outside? Will it make setup more weird than what it is right now?

If the `flutter` directory will be in the parent directory, this line will become 
```bash
pushd ../flutter && flutter clean && flutter build web --output=../cra-flutter/public/flutter && popd
```

And similar for the Angular app, we can have smth like this:

instead of 
```bash
pushd flutter && flutter clean && flutter build web && popd
```
we will have
```bash
pushd ../flutter && flutter clean && flutter build web --ouput=../ng-flutter/flutter && popd
```

These should be the only changes, besides documentation updates

ditman: Yes, as long as we can put the build output of flutter within the reach of the "resources"/"assets" of the angular/react web apps, it might be interesting to share (but I think this is not super high priority, because it's rare that people looking for an example in a framework will look at others and say: hmmm, you should have de-duplicated this!, I think it's more like: OK, here's where my flutter app goes!)

domesticmouse:
I'd prefer not to have a plethora of samples embedding Flutter into N JavaScript libraries and frameworks.

ditman:
> 
I'd prefer not to have a plethora of samples embedding Flutter into N JavaScript libraries and frameworks.

@domesticmouse fair, do you have any idea of where to store additional embedding examples? People seem to want one-per-existing-JS-framework (so potentially: infinite examples!)

Should we extend the root [README of the `web_embedding` samples](https://github.com/flutter/samples/blob/main/web_embedding/README.md) with links to community contributions, and just add a link to wherever @p-mazhnik's wants to have his React example published?

domesticmouse:
Thoughts @RedBrogdon?

web_embedding/cra-flutter/src/App/FlutterView/FlutterView.tsx
p-mazhnik: I have a related question here. If we call `initializeEngine` and `runApp` multiple times (like for the different re-renders), is it expected that `registerWith` is called multiple times for the web flutter plugins?

web_embedding/cra-flutter/src/App/App.tsx
p-mazhnik: I would say that a more natural flow is when we receive an updated value from the subscription,
so here we can have 
```js
state.onScreenChanged((screen) => { setScreen(screen) })
```

web_embedding/cra-flutter/src/App/FlutterView/FlutterView.tsx
ditman: @p-mazhnik I don't think we support very well multiple calls to `initializeEngine` and `runApp` right now, so the answer to your question is "undefined". However, according to the bootstrap of a Flutter Web app, `registerWith` gets called right before `runApp`:

* https://github.com/flutter/engine/blob/56b32827616b247ebf178ee33c08c3bb4022b141/lib/web_ui/lib/ui_web/src/ui_web/initialization.dart#L37-L43

web_embedding/cra-flutter/src/App/App.tsx
ditman: Makes sense, this should be a simple change in the `class DemoAppStateManager`. It could also expose the values as proper `getters`/`setters` rather than named `get/setSomething` methods, so `state.text` is a thing that can be read from and written to. (I wonder if `@JSExport` supports that!)

domesticmouse:
> Should we extend the root [README of the `web_embedding` samples](https://github.com/flutter/samples/blob/main/web_embedding/README.md) with links to community contributions, and just add a link to wherever @p-mazhnik's wants to have his React example published?

I think this is a good solution.

web_embedding/cra-flutter/src/App/FlutterView/FlutterView.tsx
p-mazhnik: Turns out we can't have globally stored `engineInitializerPromise` :)
Just found that if we use previous `engineInitializer`, component _sometimes_ is not rendered in the element after remounting, even though code is loaded and interoperability works.

But looks like the issue with the multiple calls to `loadEntrypoint` is gone after I refactored code from
```js
_flutter.loader.loadEntrypoint({
   entrypointUrl: src,
   onEntrypointLoaded: async (engineInitializer: any) => {
    let appRunner = await engineInitializer.initializeEngine({
      hostElement: ref.current,
      assetBase: assetBase,
    });
    await appRunner.runApp();
  }
});
```
to
```js
const initFlutterApp = async () => {
  const engineInitializer = await new Promise<any>((resolve) => {
   _flutter.loader.loadEntrypoint({
      entrypointUrl: src,
      onEntrypointLoaded: resolve,
    })
  })

  const appRunner = await engineInitializer?.initializeEngine({
    hostElement: ref.current,
    assetBase: assetBase,
  })
  await appRunner?.runApp()
}
initFlutterApp()
```

As a small downside, our html page will have script tag injected on each render, but at least everything is working.

So I am going to remove global variable

p-mazhnik: Should I create issues in the Flutter repo for my findings? I can probably provide some steps to reproduce.

However, I understand this is all new and not fully supported, so I am unsure about the actions that should be taken here and what is actually expected. All of it is related to the multiple calls of some function, which, as you said, are not well supported.
Anyway, I'll just leave them here, so I won't forget anything:

- If the standard Flutter initialization script is called two times 'too fast' (probably related to the `engineInitializer.initializeEngine` call and not to the `loadEntrypoint` as I thought before): 
`Uncaught TypeError: Cannot set properties of null (setting 'exports') at VM23285 canvaskit.js:219:18`
- If the standard Flutter initialization script is called two times 'too fast' (this one is probably related to the `appRunner.runApp` call and not to the `loadEntrypoint` as I thought before): 
`Uncaught TypeError: Cannot read properties of null (reading 'toString') at Object.HostNode_HostNode$create ... at Object.initializeEngineUi`
- if `engineInitializer` is reused, sometimes app is not rendered in the `hostElement`, but it looks like engine is still initialized (or the previous one is not cleaned up properly)
- nit: multiple calls to the `loadEntrypoint` inject multiple scripts on the page.
- (probably not related to Flutter itself, but still) `registerWith` called multiple times for the web plugins, causing different issues with the community plugins, because most of them currently assume that it will be called once.

ditman:
@p-mazhnik do you mind keeping the React example in your github.com/p-mazhnik (or wherever you want to host it), then we can send a new PR to flutter/samples sprucing up the README, and linking to wherever your code ends up being hosted?

Thanks!

ditman:
(We're continuing to work in @p-mazhnik's repo, and will send a PR to flutter/samples whenever the sample lands there. Thanks @p-mazhnik!!)

PR #1437, opened 2022-09-20
Update `navigation_and_routing` sample to `go_router` 5

johnpryan:
Updates the navigation_and_routing sample to go_router 5. 

Depends on https://github.com/flutter/flutter/issues/111981

## Pre-launch Checklist

- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I read the [Contributors Guide].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] All existing and new tests are passing.

PR #2067, opened 2023-10-21
Update navigation_and_routing sample to go_router

johnpryan:
This is a new PR to update the navigation_and_routing sample to the latest version of go_router. It is a based on https://github.com/flutter/samples/pull/1437


navigation_and_routing/pubspec.yaml
```diff
@@ -22,11 +23,10 @@ dependencies:
       path: plugins/window_size
 
 dev_dependencies:
-  analysis_defaults:
-    path: ../analysis_defaults
+  flutter_lints: ^2.0.1
```
domesticmouse, line 26: Please revert this change. We have a default set of lints for all samples.

navigation_and_routing/pubspec.yaml
```diff
@@ -22,11 +23,10 @@ dependencies:
...
   flutter_test:
     sdk: flutter
-  test: ^1.16.0
+  test: ^1.24.0
 
 flutter:
-  uses-material-design: true
+  uses-material-design: true
```
domesticmouse, line 32: Add NL at EOF

navigation_and_routing/lib/src/app.dart
```diff
@@ -2,11 +2,23 @@
...
+import 'package:bookstore/src/screens/books.dart';
+import 'package:bookstore/src/screens/sign_in.dart';
+import 'package:bookstore/src/widgets/fade_transition_page.dart';
 import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
 
-import 'auth.dart';
-import 'routing.dart';
-import 'screens/navigator.dart';
+import 'screens/authors.dart';
```
parlough, line 15: A mix of relative and absolute imports are being used in this file. Consider standardizing to relative :)

johnpryan:
@domesticmouse @parlough PTAL

domesticmouse:
Looks like the CI is cranky:

```console
Analyzing navigation_and_routing...

warning - lib/src/screens/authors.dart:11:7 - This class (or a class that this class inherits from) is marked as '@immutable', but one or more of its instance fields aren't final: AuthorsScreen.onTap - must_be_immutable
warning - lib/src/screens/books.dart:6:8 - Unused import: 'package:go_router/go_router.dart'. Try removing the import directive. - unused_import
   info - lib/src/app.dart:61:17 - The 'child' argument should be last in widget constructor invocations. Try moving the argument to the end of the argument list. - sort_child_properties_last

3 issues found.
```

navigation_and_routing/lib/src/app.dart
```diff
@@ -16,98 +28,245 @@ class Bookstore extends StatefulWidget {
...
+                },
+              );
+            },
+            routes: [
+              ShellRoute(
+                pageBuilder: (context, state, child) {
+                  return FadeTransitionPage<dynamic>(
+                    key: state.pageKey,
+                    // Use a builder to get the correct BuildContext
+                    // TODO (johnpryan): remove when https://github.com/flutter/flutter/issues/108177 lands
```
domesticmouse, line 76: Consider raising an issue in `flutter/samples` issue tracker to keep this task visible.

domesticmouse (approved):
LGTM once CI passes

navigation_and_routing/lib/src/screens/authors.dart
```diff
@@ -4,25 +4,27 @@
...
+  const AuthorsScreen({
+    required this.onTap,
+    super.key,
+  });
 
   @override
   Widget build(BuildContext context) => Scaffold(
         appBar: AppBar(
-          title: Text(title),
+          title: const Text(title),
```
parlough: I'm not sure why the analyzer isn't complaining here on newer SDK builds, but this can't be const.

```suggestion
          // ignore: prefer_const_constructors
          title: Text(title),
```

I'll look for an existing SDK issue or open one.

parlough (approved):
Looks good to me after one fix.

Please do consider updating the README for the new state of the sample though. Or stripping out the outdated parts :)

navigation_and_routing/lib/src/screens/authors.dart
parlough: Ah you fixed it already, thanks!

johnpryan: The `title` field is not const, so this makes sense. I made it a `final` field and assigned the value in the constructor so that the `Text` widget doesn't need to be const. Not the most sensible fix, but the easiest for the different versions of the analyzer to understand 🤷

parlough: Sounds good to me. Could also make it `static const` if you don't want the title to be changed.

I filed https://github.com/dart-lang/sdk/issues/53927

navigation_and_routing/lib/src/app.dart
johnpryan: https://github.com/flutter/samples/issues/2074

PR #2081, opened 2023-11-08
Update ng-flutter embedding example to Angular v17

parlough:
This should also fix the dependabot security alert currently open against the repository (https://github.com/flutter/samples/security/dependabot/8).


johnpryan (approved):
LGTM, thanks!

PR #2098, opened 2023-11-21
Use `Card.outlined` instead of using hard-coded values for outlined card

QuncCccccc:
We created the factory method [`Card.outlined`](https://github.com/flutter/flutter/pull/136229) a couple weeks ago, so this PR is just to update the outlined card to use the factory method instead of using hard-coded values. No appearance change.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/material_3_demo/lib/component_screen.dart
+++ b/material_3_demo/lib/component_screen.dart
@@ -2466,14 +2466,7 @@ class _ComponentDecorationState extends State<ComponentDecoration> {
                     focusNode.requestFocus();
                   },
                   behavior: HitTestBehavior.opaque,
-                  child: Card(
-                    elevation: 0,
-                    shape: RoundedRectangleBorder(
-                      side: BorderSide(
-                        color: Theme.of(context).colorScheme.outlineVariant,
-                      ),
-                      borderRadius: const BorderRadius.all(Radius.circular(12)),
-                    ),
+                  child: Card.outlined(
                     child: Padding(
                       padding: const EdgeInsets.symmetric(
                           horizontal: 5.0, vertical: 20.0),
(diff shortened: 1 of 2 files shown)


material_3_demo/lib/component_screen.dart
```diff
@@ -2466,14 +2466,7 @@ class _ComponentDecorationState extends State<ComponentDecoration> {
...
                   behavior: HitTestBehavior.opaque,
-                  child: Card(
-                    elevation: 0,
-                    shape: RoundedRectangleBorder(
-                      side: BorderSide(
-                        color: Theme.of(context).colorScheme.outlineVariant,
-                      ),
-                      borderRadius: const BorderRadius.all(Radius.circular(12)),
-                    ),
+                  child: Card.outlined(
```
domesticmouse, line 2469: Looks like `Card.outlined` isn't in stable yet, so this change isn't due to land yet.

One this file is reverted, you will need to alter https://github.com/flutter/samples/blob/main/tool/flutter_ci_script_stable.sh to comment out `"experimental/material_3_demo"` with a TODO to re-introduce it once stable version increments.

QuncCccccc: Oh I see. I will close this PR for now. Maybe we can land it in next stable release. There isn't any appearance change anyway:)

domesticmouse: For future reference you are more than free to land `master` only changes in the `experimental` directory. I just have it currently included in all of the CI's because it passes. I'm happy to comment `experimental` apps out of `stable` and `beta` CI as required to make the tree green.

Happy thanks giving! =)

QuncCccccc: Thanks for the info! And happy thanksgiving too!

PR #2104, opened 2023-12-05
Simplify experimental pedometer example

liamappelbe:
Since ffigen added support for [`NativeCallable.listener`](https://api.flutter.dev/flutter/dart-ffi/NativeCallable/NativeCallable.listener.html) to its ObjC bindings, this example can be simplified. We can replace the `Dart_Port` logic with `ObjCBlock.listener`, which lets us get rid of most of the native code.

We still need a small bit of native code to `retain` a reference to the callback's arguments before invoking the listener, otherwise the arguments may be ref counted and deleted before the Dart side of the callback is invoked. See https://github.com/dart-lang/native/issues/835


liamappelbe:
Nevermind, I think this is blocked on another small tweak to the ObjC bindings.

liamappelbe:
Actually I managed to fix the bug without ffigen changes. Ready for review.

domesticmouse:
Looks like there are CI issues:

```console
warning - example/lib/steps_repo.dart:5:8 - Unused import: 'dart:isolate'. Try removing the import directive. - unused_import
warning - example/lib/steps_repo.dart:102:49 - The receiver can't be null, so the null-aware operator '?.' is unnecessary. Try replacing the operator '?.' with '.'. - invalid_null_aware_operator
warning - example/lib/steps_repo.dart:104:61 - The '!' will have no effect because the receiver can't be null. Try removing the '!' operator. - unnecessary_non_null_assertion
```

liamappelbe:
> Looks like there are CI issues:

Fixed

domesticmouse:
PTAL @leighajarett

domesticmouse (approved):

PR #2147, opened 2024-01-24
Update Angular embedding sample

parlough:
- Replace `package:js` usage with `dart:js_interop`
- Remove and ignore `package-lock.json` since it keeps causing security warnings: https://github.com/flutter/samples/security/dependabot. It's not necessary for this sample and the constraints already present in the `package.json`.
- Update to Angular 17.1


domesticmouse (approved):

kevmoo:
Bless you, @parlough! 🙏

PR #2153, opened 2024-01-30
Remove experimental/material_3_demo

johnpryan:
There are two versions of material_3_demo, this removes the one that lives in `experimental/`

cc: @QuncCccccc


kevmoo:
Are there two so we can keep track of the prerelease features that haven't landed in stable yet?

QuncCccccc:
> Are there two so we can keep track of the prerelease features that haven't landed in stable yet?

Yes, I hosted the experimental one myself so I can test some master changes. Right now the experimental demo deployed the new ColorScheme change(https://github.com/flutter/flutter/pull/138521); all components's colors are slightly different, compared with the demo in the root:)

QuncCccccc:
Can we still keep this one but maybe we can provide a different name to distinguish? Sorry for the confusion🥲

domesticmouse:
The intent of the two versions was that the experimental version is allowed to compile on only `master` channel while the version in the main directory must compile on `stable` channel.

The reason they have the same name is to make it effortless to upgrade the stable version when the features on each new flutter stable release.

johnpryan:
> The reason they have the same name is to make it effortless to upgrade the stable version when the features on each new flutter stable release.

Just a suggestion, but you could use the [beta branch](https://github.com/flutter/samples/tree/beta) instead to avoid confusion.

kevmoo:
A BRANCH makes a LOT more sense here.

Like we're projecting the need for this code into the wrong dimension

parlough:
If we keep using and expanding this demo, it likely should be in its own repository anywhere, where a branch would be easier.

johnpryan:
I filed an issue with some more thoughts: https://github.com/flutter/samples/issues/2154

QuncCccccc (approved):
This makes sense. LGTM:) Sorry again if this caused any confusion.

johnpryan:
@QuncCccccc Not a problem at all! I'm glad we're on the same page.

bleroux:
> Just a suggestion, but you could use the [beta branch](https://github.com/flutter/samples/tree/beta) instead to avoid confusion.

@johnpryan  Is the flutter/samples `beta` branch supposed to compile on flutter/flutter `master` branch?
From its name one can assume it will compile on flutter/flutter `beta` branch.
Maybe a flutter/sample `alpha` branch is needed?
Wether it is 'beta' or 'alpha' (or something else) some changes on the CI side are probably needed.

johnpryan:
> Is the flutter/samples beta branch supposed to compile on flutter/flutter master branch?

No, it's for the `beta` branch as you would expect.

> Maybe a flutter/sample alpha branch is needed?

Possibly, but the the Flutter `beta` branch is updated regularly too.

> some changes on the CI side are probably needed

Right now we run CI on all samples using the Flutter SDK's `main`, `beta`, and `stable` branches. What change to CI do you have in mind?

bleroux:
I don't have a good knowledge of flutter/samples (I filed one PR some months ago), so let me be more practical:

On Monday, I saw that the `material_3_demo` sample is broken when run on flutter/flutter master branch.
So based on https://github.com/flutter/samples/tree/main/experimental documentation, I started to investigate for a fix by updating the code in experimental/material_3_demo. Now that this directory is gone, I'm not sure if I should work inside the non-experimental `material_3_demo` directory because it is possible that my fix will only be valid on flutter/flutter master channel.

> Right now we run CI on all samples using the Flutter SDK's main, beta, and stable branches. What change to CI do you have in mind?

A change that make it possible to run a samples PR only on flutter/flutter master branch.

PR #2157, opened 2024-02-07
Remove Expanded from NavigationRail

johnpryan:
https://github.com/flutter/flutter/pull/137415 changed NavigationRail to use a SingleChildScrollView, so this sample can't use `Expanded` for the `trailing` widget anymore.

cc: @goderbauer @QuncCccccc 

Fixes flutter/flutter#143061

### DIFF
--- a/material_3_demo/lib/home.dart
+++ b/material_3_demo/lib/home.dart
@@ -214,24 +214,22 @@ class _HomeState extends State<Home> with SingleTickerProviderStateMixin {
                 handleScreenChanged(screenIndex);
               });
             },
-            trailing: Expanded(
-              child: Padding(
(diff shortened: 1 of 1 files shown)


kevmoo (approved):

johnpryan:
Looks like I broke some tests, investigating...

johnpryan:
This is turning out to be more difficult than I was expecting. The sidebar seems to be too large in the medium-width case:

Screenshot 2024-02-07 at 10 42 46 AM

johnpryan:
Reverting https://github.com/flutter/flutter/pull/137415 and adding this sample to our framework tests sounds good to me (although we might want to think about how to do that in a maintainable way)

goderbauer:
I'll get the revert going...

PR #2236, opened 2024-04-08
Clean up ng-flutter a bit

parlough:
No functional change, just while I was verifying it still works with the latest Angular releases, completed some cleanup as well.


domesticmouse (approved):
Please elide from beta and master CI with an appropriate TODO description

PR #2265, opened 2024-04-30
Removes outdated 'window_size' package from navigation_and_routing app

ericwindmill:
That package was breaking the app, and 'pub get'. And according to the [plugin's Github](https://github.com/google/flutter-desktop-embedding), it's no longer necessary.

### DIFF
--- a/navigation_and_routing/lib/main.dart
+++ b/navigation_and_routing/lib/main.dart
@@ -2,12 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:io' show Platform;
-
-import 'package:flutter/foundation.dart' show kIsWeb;
 import 'package:flutter/material.dart';
 import 'package:url_strategy/url_strategy.dart';
-import 'package:window_size/window_size.dart';
 
 import 'src/app.dart';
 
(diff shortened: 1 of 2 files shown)


ericwindmill:
@johnpryan, @domesticmouse 

This package, and the `setupWindow` pattern are used in multiple samples, and I just want to get a sanity check that it's okay to remove before I remove it from the rest of the samples.

domesticmouse:
I'm happy to remove if it is breaking things. Your reading of the repo description needs some help tho, this package isn't deprecated.

domesticmouse:
I suspect this is related to https://github.com/flutter/flutter/issues/144898

domesticmouse:
> And according to the [plugin's Github](https://github.com/google/flutter-desktop-embedding), it's no longer necessary.

This is an incorrect reading of the linked page. The `window_size` package is not outdated per se. The previous contents of the linked repo has been integrated into mainline Flutter (it was the initial POC repo for Flutter Desktop), but there are a set of packages that were not integrated into Flutter, but that are still useful. 

The issue with ripping out `window_size` and not replacing it with a similar package is that a user can then resize a desktop app small enough to break the app's layout.

ericwindmill:
> > And according to the [plugin's Github](https://github.com/google/flutter-desktop-embedding), it's no longer necessary.
> 
> This is an incorrect reading of the linked page. The `window_size` package is not outdated per se. The previous contents of the linked repo has been integrated into mainline Flutter (it was the initial POC repo for Flutter Desktop), but there are a set of packages that were not integrated into Flutter, but that are still useful.
> 
> The issue with ripping out `window_size` and not replacing it with a similar package is that a user can then resize a desktop app small enough to break the app's layout.

You're right, I didn't mean outdated. I meant "no longer necessary" because it's part of flutter now. But even that I'm not sure about. Either way, I'll close this and investigate more

PR #2353, opened 2024-07-12
Compass App: Search form screen

miquelbeltran:
Part of the implementation of the Compass App for the Architecture sample.

**Merge to `compass-app`**

This PR introduces the Search Form Screen, in which users can select a region, a date range and the number of guests.

The feature is split in 5 different widgets, each one depending on the `SearchFormViewModel`. The architecture follows the same patterns implemented in the previous PR https://github.com/flutter/samples/pull/2342

TODO later on:

- Error handling is yet not implemented, we need to introduce a "logger" and a way to handle error responses.
- All repositories return local data, next steps include creating the dart server app.
- The search query at the moment only takes the "continent" and not the dates and number of guests, that would be implemented later on as well.

## Demo

[Screencast from 2024-07-12 14-30-48.webm](https://github.com/user-attachments/assets/afbcdd4e-617a-49cc-894c-8e082748e572)

## Pre-launch Checklist

- [x] I signed the [CLA].


compass_app/app/lib/data/repositories/region/region_repository_local.dart
miquelbeltran, line 1: Although the list of regions is short and will remain the same, I think we should put it on the server-side as well, even if for teaching purposes.

(besides, image urls may change, so it is good to have control over that)

compass_app/app/lib/routing/router.dart
```diff
@@ -1,21 +1,35 @@
...
+      routes: [
+        GoRoute(
+          path: 'results',
+          builder: (context, state) {
+            final viewModel = ResultsViewModel(
+              destinationRepository: context.read(),
+            );
+            final parameters = state.uri.queryParameters;
+            // TODO: Pass the rest of query parameters to the ViewModel
+            viewModel.search(continent: parameters['destination']);
```
miquelbeltran, line 28: I have a mix-up between "region" "destination" and "continent" here, I will clean this up in future PRs.

compass_app/app/test/util/mocks.dart
miquelbeltran, line 1: Personally, I prefer to use fakes instead of mocks when possible.
In this case, it was the easiest way to mock the `GoRouter` implementation and test that the route is called correctly with all query parameters.

ericwindmill (approved):

PR #2379, opened 2024-08-06
Flutter 3.24 update

ericwindmill:


tool/flutter_clean_script.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
```
ericwindmill, line 1: Added because add_to_app and platform samples often need to be cleaned

PR #2398, opened 2024-08-26
[test! do not merge!]  Refactor SearchFormViewModel to use Domain and UI models

ericwindmill:
This PR is only for demo purposes, and should not be merged. Everything works as expected. Tests haven't been updated, so take it all with a grain of salt.

### DIFF
--- a/compass_app/app/lib/data/models/continent.dart
+++ b/compass_app/app/lib/data/models/continent.dart
@@ -5,15 +5,15 @@ part 'continent.freezed.dart';
 part 'continent.g.dart';
 
 @freezed
-class ContinentApi with _$ContinentApi {
-  const factory ContinentApi({
+class Continent with _$Continent {
+  const factory Continent({
     /// e.g. 'Europe'
     required String name,
 
     /// e.g. 'https://rstr.in/google/tripedia/TmR12QdlVTT'
     required String imageUrl,
-  }) = _ContinentApi;
+  }) = _Continent;
 
-  factory ContinentApi.fromJson(Map<String, Object?> json) =>
-      _$ContinentApiFromJson(json);
+  factory Continent.fromJson(Map<String, Object?> json) =>
+      _$ContinentFromJson(json);
 }
--- a/compass_app/app/lib/ui/search_form/view_models/search_form_state.dart
+++ b/compass_app/app/lib/ui/search_form/view_models/search_form_state.dart
@@ -1,0 +1,40 @@
+import 'package:flutter/material.dart';
+
+import '../../../data/models/continent.dart';
+
+@immutable
+class SearchFormState {
+  SearchFormState({
+    /// List of the 7 inhabited continents:
+    /// North America, South America, Africa, Europe, Oceania, Asia, and Australia
+    this.continents = const <Continent>[],
+
+    /// The continent currently selected by the user
+    this.selectedContinent,
+
+    /// The selected date range
+    this.dateRange,
+
+    /// The number of guests
+    this.guests = 0,
+  });
+
+  final List<Continent> continents;
+  final Continent? selectedContinent;
+  final DateTimeRange? dateRange;
+  final int guests;
+
+  SearchFormState copy({
+    continents,
+    selectedContinent,
+    dateRange,
+    guests,
+  }) {
+    return SearchFormState(
+      continents: continents ?? this.continents,
+      selectedContinent: selectedContinent ?? this.selectedContinent,
+      dateRange: dateRange ?? this.dateRange,
+      guests: guests ?? this.guests,
+    );
+  }
+}
--- a/compass_app/app/lib/ui/search_form/view_models/search_form_viewmodel.dart
+++ b/compass_app/app/lib/ui/search_form/view_models/search_form_viewmodel.dart
@@ -1,11 +1,16 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:compass_model/model.dart';
 import 'package:logging/logging.dart';
 
+import '../../../data/models/continent.dart';
 import '../../../data/repositories/continent/continent_repository.dart';
 import '../../../data/repositories/itinerary_config/itinerary_config_repository.dart';
 import '../../../utils/command.dart';
 import '../../../utils/result.dart';
+import '../../../utils/view_command.dart';
+import 'search_form_state.dart';
 
 /// View model for the search form.
 ///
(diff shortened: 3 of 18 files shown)


compass_app/app/lib/ui/search_form/view_models/search_form_state.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:flutter/material.dart';
+
+import '../../../data/models/continent.dart';
+
+@immutable
```
ericwindmill, line 5: We would use Freezed IRL? I just used @immutable for demonstration purposes

compass_app/app/lib/ui/search_form/view_models/search_form_viewmodel.dart
```diff
@@ -24,55 +29,55 @@ class SearchFormViewModel extends ChangeNotifier {
   final _log = Logger('SearchFormViewModel');
   final ContinentRepository _continentRepository;
   final ItineraryConfigRepository _itineraryConfigRepository;
-  List<Continent> _continents = [];
-  String? _selectedContinent;
-  DateTimeRange? _dateRange;
-  int _guests = 0;
+
+  // Immutable data class that holds state the View needs to display
+  SearchFormState searchFormState = SearchFormState();
```
ericwindmill, line 34: This replaces all data properties used by the views.

compass_app/app/lib/ui/search_form/view_models/search_form_viewmodel.dart
```diff
@@ -24,55 +29,55 @@ class SearchFormViewModel extends ChangeNotifier {
...
   /// Set to null to clear the selection.
-  set selectedContinent(String? continent) {
-    _selectedContinent = continent;
+  /// A [TypedCommand] that set's the selected continent on the [SearchFormState]
+  void updateSelectedContinent(Continent? continent) {
+    if (continent == searchFormState.selectedContinent) {
+      searchFormState = searchFormState.copy(selectedContinent: null);
+    } else {
+      searchFormState = searchFormState.copy(selectedContinent: continent);
+    }
```
ericwindmill, line 50: I replaced all setters with functions. It's not necessary, but look at the new 'view_command' file I added for context.

compass_app/app/lib/ui/search_form/widgets/search_form_continent.dart
```diff
@@ -45,117 +41,7 @@ class SearchFormContinent extends StatelessWidget {
...
-                    } else {
-                      viewModel.selectedContinent = name;
-                    }
-                  },
-                ),
-              ),
-            ),
-          ],
-        ),
+        child: SearchFormContinentCarousel(viewModel: viewModel),
```
ericwindmill, line 44: I replaced this with a separate file ONLY because I'm writing about it in the documentation, so it's cleaner for the reader if the file is as small as possible. I might need to rename things to make this separation make sense.

compass_app/app/lib/ui/search_form/widgets/search_form_continent.dart
```diff
@@ -45,117 +41,7 @@ class SearchFormContinent extends StatelessWidget {
...
-    required this.viewModel,
-  });
-
-  final String imageUrl;
-  final String name;
-  final SearchFormViewModel viewModel;
-
-  bool _selected() =>
-      viewModel.selectedContinent == null ||
-      viewModel.selectedContinent == name;
```
ericwindmill, line 89: This refactor moves this logic to the ViewModel

compass_app/app/lib/ui/search_form/widgets/search_form_continent.dart
```diff
@@ -45,117 +41,7 @@ class SearchFormContinent extends StatelessWidget {
...
-            // Handle taps
-            Positioned.fill(
-              child: Material(
-                color: Colors.transparent,
-                child: InkWell(
-                  onTap: () {
-                    if (viewModel.selectedContinent == name) {
-                      viewModel.selectedContinent = null;
-                    } else {
-                      viewModel.selectedContinent = name;
```
ericwindmill, line 151: This refactor moves this logic to the viewModel

compass_app/app/lib/ui/search_form/widgets/search_form_continent_carousel.dart
```diff
@@ -0,0 +1,124 @@
...
+          padding: Dimens.of(context).edgeInsetsScreenHorizontal,
+          itemBuilder: (BuildContext context, int index) {
+            final continent = viewModel.searchFormState.continents[index];
+            return _CarouselItem(
+              key: ValueKey(continent.name),
+              continent: continent,
+              isSelectedContinent:
+                  viewModel.searchFormState.selectedContinent == continent,
+              updateSelectedContinentCommand: () =>
+                  viewModel.updateSelectedContinent(continent),
```
ericwindmill, line 35: _CarouselItem widget would need the Continent and the ViewModel passed in, so I just passed in the properties it needs from the VM instead. This is the only place where the ViewModel isn't passed around. There may be a good reason not to do this.

compass_app/app/lib/utils/view_command.dart
```diff
@@ -0,0 +1,112 @@
+// Copyright 2022 Google LLC
```
ericwindmill, line 1: This is a file from the library that Stadia used for architecture. These are the "Command" types that the ViewModel can expose to the View. (I changed the based class of Command to ViewCommand, so it wouldn't clash. Otherwise, it's the copied file.)

I'm not sure if the commands currently in Compass app and these commands are compatible, since the ones in the Compass app are handling error logic and more. We don't need to call everything a command, I just wanted to show you this incase its helpful.

compass_app/app/lib/utils/view_command.dart
```diff
@@ -0,0 +1,112 @@
...
+/// import 'package:flutter/widgets.dart' as flutter;
+///
+/// part 'view_model.g.dart';
+///
+/// abstract class ViewModel implements Built<...> {
+///   ...
+///   TypedCommand<flutter.BuildContext> get myCommand;
+/// }
+/// ```
+typedef ContextCommand = void Function(BuildContext);
```
ericwindmill, line 105: This is the most interesting part. Presumably this is useful for moving routing logic outside of the widgets, and for getting device information if necessary. 

Looking at this example, my updates in this PR differ from Stadias in that the methods are being directly accessed from the ViewModel. Their "ViewModels" don't contain logic if I understand correctly, but they have a library that binds ViewModels to Components, so I don't know that we'd want to concern ourselves with that.

compass_app/app/lib/utils/view_command.dart
ericwindmill: I find this file really interesting because without even seeing the ViewModel, Component and View classes, I can start to see how it all works.

compass_app/app/lib/utils/view_command.dart
miquelbeltran: One thing I find interesting with this, is that the `BuildContext` is taken outside Widgets and into the app logic. For example, you could start to have navigation code or showing snackbars implemented in the ViewModel side... I think this is very dangerous if used incorrectly and we should not encourage this. Instead, code that uses BuildContext should remain in the Widget.

compass_app/app/lib/ui/search_form/view_models/search_form_viewmodel.dart
miquelbeltran: This reminds me of Flutter Bloc "Cubit" implementation, where you have a State object and a Cubit, and then you subscribe to changes for this Cubit, getting a new State object: https://bloclibrary.dev/bloc-concepts/#cubit

It doesn't fully break from MVVM, but has some impracticalities.

For example, each time you change a property, you need to copy the old state and modify the value. This operation does not update the UI, but for that you also need to call to `NotifyListeners()`. This is where a package solution like `Cubit` comes in handy, since the `emit(new state)` call would replace the old state and trigger a UI refresh.

I'd probably prefer to keep ViewModels as they are, which follows the standard MVVM design pattern, where properties are directly exposed in the ViewModel instead of in a State object.

But I think it is very interesting to show an example on how this could be refactored into Flutter Bloc, Cubit or something like MVC.

compass_app/app/lib/data/models/continent.dart
miquelbeltran, line 1: Yes, this is the idea I had. Even maybe going further and having a `data/services/api/models` folder, since you could have other data sources that have different models (a database for example).

For example, `api_client.dart` would have their own model classes in `data/services/api/models` and the `api_client.dart` would go in `data/services/api`.

miquelbeltran: This is what I mean

image

ericwindmill: This looks good to me. The longer we work on this, the more I like as much separation as we can get

miquelbeltran: Good! I am already putting this in place for the User Home feature so we can see how it looks, later I will move the rest of the models.

PR #2421, opened 2024-08-30
Veggieseasons CI

ericwindmill:
Veggieseasons was merged despite CI failures due to a time crunch. This PR removed it from CI until it can be fixed


tool/flutter_ci_script_beta.sh
```diff
@@ -58,7 +58,8 @@ declare -ar PROJECT_NAMES=(
     "simplistic_calculator"
     "simplistic_editor"
     "testing_app"
-    "veggieseasons"
+#    todo @ewindmill
+#    "veggieseasons"
```
ericwindmill, line 62: @sethladd FYI -- There's a ci file for each flutter envrionment -- stable, beta, and main. Each file has a list of all the samples, and we can remove them from CI by commenting out the line for any given sample. 

We mainly do this ahead of releases when we're making updates to samples based on features that aren't in stable, or when there are samples that are considered experimental. We're pretty loose about commenting out samples as necessary, and I'll make sure they get un-commented and fixed when we update for the quarterly release.

sethladd:
RSLGTM

thank you!

tool/flutter_ci_script_beta.sh
domesticmouse: Do not remove `veggieseasons` from CI, I will fix in a follow up PR. My PR should not have been landed to master broken.

PR #2445, opened 2024-09-23
Rename components to use cases. 

ericwindmill:
We decided it was better to use the common term 'use case'.  This PR updates the names of files, classes and several variables.


miquelbeltran (approved):
LGTM!

domesticmouse (approved):

PR #2446, opened 2024-09-24
Compass app

ericwindmill:
Adds completed "compass_app" to the main branch. The compass app is used for the forthcoming architecture guidance on Flutter dev.  

This code has undergone many reviews, which you can see in [now closed PRs](https://github.com/flutter/samples/pulls?q=is%3Apr+is%3Aclosed+compass). Still, this code will be prominently featured in our documentation, and any and all feedback is helpful!

## Pre-launch Checklist

- [x] I signed the [CLA].


compass_app/app/README.md
```diff
@@ -0,0 +1,22 @@
+# compass_app
+
+A new Flutter project.
+
```
reidbaker, line 4: Consider providing a more complete description here.

compass_app/app/android/app/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,18 @@
...
+    <style name="LaunchTheme" parent="@android:style/Theme.Light.NoTitleBar">
+        <!-- Show a splash screen on the activity. Automatically removed when
+             the Flutter engine draws its first frame -->
+        <item name="android:windowBackground">@drawable/launch_background</item>
+    </style>
+    <!-- Theme applied to the Android Window as soon as the process has started.
+         This theme determines the color of the Android Window while your
+         Flutter UI initializes, as well as behind your Flutter UI while its
+         running.
+
```
reidbaker, line 13: This is probably generated code but it can be removed. V2 embedding is the only embedding that is supported.

reidbaker (commented):
One top level thing I notice is a bunch of these files do not have copyright licence at the top. Should they?

domesticmouse:
> One top level thing I notice is a bunch of these files do not have copyright licence at the top. Should they?

Yes

compass_app/app/test/data/repositories/activity/activity_repository_remote_test.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:compass_app/data/repositories/activity/activity_repository.dart';
+import 'package:compass_app/data/repositories/activity/activity_repository_remote.dart';
+import 'package:compass_app/utils/result.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '../../../../testing/fakes/services/fake_api_client.dart';
```
domesticmouse, line 6: Why do you have both `test` and `testing`? What problem does this organisation fix?

compass_app/app/test/data/repositories/activity/activity_repository_remote_test.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:compass_app/data/repositories/activity/activity_repository.dart';
```
domesticmouse, line 1: Add copyright header to *all* dart files.

domesticmouse (changes requested):

compass_app/app/test/data/repositories/activity/activity_repository_remote_test.dart
reidbaker: This came up during review. Testing is for shared test enablement code like common mocks or default complex data objects. 
Test is for tests that run.

reidbaker: I would think of testing like production code that does not ship to customers of your package.

compass_app/app/test/data/repositories/activity/activity_repository_remote_test.dart
ericwindmill: Done

compass_app/app/analysis_options.yaml
```diff
@@ -0,0 +1,5 @@
+include: package:flutter_lints/flutter.yaml
+
+linter:
+  rules:
+    - prefer_relative_imports
```
reidbaker, line 5: nit: newline

compass_app/app/android/app/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+
+android {
+    namespace = "com.example.compass_app"
+    compileSdk = flutter.compileSdkVersion
+    ndkVersion = flutter.ndkVersion
+
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
+    }
```
reidbaker, line 34: You probably dont need these lines. We use them for max compatibility in plugins but for apps you probably want the default behavior which is java 17.

compass_app/app/android/app/src/main/res/values-night/styles.xml
```diff
@@ -0,0 +1,18 @@
...
+        <!-- Show a splash screen on the activity. Automatically removed when
+             the Flutter engine draws its first frame -->
+        <item name="android:windowBackground">@drawable/launch_background</item>
+    </style>
+    <!-- Theme applied to the Android Window as soon as the process has started.
+         This theme determines the color of the Android Window while your
+         Flutter UI initializes, as well as behind your Flutter UI while its
+         running.
+
+         This Theme is only used starting with V2 of Flutter's Android embedding. -->
```
reidbaker, line 14: This comment can be removed. V2 embedding is the only embedding users can use right now.

compass_app/app/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.3-all.zip
```
reidbaker, line 5: You probably want to use something in the next major version of gradle. Maybe gradle 8.4

compass_app/app/android/settings.gradle
```diff
@@ -0,0 +1,25 @@
...
+    repositories {
+        google()
+        mavenCentral()
+        gradlePluginPortal()
+    }
+}
+
+plugins {
+    id "dev.flutter.flutter-plugin-loader" version "1.0.0"
+    id "com.android.application" version "7.3.0" apply false
```
reidbaker, line 21: You definitely want to use a higher version of AGP. Something that start with 8 at a minimum.

compass_app/app/android/settings.gradle
```diff
@@ -0,0 +1,25 @@
...
+    id "org.jetbrains.kotlin.android" version "1.8.0" apply false
```
reidbaker, line 22: @gmackall do you have a kotlin version recommendation?

compass_app/app/assets/activities.json
reidbaker, line 1: What are activities.json and destinations.json? Github wont even let me review it.

compass_app/app/integration_test/app_local_data_test.dart
```diff
@@ -0,0 +1,142 @@
...
+      // Activities Screen
+      expect(find.byType(ActivitiesScreen), findsOneWidget);
+
+      // Select one activity
+      await tester.tap(find.byType(CustomCheckbox).first);
+      await tester.pumpAndSettle();
+      expect(find.text('1 selected'), findsOneWidget);
+
+      // Submit selection
+      await tester.tap(find.byKey(const ValueKey('confirm-button')));
```
reidbaker, line 124: Nit: Shouldn't the key text come from a constant in the tested class. Keys unlike accessibility tags are not user visible and I do not see the value in this test failing if the key changed to "confirmation-button" as long as the button was found and the tap did the expected action.

compass_app/app/integration_test/app_server_data_test.dart
```diff
@@ -0,0 +1,207 @@
...
+import 'package:shared_preferences/shared_preferences.dart';
+
+/// This Integration Test starts the Dart server
+/// before launching the Compass-App with the remote configuration.
+/// The app connects to its endpoints to perform login and create a booking.
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  group('end-to-end test with remote data', () {
+    const port = '8080';
```
reidbaker, line 28: Consider a comment indicating how you decided on this port and what else would need to change if the value changed.

compass_app/app/integration_test/app_server_data_test.dart
```diff
@@ -0,0 +1,207 @@
...
+
+      // Home screen
+      expect(find.byType(HomeScreen), findsOneWidget);
+      await tester.pumpAndSettle();
+
+      // Should show user name
+      expect(find.text('Sofie\'s Trips'), findsOneWidget);
+
+      // Tap on booking (Alaska is created by default)
+      await tester.tap(find.text('Alaska, North America'));
```
reidbaker, line 91: There should probably be a test with an example of finding a localized string.

compass_app/app/integration_test/app_server_data_test.dart
```diff
@@ -0,0 +1,207 @@
...
+      // Select guests
+      await tester.tap(find.byKey(const ValueKey('add_guests')),
+          warnIfMissed: false);
+
+      // Refresh screen state
+      await tester.pumpAndSettle();
+
+      // Perform search and navigate to next screen
+      await tester.tap(find.byKey(const ValueKey('submit_button')));
+      await tester.pumpAndSettle(const Duration(seconds: 2));
```
reidbaker, line 165: Someplace you should document how you came up with 2 seconds (and I think the number should be much shorter for all of these). I think you should only need multi section settle if there is an animation playing.

reidbaker (commented):
Still reviewing but more comments added.

compass_app/app/assets/activities.json
miquelbeltran: image

Those contain Activities and Destinations to load as local data source. They come from the original Compass App sample by the Firebase team: https://github.com/FirebaseExtended/compass-ai-travel-planning-sample-flutter/blob/main/assets/activities.json

compass_app/app/android/app/build.gradle
miquelbeltran: @ericwindmill since the project was created some months ago, should we recreate the Android subproject with the current Flutter version before merging it?

compass_app/app/integration_test/app_server_data_test.dart
miquelbeltran: 8080 is the default port as set by the Dart shelf template. Both the app and the server are configured to use it by default.

It shouldn't be necessary to specify the port in this e2e integration test, so this probably is due to copying the `Process.start(...` section from the server tests.

@ericwindmill fyi, the port can be set both as env variable when launching the server, and when configuring the `ApiClient` in the app side. This could be documented in the root `README.md`.

compass_app/app/integration_test/app_server_data_test.dart
miquelbeltran: I think I left these by mistake when I was implementing the integration tests. After running the tests again without the duration value, it works as well, so they can be removed.

miquelbeltran: @ericwindmill do you want me to push directly to `compass-app`, or should I create a new PR?

ericwindmill: @miquelbeltran I've been pushing directly when addressing these comments. Feel free to do the same

compass_app/app/android/settings.gradle
ericwindmill: I re-created the Android app and the version is now 1.8.22

compass_app/app/integration_test/app_server_data_test.dart
ericwindmill: I'll add this documentation to the README, which I'm working on now

ericwindmill:
@miquelbeltran @domesticmouse Any idea why tests from the `/server` would be failing on Windows stable specifically? I've been searching but can't find anything.

miquelbeltran:
@ericwindmill the server tests start a local server and the unit tests connect to it, in order to test the enpoints.

```
ClientException with SocketException: A request to send or receive data was disallowed because the socket is not connected and (when sending on a datagram socket using a sendto call) no address was supplied.
   (OS Error: A request to send or receive data was disallowed because the socket is not connected and (when sending on a datagram socket using a sendto call) no address was supplied.
  , errno = 10057), address = 0.0.0.0, port = 50740, uri=http://0.0.0.0:8080/continent
  package:http/src/io_client.dart 154:7    IOClient.send
  ===== asynchronous gap ===========================
  package:http/src/base_client.dart 93:32  BaseClient._sendUnstreamed
  ===== asynchronous gap ===========================
  package:http/http.dart 167:12            _withClient
  ===== asynchronous gap ===========================
  test\server_test.dart 42:22              main.<fn>
```

My guess is that Windows doesn't like `0.0.0.0` as the IP. I will check if I can reproduce this on my Win machine and then see if an alternative works (e.g. `127.0.0.1`). Otherwise I will set the test to skip if platform is Windows or similar.

Note that this Dart server test setup was created by the Dart server template,

miquelbeltran:
Using `127.0.0.1` did the trick.

RedBrogdon (approved):

domesticmouse (approved):

PR #2449, opened 2024-09-25
fix: compass-app bug in dismissable homescreen item

miquelbeltran:
Doing some testing on the "Optimistic State" I noticed the use of `Dismissable` was incorrect.

If the delete request failed and the item cannot be removed, the removed item then gets stays in the item list.

This causes the `Dismissable` widget to throw an exception, because it expects that the item with a unique key has been removed. To fix that, I switched from `onDismissed` to `confirmDismiss`, and use the `Command` status result to know if the dismiss action was successful or not.

I have created a widget test that reproduces the issue.

## Pre-launch Checklist

- [x] I signed the [CLA].


ericwindmill (approved):
LGTM. I love that this bug validates the usefulness of commands.

PR #2490, opened 2024-11-21
Remove io sample

ericwindmill:
This sample has been moved to my personal Github, because we shouldn't maintain samples that are tied to events.  

- Moved sample to personal repo
- Removed sample from this repo
- Updated README here and the link to this repo in the YouTube video


domesticmouse:
If that were the justification, I'd have to delete half of my codelabs. 

I'd vote we keep samples that have value for our users, and AI is pretty hawt right now. So unless we have an alternate AI sample to replace it with, I'd be leaning heavily on the keep it.

ericwindmill:
> If that were the justification, I'd have to delete half of my codelabs.
> 
> I'd vote we keep samples that have value for our users, and AI is pretty hawt right now. So unless we have an alternate AI sample to replace it with, I'd be leaning heavily on the keep it.

We do have an alternative sample: https://github.com/flutter/samples/tree/main/gemini_tasks
I prefer that sample because this a vast majority of the code in this sample has nothing to do with Gemini. It's gonna be laborious to maintain, especially given that AI moves so quickly. 

Also, there are samples in the [pub package](https://github.com/google-gemini/generative-ai-dart/tree/main/samples) and a few more in [google-gemini github](https://github.com/google-gemini) I would much prefer to link to these from the forthcoming index page on the website, or even from the readme in this repo.

domesticmouse (approved):
LGTM

PR #2540, opened 2024-12-05
[compass_app] Mark classes used for namespacing as `abstract final`

parlough:
As they are just used for namespacing static properties, these classes shouldn't be extended, implemented, or instantiated.


miquelbeltran (approved):
neat trick!

domesticmouse (approved):

PR #2541, opened 2024-12-05
[compass_app] Don't expose Dimens subclasses

parlough:
Avoid exposing the subclasses as they shouldn't be instantiated again or overridden.

Also consistently use fields and getters in the declarations.


miquelbeltran (approved):
Much better, thanks!

domesticmouse (approved):

PR #2880, opened 2026-08-13
[compass_app] Refresh home after booking

### LINKED ISSUE #2877: [compass_app] UI state never updates for the HomeScreen (regression after the implementation of #2788)
Serious bug, application doesn't work the inteded way.

(Flutter 3.44.8; platform: Windows; tested on commit 978919b)

Steps to reproduce: 
1) Book a new trip;
2) Come back to a home screen;
3) You can't see new trip, unless you hot reload.

Before moving initialization of HomeViewModel into HomeScreenContainer, it relied on rebuild that go_router did: 
1) User moves to the home route '/' or deeper;
2) All previous routes including the one where user moved are getting rebuilt, which results in creating new instances of viewModels, which results in them calling _load and fetching latest data from repository.

Now it seems that HomeViewModel is just initialized once in HomeScreenContainer and never updates.

---

Generally according to [this statement](https://github.com/flutter/samples/issues/2604#issuecomment-3164086227), the current way go_router "manages" ui states, is wrong, which already was unfolded in [issue#2574](https://github.com/flutter/samples/issues/2574), but outside of users sharing their ways on how to avoid this behavior, it never really recieved much attention. 

So maybe a solution to this bug will also lead to a solution for issue#2574?

I myself temporarly solved issue#2574 in a way where viewModel only updates when user moves directly to a route it was going to. Hopefully this helps:
```
class RouteLifecycleListener extends StatefulWidget {
  const RouteLifecycleListener({
    required this.onResume,
    required this.child,
    super.key,
  });
  final VoidCallback onResume;
  final Widget child;

  @override
  State<RouteLifecycleListener> createState() => _RouteLifecycleListenerState();
}

class _RouteLifecycleListenerState extends State<RouteLifecycleListener>
    with RouteAware {
  @override
  void didChangeDependencies() {
    super.didChangeDependencies();
    routeObserver.subscribe(this, ModalRoute.of(context)! as PageRoute);
  }

  @override
  void didPopNext() {
    widget.onResume();
  }

  @override
  void dispose() {
    routeObserver.unsubscribe(this);
    super.dispose();
  }

  @override
    return widget.child;
  }
}
```

```
...
    GoRoute(
      path: '/home',
      name: AppRoute.home.name,
      builder: (context, state) => Provider(
        create: (context) {
          return HomeViewModel(
            ...
          );
        },
        builder: (context, _) {
          return RouteLifecycleListener(
            onResume: () => context.read<HomeViewModel>().load.execute(),
            child: HomeScreen(viewModel: context.read()),
          );
        },
      ),
      routes: [ ...
```

Ersaoktaviannn:
## Summary

- reload `HomeViewModel` when the retained home route becomes active again
- keep route-listener ownership alongside the view model in `HomeScreenContainer`
- add a regression test covering a booking created while a child route is active
- document the fix in the changelog

Fixes #2877.

## Testing

- `flutter analyze --no-pub` (from `compass_app/app`)
- `flutter test --no-pub` (68 tests)
- verified the regression test fails without the route listener and passes with the fix

## Pre-launch Checklist

- [x] I read the [Flutter Style Guide] recently, and have followed its advice.
- [x] I signed the [CLA].
- [x] I have added sample code updates to the [changelog].
- [ ] I updated/added relevant documentation (no public API documentation changes are needed).

[changelog]: https://github.com/flutter/samples/blob/main/CHANGELOG.md


Ersaoktaviannn:
@googlebot rescan

PR #2884, opened 2026-09-02
Migrate standard sample apps to UIScene lifecycle

LouiseHsu:
Fixes https://github.com/flutter/flutter/issues/176957


add_to_app/fullscreen/ios_fullscreen/IOSFullScreen/AppDelegate.swift
```diff
@@ -2,20 +2,40 @@
...
+      name: "Default Configuration",
+      sessionRole: connectingSceneSession.role
+    )
+    configuration.delegateClass = FlutterSceneDelegate.self
+    return configuration
+  }
 
-        return true
-    }
+  func application(
```
okorohelijah, line 36: Since this is optional and no-op, we can either delete it or keep it with a one-line comment explaining what a host would do here

add_to_app/fullscreen/ios_fullscreen/Podfile
```diff
@@ -1,5 +1,4 @@
-# Uncomment the next line to define a global platform for your project
-# platform :ios, '9.0'
```
okorohelijah, line 2: I think we can remove this?

okorohelijah (changes requested):

